PR #2, opened 2024-07-10
refactor: Improve build configuration

dcalhoun:
- Set a project Node.js version mirroring the Gutenberg project to avoid version conflicts. 
- Enable web view inspection to improve debugging. 
- Add effect clean up to avoid block registration errors.


Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -81,6 +81,10 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         ])
 
         webView.alpha = 0
+        
+        if #available(iOS 16.4, *) {
+            webView.isInspectable = true
```
kean, line 86: I'm not sure what's the implication of making a web view inspectable, but I decided to keep it off by default. The app enables it in the debug mode.

kean (commented):
This repo was originally intended to be a throw-away prototype—especially the CSS code—but there is probably some value in keeping it. The Swift package, the WKWebView management code, and the Vite setup aren't so bad.

I'm going to move this to the `wordpress-mobile` organization.

PR looks good, except for one comment I left.

Sources/GutenbergKit/Sources/EditorViewController.swift
dcalhoun: I interpet "debug mode" to mean the scheme configuration for running the app. I have that enabled, but the web view does not display as inspectable for me in Safari. Am I missing something? 

| Xcode Scheme | Safari Inspector |
| - | -|
| run-with-debug | no-inspectable-applications |

dcalhoun:
> This repo was originally intended to be a throw-away prototype—especially the CSS code—but there is probably some value in keeping it. The Swift package, the WKWebView management code, and the Vite setup aren't so bad.
> 
> I'm going to move this to the `wordpress-mobile` organization.

Make sense. I still consider it throw away as well, but figured continued exploration were worth sharing in a central location.

Sources/GutenbergKit/Sources/EditorViewController.swift
kean: I moved this to the Demo if you don't mind – merging.

PR #4, opened 2024-07-11
feat: Improve block inspector accessibility

dcalhoun:
Explore accessibility improvements spanning across both web and native UI. 

* Communicate dialog UI to accessibility tools. 
* Prevent accessibility tools moving selection outside of the block inspector dialog.


dcalhoun:
Closing as this exploratory PR as it served its purpose, it can always be referenced as needed. The targeted functionality will be implemented in a few separate PRs.

PR #5, opened 2024-07-16
Add Android app code

jkmassel:
Adds a basic Android app for testing


ReactApp/package.json
```diff
@@ -5,11 +5,11 @@
 	"version": "0.0.0",
 	"type": "module",
 	"scripts": {
-		"dev": "vite",
+		"dev": "vite --host",
```
geriux, line 8: Nice! I like having the `host` arg as the default.

Sources/GutenbergKit/Gutenberg/assets/index.css
geriux, line 1: If we plan to delete the build assets, how can we have them in the host apps when they import GutenbergKit?

I guess we'll have some CI jobs that will do this for releases or for any new commits in `trunk`, should we keep them until we have something implemented for that?

ReactApp/src/misc/Helpers.js
```diff
@@ -1,8 +1,43 @@
-export function postMessage(message, parameters = {}) {
-	if (window.webkit) {
-		const value = { message: message, body: parameters };
-		window.webkit.messageHandlers.editorDelegate.postMessage(value);
+export function editorLoaded() {
+	console.log("Firing JS editorLoaded event");
```
geriux: If this was just for testing, could we remove it?

ReactApp/src/misc/Helpers.js
```diff
@@ -1,8 +1,43 @@
...
+
+	if(window.editorDelegate) {
+		editorDelegate.onEditorLoaded();
```
geriux: Although by itself `editorDelegate` exists, to prevent ESLint formatting warnings, could we update this to:

```suggestion
		window.editorDelegate.onEditorLoaded();
```

ReactApp/src/misc/Helpers.js
```diff
@@ -1,8 +1,43 @@
...
+		window.webkit.messageHandlers.editorDelegate.postMessage({
+			message: 'onEditorLoaded',
+			body: {}
+		});
+	}
+}
+
+export function onBlocksChanged(isEmpty = false) {
+	if(window.editorDelegate) {
+		editorDelegate.onBlocksChanged(isEmpty);
```
geriux: Although by itself `editorDelegate` exists, to prevent ESLint formatting warnings, could we update this to:

```suggestion
		window.editorDelegate.onBlocksChanged(isEmpty);
```

ReactApp/src/misc/Helpers.js
```diff
@@ -1,8 +1,43 @@
...
+		window.webkit.messageHandlers.editorDelegate.postMessage({
+			message: 'onBlocksChanged',
+			body: { isEmpty: isEmpty }
+		});
+	}
+}
+
+export function showBlockPicker() {
+	if(window.editorDelegate) {
+		editorDelegate.showBlockPicker();
```
geriux: Same as above:

```suggestion
		window.editorDelegate.showBlockPicker();
```

ReactApp/src/misc/Helpers.js
geriux, line 1: I noticed some formatting warnings when opening this file. I usually set Visual Studio Code to apply formatting fixes when I save changes, but we could also do `npm run format`.

Can we please fix the formatting warnings before merging these changes? Thanks!

geriux (commented):
Thank you for working on adding the Android Demo app!

I noticed that the WebView's height doesn't adapt when the keyboard is opened causing the Toolbar to not be visible, as well as when you add content or add new paragraphs, the content gets hidden under the keyboard.

<kbd></kbd>

I tested two different simulators to verify it wasn't something specific to a device's config. Can you replicate this issue?

Demo-Android/app/src/main/java/com/example/gutenbergkit/GutenbergView.kt
```diff
@@ -0,0 +1,112 @@
...
+        // this value out of the `dist` directory after building GutenbergKit
+        this.loadUrl("file:///android_asset/index.html")
+
+        // Dev mode – you can connect the app to a local dev server and have it refresh as
+        // changes are made. To start the server, run `make dev-server` in the project root
+        // directory.
+        //
+        // You'll need to set this to the IP address of your local machine
+        //
+        // this.loadUrl("http://192.168.5.248:5173/")
```
dcalhoun, line 71: Long term, I suppose we might mirror iOS' setup where a `GUTENBERG_EDITOR_URL` environment variable toggles dev mode and sets the IP and port. I don't feel like we have to address this before merging.

Demo-Android/app/src/main/java/com/example/gutenbergkit/GutenbergView.kt
```diff
@@ -0,0 +1,112 @@
...
+                    Log.i("GutenbergView", consoleMessage.message())
+                } else {
+                    Log.i("GutenbergView", "null message")
+                }
+                return super.onConsoleMessage(consoleMessage)
+            }
+        }
+
+        // Production mode – load the assets from the app bundle – you'll need to copy
+        // this value out of the `dist` directory after building GutenbergKit
```
dcalhoun, line 62: Is this comment still true now that a `Makefile` exists that copies the built assets?

Demo-iOS/Sources/ContentView.swift
```diff
@@ -4,7 +4,7 @@ import GutenbergKit
 struct ContentView: View {
     var body: some View {
         NavigationView {
-            EditorView()
+            EditorView(editorURL: URL(string: "http://localhost:5173/")!)
```
dcalhoun, line 7: Should this rely upon the `GUTENBERG_EDITOR_URL` environment variable instead?

ReactApp/src/components/EditorToolbar.jsx
```diff
@@ -20,7 +20,7 @@ const EditorToolbar = (props) => {
 		addBlockButton = <Inserter />;
 	} else {
 		addBlockButton = (
-			<button onClick={() => postMessage('showBlockPicker')}>+</button>
+			<button onClick={() => showBlockPicker() }>+</button>
```
dcalhoun: We might remove the unnecessary anonymous function to improve readability. Feel free to disagree. 
```suggestion
			<button onClick={ showBlockPicker }>+</button>
```

ReactApp/src/misc/Helpers.js
```diff
@@ -1,8 +1,43 @@
...
+
+export function onBlocksChanged(isEmpty = false) {
+	if(window.editorDelegate) {
+		editorDelegate.onBlocksChanged(isEmpty);
+	}
+
+	if(window.webkit) {
+		window.webkit.messageHandlers.editorDelegate.postMessage({
+			message: 'onBlocksChanged',
+			body: { isEmpty: isEmpty }
```
dcalhoun: We might rely upon the shorthand notation to improve readability. Feel free to disagree. 
```suggestion
			body: { isEmpty }
```

Sources/GutenbergKit/Gutenberg/assets/index.css
dcalhoun: > If we plan to delete the build assets, how can we have them in the host apps when they import GutenbergKit?

I presume building GutenbergKit (`make build`) before building the host app would address this. Am I overlooking something? 

If we remove these file (which I support), I suggest we also remove the `Sources/GutenbergKit/Gutenberg/index.html` file and add the build output to the `.gitignore` configuration to avoid unnecessary untracked files messages: 

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        Demo-Android/app/src/main/assets/assets/
        Sources/GutenbergKit/Gutenberg/assets/
```

Sources/GutenbergKit/Sources/GBWebView.swift
```diff
@@ -0,0 +1,10 @@
+import WebKit
+
+class GBWebView: WKWebView {
+
+    /// Disables the default bottom bar that competes with the Gutenberg inserter
+    ///
+    override var inputAccessoryView: UIView? {
+        nil
+    }
```
dcalhoun, line 9: For my own edification, is this hiding the toolbar with the previous/next/done buttons above the keyboard while focusing an HTML text input?

dcalhoun (commented):
Thanks for completing this. Very helpful. 👍🏻 The Android demo app worked for me. 

I left a few suggestions, but feel free to disregard them as items we could complete later. 

As @geriux [mentioned](https://github.com/wordpress-mobile/GutenbergKit/pull/5#discussion_r1724633522), I recommend we format the JavaScript (`npm run format`) before merging. Eventually, we'll set up a linter to avoid needing to discuss that particular subject during code review.

Sources/GutenbergKit/Sources/GBWebView.swift
jkmassel: Yep, but more importantly this dismisses it while using an external keyboard (which messes with the block inserter)

Demo-iOS/Sources/ContentView.swift
jkmassel: Yeah, probably – this could use a bigger refactor anyway so I'm comfortable leaving it for a future change

jkmassel:
I think I've addressed most of the feedback – there are some other forthcoming PRs that'll add an Android module and some other stuff so some of the more structural stuff should probably be addressed there.

ReactApp/src/components/Editor.jsx
```diff
@@ -53,8 +53,8 @@ const POST_MOCK = {
 
 function Editor({ post = POST_MOCK }) {
 	const [blocks, setBlocks] = useState([]);
-	const [registeredBlocks, setRegisteredBlocks] = useState([]);
-	const [isCodeEditorEnabled, setCodeEditorEnabled] = useState(false);
+	const [registeredBlocks] = useState([]);
+	const [setCodeEditorEnabled] = useState(false);
```
dcalhoun: I interpret the unused values were removed to address linter warnings. I suggest we prefix them with `_` instead of removing them, as removing `isCodeEditorEnabled` modifies the code behavior. 

The [`useState`](https://react.dev/reference/react/useState#reference) React Hook returns a positional array. Removing the `isCodeEditorEnabled` from this destructured array effectively renames the first item (`Boolean`) to `setCodeEditorEnabled` (implied `Function`). 

```suggestion
	const [registeredBlocks] = useState([]);
	const [_isCodeEditorEnabled, setCodeEditorEnabled] = useState(false);
```

Sources/GutenbergKit/Gutenberg/assets/index.css
dcalhoun: I realize now that @geriux was likely referring to the ability to have a host app in-progress branch point to a GutenbergKit commit, allowing the CI server to create a prototype build for a given PR. The ability will be broken until we build a GutenbergKit CI publishing build artifacts. 

I say we merge this and reinstate the GutenbergKit build assets in a separate PR, if desired.

dcalhoun (approved):
Latest changes work well for me on Android and iOS. 🚀 Thank you!

Sources/GutenbergKit/Gutenberg/assets/index.css
geriux: Exactly that's what I meant. That sounds good to me! 🚀

PR #9, opened 2024-08-05
Adds Post Title

geriux:
This PR introduces support for the `PostTitle` component in the editor by adding the following:

- Implements initial `setupEditor` support, currently setting up mock data for a post.
- Adds the `@wordpress/editor` and `@wordpress/core-data` packages.
- Updates other packages to maintain compatibility with the new additions.

<kbd></kbd>

### DIFF
--- a/ReactApp/src/components/Editor.jsx
+++ b/ReactApp/src/components/Editor.jsx
@@ -107,33 +107,31 @@ function Editor({ post = POST_MOCK }) {
 	// }
 
 	return (
-		<>
+		<BlockEditorProvider
+			value={blocks}
+			onInput={didChangeBlocks}
+			onChange={didChangeBlocks}
+			settings={settings}
+		>
 			<div className="editor-visual-editor__post-title-wrapper">
 				<PostTitle ref={titleRef} />
 			</div>
-			<BlockEditorProvider
-				value={blocks}
-				onInput={didChangeBlocks}
-				onChange={didChangeBlocks}
-				settings={settings}
-			>
-				<BlockTools>
-					<div className="editor-styles-wrapper">
-						<BlockEditorKeyboardShortcuts.Register />
-						<WritingFlow>
-							<ObserveTyping>
-								<BlockList />
-								<EditorToolbar
-									registeredBlocks={registeredBlocks}
-								/>{' '}
-								{/* not sure if optimal placement */}
-							</ObserveTyping>
-						</WritingFlow>
-					</div>
-				</BlockTools>
-				<Popover.Slot />
-			</BlockEditorProvider>
-		</>
+			<BlockTools>
+				<div className="editor-styles-wrapper">
+					<BlockEditorKeyboardShortcuts.Register />
+					<WritingFlow>
+						<ObserveTyping>
+							<BlockList />
+							<EditorToolbar
+								registeredBlocks={registeredBlocks}
+							/>{' '}
+							{/* not sure if optimal placement */}
+						</ObserveTyping>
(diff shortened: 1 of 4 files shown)


ReactApp/src/components/Editor.jsx
```diff
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
```
dcalhoun, line 1: Unrelated to these changes, but we should likely source these from `@wordpress/element`. I believe that is the common Gutenberg practice.

ReactApp/src/components/Editor.jsx
```diff
@@ -35,10 +37,21 @@ import { postMessage } from '../misc/Helpers';
 // Current editor (assumes can be only one instance).
 let editor = {};
 
-function Editor() {
+const POST_MOCK = {
+	id: 1,
+	type: 'post',
+};
```
dcalhoun, line 43: Do you consider this a temporary measure until we pass this information from the host app?

ReactApp/src/components/Editor.jsx
```diff
@@ -35,10 +37,21 @@ import { postMessage } from '../misc/Helpers';
...
+function Editor({ post = POST_MOCK }) {
 	const [blocks, setBlocks] = useState([]);
 	const [registeredBlocks, setRegisteredBlocks] = useState([]);
 	const [isCodeEditorEnabled, setCodeEditorEnabled] = useState(false);
+	const titleRef = useRef();
+	const { setupEditor } = useDispatch(editorStore);
+
+	useEffect(() => {
+		setupEditor(post, [], {});
+	}, []);
```
dcalhoun, line 54: I researched briefly, but I was unable to determine why the [web editor invokes](https://github.com/WordPress/gutenberg/blob/e44bc77a252e81a549652e6ab480d62cb8184d5e/packages/editor/src/components/provider/index.js#L235) `setupEditor` within a `useLayoutEffect` call. Any ideas as to why that is or if we should do so?

dcalhoun (approved):
Thanks for implementing this. 

I noted that the actual post title does not populate this UI when opening a post in the WPiOS. I presume that is expected. 

I left a few suggestions, but nothing I consider a blocker.

ReactApp/src/components/Editor.jsx
geriux: I agree! We should add the element package to adopt that practice in this project. Should we tackle that in a separate PR?

ReactApp/src/components/Editor.jsx
geriux: Exactly, we should expect a `post` object from the host app, for now, I added the mock data.

ReactApp/src/components/Editor.jsx
dcalhoun: Addressing this separately makes sense to me. 👍🏻

ReactApp/src/components/Editor.jsx
geriux: Yeah, it wasn't clear to me why `useLayoutEffect` was used, so I ended up using `useEffect`.

Since we might need to adjust this if we expect changes in the post object (I'm not sure if that happens once the editor is opened on the host app), we could evaluate whether to keep it as it is or follow the same approach. What do you think?

dcalhoun: Relying upon `useEffect` until we identify a need for `useLayoutEffect` makes sense to me.

PR #11, opened 2024-08-09
Add media upload support within the Web editor

geriux:
This PR adds support for media uploads within the Web editor.

It checks the user's capabilities to determine media upload permissions, with the default value set to `true`.

When connected to the API, it uploads images using the built-in Web functionality.

<video src="https://github.com/user-attachments/assets/8f43d2fc-5bbb-44a4-8c62-c08e4187585a" width=250 />

### DIFF
--- a/ReactApp/package-lock.json
+++ b/ReactApp/package-lock.json
@@ -10,9 +10,9 @@
 			"dependencies": {
 				"@monaco-editor/react": "^4.6.0",
 				"@wordpress/api-fetch": "^7.4.0",
-				"@wordpress/block-editor": "^13.0.0",
-				"@wordpress/block-library": "^9.0.0",
-				"@wordpress/format-library": "^5.1.0",
+				"@wordpress/block-editor": "^14.0.0",
+				"@wordpress/block-library": "^9.5.0",
+				"@wordpress/format-library": "^5.5.0",
 				"react": "^18.3.1"
 			},
 			"devDependencies": {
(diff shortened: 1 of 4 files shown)


dcalhoun (approved):
Changes look good to me. 

Was there anything you'd like to me to test? When attempting to use this feature, I noted that: 

* In Safari/Chrome, uploading fails due to a missing `http://localhost:5173/wp/v2/media?_locale=user` endpoint;
* And in WPiOS the upload option is absent. 

Is that expected?

geriux:
> Was there anything you'd like to me to test? 

I think we'll need to have a local setup where we can store our API credentials for a testing site to test future changes. Currently, I manually set these in the `api-fetch-setup`.

For the host app, once we implement a way to retrieve the credentials, that could be another way to test changes.

So, what you encountered was expected:

When testing in a web browser without API credentials, the endpoints won't work and will point to the localhost path.
For the host apps, the upload option might be present when testing on the local server, but it won't work without the API credentials.

I think it's fine if you don't test it but if you want to I do this:

```diff

diff --git a/ReactApp/src/misc/api-fetch-setup.js b/ReactApp/src/misc/api-fetch-setup.js
index 17cb0e8..f1b080e 100644
--- a/ReactApp/src/misc/api-fetch-setup.js
+++ b/ReactApp/src/misc/api-fetch-setup.js
@@ -10,7 +10,12 @@ import apiFetch from '@wordpress/api-fetch';
  * and preloads some endpoints with mock data for specific components.
  */
 export function initializeApiFetch() {
-	apiFetch.use(apiFetch.createRootURLMiddleware(''));
+	const url = 'https://my-site/wp-json/';
+	const username = 'user';
+	const appPassword = 'secret';
+
+	const basicAuth = btoa(`${username}:${appPassword}`);
+	apiFetch.use(apiFetch.createRootURLMiddleware(url));
 
 	apiFetch.use((options, next) => {
 		options.headers = options.headers || new Headers();
@@ -19,7 +24,7 @@ export function initializeApiFetch() {
 		return next({
 			...options,
 			headers: {
-				Authorization: 'Basic ',
+				Authorization: `Basic ${basicAuth}`,
 			},
 		});
 	});
```

PR #12, opened 2024-08-12
Adds WordPress Element package

geriux:
Following this [comment](https://github.com/wordpress-mobile/GutenbergKit/pull/9#discussion_r1711677965) we are adding the [@wordpress/element package](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-element/) to follow Core's practices to not import from React directly.


dcalhoun (approved):
Thanks!

PR #14, opened 2024-08-13
feat: Authenticated API requests

dcalhoun:
Enable authenticated API requests to a site URL via the basic authentication
scheme with application passwords.

## Testing Instructions

1. Manually set `siteURL` and `authToken` values within the
   `EditorViewController` constructor.
1. Build and run the demo app.
1. Inspect the network traffic and verify the preloaded requests succeed.

### DIFF
--- a/ReactApp/src/misc/api-fetch-setup.js
+++ b/ReactApp/src/misc/api-fetch-setup.js
@@ -19,22 +19,8 @@ export function initializeApiFetch() {
 	const rootURL = siteURL ? `${siteURL}/wp-json/` : undefined;
 
 	apiFetch.use(apiFetch.createRootURLMiddleware(rootURL));
-	apiFetch.use((options, next) => {
-		options.mode = 'cors';
-		return next(options);
-	});
-	apiFetch.use((options, next) => {
-		options.headers = {
-			...options.headers,
-			'Content-Type': 'application/json',
-		};
-
-		if (authToken) {
-			options.headers.Authorization = `Basic ${authToken}`;
-		}
-
-		return next(options);
-	});
+	apiFetch.use(corsMiddleware);
+	apiFetch.use(createHeadersMiddleware(authToken));
 
 	// Preload some endpoints to return data needed for some components
 	// Like PostTitle.
@@ -117,4 +103,24 @@ export function initializeApiFetch() {
 			},
 		})
 	);
+}
+
+function corsMiddleware(options, next) {
+	options.mode = 'cors';
+	return next(options);
+}
+
+function createHeadersMiddleware(authToken) {
+	return (options, next) => {
+		options.headers = {
+			...options.headers,
+			'Content-Type': 'application/json',
+		};
+
+		if (authToken) {
+			options.headers.Authorization = `Basic ${authToken}`;
+		}
+
+		return next(options);
+	};
 }
--- a/ReactApp/src/misc/api-fetch-setup.js
+++ b/ReactApp/src/misc/api-fetch-setup.js
@@ -112,10 +112,7 @@ function corsMiddleware(options, next) {
 
 function createHeadersMiddleware(authToken) {
 	return (options, next) => {
-		options.headers = {
-			...options.headers,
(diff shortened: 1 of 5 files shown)


ReactApp/src/misc/api-fetch-setup.js
```diff
@@ -3,26 +3,24 @@
...
  * and preloads some endpoints with mock data for specific components.
  */
 export function initializeApiFetch() {
-	apiFetch.use(apiFetch.createRootURLMiddleware(''));
+	const { siteURL, authToken } = getGBKit();
+	const rootURL = siteURL ? `${siteURL}/wp-json/` : undefined;
 
-	apiFetch.use((options, next) => {
-		options.headers = options.headers || new Headers();
-		options.headers.set('Content-Type', 'application/json');
```
dcalhoun, line 17: I found that using the `Headers` interface and `Headers.prototype.set` resulted in the values never being present in the request headers. It appears that `apiFetch` [expects an object](https://github.com/WordPress/gutenberg/blob/5dae25b15e82a59d28c6c065b5a3745f3f83980f/packages/api-fetch/src/index.js#L93-L94) instead, so I refactored this to an object.

dcalhoun: Additionally, the `Content-Type` request header is [managed by](https://github.com/WordPress/gutenberg/blob/5dae25b15e82a59d28c6c065b5a3745f3f83980f/packages/api-fetch/src/index.js#L96-L100) `@wordpress/api-fetch` as long as we use the default fetcher, so I removed this redundant value.

ReactApp/src/misc/api-fetch-setup.js
```diff
@@ -3,26 +3,24 @@
...
 /**
  * Initializes the API fetch configuration and middleware.
  *
  * This function sets up the root URL middleware, adds headers to requests,
  * and preloads some endpoints with mock data for specific components.
  */
 export function initializeApiFetch() {
-	apiFetch.use(apiFetch.createRootURLMiddleware(''));
+	const { siteURL, authToken } = getGBKit();
+	const rootURL = siteURL ? `${siteURL}/wp-json/` : undefined;
```
geriux, line 19: I've just realized that for WordPress.com sites, we'll need to use `https://public-api.wordpress.com/` instead. This means we'll need to remove the wp-json path and include it in the `siteURL` parameter. I suggest we address this in a separate PR, as I encountered some issues with the endpoint paths when attempting to connect to `https://public-api.wordpress.com/`.

For now this will work perfectly for self-hosted sites.

geriux (approved):
LGTM! I've tested it using the host app and I was able to pass the data successfully. 🚀 

Nice work!

ReactApp/src/misc/api-fetch-setup.js
dcalhoun: Interesting. I make successful requests with this approach for an Atomic WPCOM site. Is the use of `https://public-api.wordpress.com/` merely a recommendation and/or potentially required for Simple sites?

geriux: I was trying with a simple site yeah. For Atomic, it's the same as self-hosted using the site's URL, but for Simple sites I think we'll need to go through the WordPress.com API 🤔 

I ended up trying this because I checked the editor's Web view source code and for Simple sites, it sets `https://public-api.wordpress.com/` for the `createRootURLMiddleware` middleware and for Atomic, it just uses the site's URL e.g. `https://geriuxxtestingbusiness123.wpcomstaging.com/wp-json/`

dcalhoun: Makes sense. When we address this, we might consider renaming `siteURL` to something like `siteAPIURL`.

geriux: Addressed [here](https://github.com/wordpress-mobile/GutenbergKit/pull/15/files#diff-230d4e17b4d9726d839cbab42a1cb29f2ad91e6ada136a558c6df6e421b4976fR18)

PR #15, opened 2024-08-14
Update API Fetch config

geriux:
This PR updates the API Fetch configuration to support different types of API endpoints.

- Renames `siteURL` to `siteApiRoot`, `authToken` to `authHeader`.
- Removes the appending of `/wp-json/` to the API URL, as this should be set by `siteApiRoot` if applies.
- Adds the `siteApiNamespace` parameter.
- Adds `apiPathModifierMiddleware`, it introduces new logic to append some URL namespaces depending on whether the `siteApiNamespace` parameter exists. This change adds support for certain WordPress.com sites that use the `https://public-api.wordpress.com/` endpoint. Currently, this logic is handled by the host apps ([see this](https://github.com/wordpress-mobile/WordPress-iOS/blob/46c0aaa3131e31b7e729db451ae77814dc3f5792/WordPress/Classes/ViewRelated/Gutenberg/GutenbergNetworking.swift#L60-L64)), but it might make more sense to include it within the API fetch setup in a more general way.
- Updates the `injectEditorConfiguration` logic due to issues found while loading the WebView in a prototype build using the built assets. After debugging, I discovered that the data wasn't being set, so I moved this to be added with [addUserScript](https://developer.apple.com/documentation/webkit/wkusercontentcontroller/1537448-adduserscript).

## To test

> [!NOTE]  
> You can use the following [WordPress iOS build](https://github.com/wordpress-mobile/WordPress-iOS/pull/23499#issuecomment-2289528270)

The current implementation in the above PR does not support all sites, currently only for .com sites.

### Test case 1 - Image upload
**Note:** For private sites, the images will be uploaded but they won't be shown, unless the site is in "Coming Soon" mode. I think this could be solved from the host apps by adding the authentication cookies.

- Open the app
- Open the editor
- Add a Gallery block
- Select some images
- **Expect** to see the images to upload correctly

### Test case 2 - Dynamic block

- Open the app
- Open the editor
- Add the Latest Posts or Query block
- **Expect** the block to fetch the data

### Test case 3 - Embed block

- Open the app
- Open the editor
- Add an embed block like `Twitter Embed`
- **Expect** the block to fetch the embed data


ReactApp/src/misc/api-fetch-setup.js
```diff
@@ -110,12 +110,24 @@ function corsMiddleware(options, next) {
...
+		if (
+			options.path &&
+			siteID.length &&
+			!options.path.includes(`/sites/${siteID}/`)
+		) {
+			options.path = options.path
+				.replace('/wp/v2/', `/wp/v2/sites/${siteID}/`)
+				.replace('/wpcom/v2/', `/wpcom/v2/sites/${siteID}/`)
+				.replace('/oembed/1.0/', `/oembed/1.0/sites/${siteID}/`);
+		}
```
dcalhoun: This logic appears unrelated to headers. I suggest we relocate it to a new middleware function — e.g., `createPathModifcationMiddleware` or similar. 

However, given our long-term hope of merging this into the Gutenberg project, this feels out of place in GutenbergKit. I wonder if we should hoist this middleware definition into the host app. WDYT? 

```swift
private func getEditorConfiguration() -> WKUserScript {
    // ...
    let jsCode = """
    window.GBKit = {
        siteAPIURL: '\(siteAPIURL)',
        authToken: '\(authToken)',
        siteID: '\(siteIDString)',
        modifyAPIPath
    };
    function modifyAPIPath(path) {
      if (
        !path ||
        !siteID.length ||
        path.includes(`/sites/${siteID}/`)
      ) {
        return path;
      }

      return path
        .replace('/wp/v2/', `/wp/v2/sites/${siteID}/`)
        .replace('/wpcom/v2/', `/wpcom/v2/sites/${siteID}/`)
        .replace('/oembed/1.0/', `/oembed/1.0/sites/${siteID}/`);
    }
    // ...
}
```

```js
apiFetch.use(pathModificationMiddleware); 

// ...

function pathModificationMiddleware(options, next) {
  const{ modifyAPIPath = () => {} } = getGBKit();
  options.path = modifyAPIPath(options.path);
  return next(options);
}
```

Sources/GutenbergKit/Gutenberg/assets/index.css
dcalhoun, line 1: I suggest we remove the built assets from the repository and add them to `.gitignore`. Doing so might reduce noise. If you agree, we could do so in a separate PR. 

WDYT?

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -115,6 +118,13 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
             // TOOD: relay to the client
         }
     }
+    
+    private func setupEditor() {
+        let webViewConfiguration = webView.configuration
+        let userContentController = webViewConfiguration.userContentController
+        let editorInitialConfig = getEditorConfiguration()
```
dcalhoun, line 125: Great improvement. Thank you!

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -229,7 +244,7 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         do {
             switch message.type {
             case .onLoaded:
-                injectEditorConfiguration()
+                return
```
dcalhoun: I suggest we remove this now unused event by reverting the relevant portions of https://github.com/wordpress-mobile/GutenbergKit/pull/14/commits/3606c6dbea701cb03f534b62ca0c17a50e543ef6. WDYT?

ReactApp/src/misc/api-fetch-setup.js
```diff
@@ -15,12 +15,12 @@ import { getGBKit } from './store';
  * and preloads some endpoints with mock data for specific components.
  */
 export function initializeApiFetch() {
-	const { siteURL, authToken } = getGBKit();
-	const rootURL = siteURL ? `${siteURL}/wp-json/` : undefined;
+	const { siteAPIURL, authToken, siteID } = getGBKit();
+	const rootURL = siteAPIURL ?? undefined;
 
 	apiFetch.use(apiFetch.createRootURLMiddleware(rootURL));
```
dcalhoun: I introduced `rootURL` solely to avoid passing a partial URL (e.g., `undefined/wp-json` to the API fetch configuration. With this full URL value provided by the host app, I think we should remove `rootURL` and pass `siteAPIURL` directly to `apiFetch.createRootURLMiddleware`. WDYT?

```suggestion

	apiFetch.use(apiFetch.createRootURLMiddleware(siteAPIURL));
```

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -92,7 +94,8 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
 //            assert(Thread.isMainThread)
 //
 //        }.store(in: &cancellables)
-
+        
+        setupEditor()
```
dcalhoun: Small grammatical improvement. Feel free to ignore, if you prefer. 
```suggestion
        setUpEditor()
```

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -115,6 +118,13 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
             // TOOD: relay to the client
         }
     }
+    
+    private func setupEditor() {
```
dcalhoun: Small grammatical improvement. Feel free to ignore, if you prefer. 
```suggestion
    private func setUpEditor() {
```

dcalhoun (commented):
Nice work! I left a few suggestions for consideration. 

The Gallery test case did not pass for me. Both Simple and Atomic WPCOM site's result in a "Sorry, you are not allowed to upload media to this post." error when creating a new, empty post. I used the prototype build on an iPhone SE running iOS 17.6. Do you experience a different outcome? 


<details><summary>Error Screenshot</summary>

Sorry, you are not allowed to upload media to this post error message within the editor

</details>

Sources/GutenbergKit/Gutenberg/assets/index.css
geriux: I think we can do that once we have a process in place to generate these assets in CI like when making a release.

I manually ran `./Scripts/build.sh` to ensure I had the latest changes in a WordPress iOS prototype build.

ReactApp/src/misc/api-fetch-setup.js
geriux: > However, given our long-term hope of merging this into the Gutenberg project, this feels out of place in GutenbergKit. I wonder if we should hoist this middleware definition into the host app. WDYT?

Good idea, it makes sense to have this logic coming from the host app. I'll update it 👍

Sources/GutenbergKit/Sources/EditorViewController.swift
geriux: Sure! Removed in https://github.com/wordpress-mobile/GutenbergKit/pull/15/commits/5be8e6e592ac195256d880ab8d81fc305ff578e6

ReactApp/src/misc/api-fetch-setup.js
geriux: Updated in https://github.com/wordpress-mobile/GutenbergKit/pull/15/commits/7656e760f3a3ae2fc13162e41ddeff90d4a4f725

Sources/GutenbergKit/Sources/EditorViewController.swift
geriux: Updated in https://github.com/wordpress-mobile/GutenbergKit/pull/15/commits/29324052a16a172141e0821c789cc0c0595b7fcd

Sources/GutenbergKit/Sources/EditorViewController.swift
geriux: Updated in https://github.com/wordpress-mobile/GutenbergKit/pull/15/commits/29324052a16a172141e0821c789cc0c0595b7fcd

ReactApp/src/misc/api-fetch-setup.js
geriux: Updated in https://github.com/wordpress-mobile/GutenbergKit/pull/15/commits/23f71cbc44c666e61a588ffb60b83f3d2ee4e573

I also moved the logic for the `Authorization` header to the native side.

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -125,15 +135,30 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
...
         };
+        function apiPathModifier(path) {
+            const siteID = '\(siteIDString)';
+            if (path && siteID.length && !path.includes(`/sites/${siteID}/`)) {
+                return path
+                    .replace('/wp/v2/', `/wp/v2/sites/${siteID}/`)
+                    .replace('/wpcom/v2/', `/wpcom/v2/sites/${siteID}/`)
+                    .replace('/oembed/1.0/', `/oembed/1.0/sites/${siteID}/`);
+            }
+        }
```
dcalhoun: When I proposed passing a function, I didn't consider our current practice of running `JSON.stringify` on these values. Passing a function to that will nullify it. Not an urgent issue given local storage is only used for development purposes, but we might address it at a different time.

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -125,15 +135,30 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
...
-        evaluate("""
+    private func getEditorConfiguration() -> WKUserScript {
+        let siteIDString = siteID?.stringValue ?? ""
+        let authType = siteID != nil ? "Bearer" : "Basic"
+        let jsCode = """
         window.GBKit = {
-            siteURL: '\(siteURL)',
-            authToken: '\(authToken)'
+            siteAPIURL: '\(siteAPIURL)',
+            authToken: '\(authType) \(authToken)',
```
dcalhoun: I suppose `authToken` name is slightly misleading now. We might consider renaming this to `authHeader` in the future.

dcalhoun (approved):
Great work! I left a few suggestions, but we can always implement them later, if desired. Feel free to merge! 🚀

dcalhoun:
> The Gallery test case did not pass for me. Both Simple and Atomic WPCOM site's result in a "Sorry, you are not allowed to upload media to this post." error when creating a new, empty post. I used the prototype build on an iPhone SE running iOS 17.6. Do you experience a different outcome?

Noting the Gallery test case still does not pass for me. I receive the permission error toast message.

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -11,8 +11,9 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
     private let controller = GutenbergEditorController()
     private let timestampInit = CFAbsoluteTimeGetCurrent()
     private let service: EditorService
-    private let siteURL: String
+    private let siteAPIURL: String
     private let authToken: String
+    private let siteID: NSNumber?
```
jkmassel: Can we use a scalar type here? 

Ideally we'd be more specific than `NSNumber` for stuff like this

Related – do we expect that the `siteID` will be optional – I'm thinking that the `siteApiRoot` would have all the information we need, but WDYT?

jkmassel (changes requested):

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -11,8 +11,9 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
     private let controller = GutenbergEditorController()
     private let timestampInit = CFAbsoluteTimeGetCurrent()
     private let service: EditorService
-    private let siteURL: String
+    private let siteAPIURL: String
```
jkmassel: WDYT about `siteApiRoot`?

geriux:
> Noting the Gallery test case still does not pass for me. I receive the permission error toast message.

I found the problem here, it was working for me with a different test account but not with my a8c one. I'll address this issue shortly.

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -274,8 +286,8 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         let editorViewController = EditorViewController(
             content: "",
             service: EditorService(client: MockClient()),
-            siteURL: "",
-            authToken: ""
+            siteApiRoot: "",
+            authHeader: ""
```
dcalhoun: I realized passing these `siteApiRoot`, `authHeader`, and `content` appear unnecessary in this context. We could remove them, if preferred.

Sources/GutenbergKit/Sources/EditorViewController.swift
geriux: Sure! I've updated it in https://github.com/wordpress-mobile/GutenbergKit/commit/f268246891ae3a76be68c72a409e5f5ee64d40c2#diff-8d684f85d6faf91b7cef2e39f371584ec64f48bee75eec04292272b85272a5d4

Sources/GutenbergKit/Sources/EditorViewController.swift
geriux: Thanks for the suggestion, I ended up removing `siteID` in https://github.com/wordpress-mobile/GutenbergKit/commit/f268246891ae3a76be68c72a409e5f5ee64d40c2#diff-8d684f85d6faf91b7cef2e39f371584ec64f48bee75eec04292272b85272a5d4

>Related – do we expect that the siteID will be optional – I'm thinking that the siteApiRoot would have all the information we need, but WDYT?

I tried to unify the information we need in `siteApiRoot` but for sites that use API URLs like `https://public-api.wordpress.com/wp/v2/sites/mysite.wordpress.com` don't work well with the editor's internal API namespaces, where the REST namespace is appended to the root API URL ending up with URLs like `/wp/v2/sites/mysite.wordpress.com/wp/v2/media` or in other cases `/wp/v2/sites/mysite.wordpress.com/oembed/1.0/proxy`.

As a potential solution, I added `siteApiNamespace` that will have a value like `sites/mysite.wordpress.com` (if it's a .com site) so it's easier to format the final URL.

Sources/GutenbergKit/Sources/EditorViewController.swift
geriux: Thanks for the suggestion!

I'm trying a different approach in https://github.com/wordpress-mobile/GutenbergKit/pull/15/commits/f268246891ae3a76be68c72a409e5f5ee64d40c2

Where a new param is added `siteApiNamespace` to be able to format the final URL in a more dynamic way. What do you think about that approach?

geriux:
Thank you both for the feedback, the PR is ready for another review and the [testing build](https://github.com/wordpress-mobile/WordPress-iOS/pull/23499#issuecomment-2289528270) should be working for any .com site.

I've updated the PR description and the test cases.

ReactApp/src/components/Editor.jsx
```diff
@@ -46,9 +46,7 @@ import { postMessage } from '../misc/Helpers';
 let editor = {};
 
 const POST_MOCK = {
-	id: 1,
```
geriux, line 49: I've removed this due to issues with the API. Since this is not an existing post, the API was receiving this as a parameter and throwing a permissions error.

ReactApp/src/misc/api-fetch-setup.js
```diff
@@ -83,24 +80,6 @@ export function initializeApiFetch() {
...
-						raw: '',
-						rendered: '',
-						protected: false,
-						block_version: 0,
-					},
-					excerpt: { raw: '', rendered: '', protected: false },
-					author: LOCAL_AUTHOR_ID,
-					featured_media: 0,
-				},
-			},
```
geriux, line 103: This was not needed anymore and it was causing problems when connecting with the API

ReactApp/src/misc/api-fetch-setup.js
```diff
@@ -110,12 +89,30 @@ function corsMiddleware(options, next) {
...
+		const originalPath = options.path;
+
+		// Determine the initial segments of the path where the namespace should be inserted.
+		// E.g /wp/v2/, /oembed/1.0/
+		const initialPathSegments =
+			originalPath.split('/').splice(1, 2).join('/') + '/';
+		const remainingPath = originalPath.replace(initialPathSegments, '');
+
+		// Update the path with the new namespace.
+		options.path = `${initialPathSegments}${siteApiNamespace}${remainingPath}`;
```
dcalhoun: This approach is far better than my original idea of passing a function from the host app. Nicely done! 👏🏻 

This is undoubtedly subjective, but I find the string splitting a bit difficult to comprehend. I feel a regex replace is simpler, but that may just be my familiarity with it. I recognize not everyone feels comfortable reading regex. 

WDYT? 

```suggestion
		// Insert the API namespace after the first two path segments. 
		options.path = options.path.replace(/^((?:\/[\w.-]+){2})/, `$1/${siteApiNamespace}`);
```

ReactApp/src/misc/api-fetch-setup.js
```diff
@@ -110,12 +89,30 @@ function corsMiddleware(options, next) {
 	return next(options);
 }
 
-function createHeadersMiddleware(authToken) {
+function apiPathModifierMiddleware(options, next) {
```
geriux, line 92: This is a different approach to replacing the URLs for .com-connected sites. I'll update it so it catches any potential errors if the path does not meet the expected format before splitting, etc. But any initial thoughts are welcome!

dcalhoun (approved):
The latest build worked for me. Nicely done! 

I left a few suggestions for you consider, but feel free to disregard them.

ReactApp/src/misc/api-fetch-setup.js
geriux: Oooh I like this, thank you for suggesting that!

Sources/GutenbergKit/Sources/EditorViewController.swift
geriux: Nice catch, updated in https://github.com/wordpress-mobile/GutenbergKit/pull/15/commits/eab7e99ac5a6887ee1970514e867cb1f524bd809

geriux:
Hey @jkmassel 👋  let me know if you have any other feedback or if we could merge this as it is, thank you!

jkmassel (approved):

PR #16, opened 2024-08-21
Use Android Module

jkmassel:
That will allow us to publish the library for Android.

To Test:
1. Run `make local-android-library`
2. Note that there's a build product at `~/.m2/repository/org/wordpress/gutenbergkit/android/unspecified/android-unspecified.aar`
3. Rename that file to end with `.zip`, then extract it
4. Open that folder, note that the assets are present

~~We'll add automatic publishing for each PR commit in a subsequent PR~~ Thanks to @mokagio, this PR now publishes the library too.


Demo-Android/Gutenberg/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
...
+
+project.afterEvaluate {
+    publishing {
+        publications {
+            create<MavenPublication>("maven") {
+                from(components["release"])
+
+                groupId = "org.wordpress.gutenbergkit"
+                artifactId = "android"
+                // version is set by 'publish-to-s3' plugin
```
geriux, line 55: How would this work for release tags?

I'm seeing that for the latest commit it has version `16-commit-hash`. Will the `publish-to-s3` plugin automatically use the `tag` when it is created and the CI tasks run?

Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -29,9 +35,8 @@ class GutenbergView : WebView {
         defStyle
     )
 
+    @SuppressLint("SetJavaScriptEnabled") // Without JavaScript we have no Gutenberg
     fun start() {
-        this.settings.allowFileAccess = true
```
geriux, line 33: I haven't tested it but would this prevent the Editor from loading local files when they're being uploaded?

geriux (approved):
LGTM! I've tested the test case, and it worked as expected.

Regarding the local generated path, I assume that the `unspecified` in the file/folder names will differ from the artifacts generated in CI, correct? Will the CI use the version added by the `publish-to-s3` plugin instead of that?

In the CI steps, we have the "Build React App" step that already generates the assets. In the future, we could improve this process by uploading these artifacts and reusing them for the Android/iOS steps.

I left two other comments that don't block merging this PR.

Thanks for working on this!

Demo-Android/Gutenberg/build.gradle.kts
jkmassel: > Will the publish-to-s3 plugin automatically use the tag when it is created and the CI tasks run
Yep that's exactly right

Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
jkmassel: Possibly – to address this, we'd need to tell the asset loader to allow uploads from a specific directory (wherever we store the file until we're sure it's been uploaded)

geriux: Sounds good!

PR #18, opened 2024-08-21
Add Buildkite pipeline building the package and the integration app

mokagio:
_What it says on the title_ 😄 

Infernal ref pdnsEh-1Ph-p2

Originally, I had steps to `build` and `test` with `xcodebulid`, too, but they fail in CI. Given it seems `make build` is enough to validate, I decided to push those to #19 and have this PR green.


.buildkite/pipeline.yml
```diff
@@ -0,0 +1,16 @@
...
+
+env:
+  IMAGE_ID: $IMAGE_ID
+
+steps:
+  - label: ":swift: Build Swift Package"
+    command: make build_swift_package
+    plugins: &plugins
+      - $CI_TOOLKIT_PLUGIN
+      - $NVM_PLUGIN
```
mokagio: I'm not sure this is necessary given the README instructions and what `make build` does. But... if felt odd to have a package and not build it 😅

Makefile
```diff
@@ -17,4 +17,18 @@ fmt-js: npm-dependencies
...
+		-sdk iphonesimulator \
+		-destination 'OS=17.5,name=iPhone 15 Plus' \
+		| xcbeautify
+
+test_swift_package:
+	xcodebuild test \
+		-scheme GutenbergKit \
+		-sdk iphonesimulator \
+		-destination 'OS=17.5,name=iPhone 15 Plus' \
+		| xcbeautify
```
mokagio: This `xcodebuild` invocation could be DRYed, and we could have a check for `xcbeautify`

But since I'm not sure whether these commands are needed, I haven't gone down that path yet.

jkmassel (approved):
Looks great, thank you!!! 🙇

.buildkite/shared-pipeline-vars
```diff
@@ -0,0 +1,13 @@
...
+
+# This file is `source`'d before calling `buildkite-agent pipeline upload`, and can be used
+# to set up some variables that will be interpolated in the `.yml` pipeline before uploading it.
+
+# The ~> modifier is not currently used, but we check for it just in case
+XCODE_VERSION=$(sed -E -n 's/^(~> )?(.*)/xcode-\2/p' .xcode-version)
+CI_TOOLKIT_PLUGIN_VERSION="3.5.1"
+NVM_PLUGIN_VERSION='0.3.0'
+
+export IMAGE_ID="$XCODE_VERSION"
```
AliSoftware, line 11: 💡 given the variable name, I wonder if it wouldn't be clearer if `XCODE_VERSION` would only contain the version, and `IMAGE_ID` would be the one adding the prefix?

Would also make the `sed` simpler and easier to understand—as the pattern would now just tell "remove the `~> ` prefix if present" without the need for `?`, captures, nor backreferences 🙂

```suggestion
XCODE_VERSION=$(sed 's/^~> *//' .xcode-version)
CI_TOOLKIT_PLUGIN_VERSION="3.5.1"
NVM_PLUGIN_VERSION='0.3.0'

export IMAGE_ID="xcode-$XCODE_VERSION"
```

AliSoftware (approved):

PR #21, opened 2024-08-27
Feature: Add Block Editor Settings and UI tweaks

geriux:
**Related PRs:**

- https://github.com/wordpress-mobile/WordPress-iOS/pull/23499

## Description

This PR adds the following core functionality to the editor:

### Block Editor Settings

It adds the editor's settings data, currently by fetching the block editor settings endpoint [within the editor](https://github.com/wordpress-mobile/GutenbergKit/pull/21/files#diff-11987f198cb9cf62a83bd4322a3e37df80cecd4ce6f20a5a72038a7f93c0905dR26).

This is a temporary approach, as this data should be provided by the host apps.

This is needed to obtain important data like `allowedMimeTypes`, theme `styles`, and other settings required for the editor to function fully.

### Loading Post Content in the Editor

It adds some post type entities to allow a post to be loaded by passing the content without fetching it from the server. This is similar to the approach [we used with the React Native editor](https://github.com/WordPress/gutenberg/blob/20f7dfd8a68b0db7965f12d165f6f628b6c4b68f/packages/editor/src/components/provider/index.native.js#L47).

After adding the entities, it uses `receiveEntityRecords` with the post data from the host apps, if available.

###  Media and Patterns Support

With the block editor settings, media uploads are now enabled, as well as fetching the available patterns and reusable blocks.

### EditorToolbar

Updates this component to remove unused code and improve functionality by reusing Gutenberg components.

### Usage of Experimental APIs

You'll see some usage of experimental components by unlocking some private editor APIs. These are currently used in the web editor, so there shouldn't be any issues.

### Style Updates

This PR also updates several style issues related to the editor, the inserter, and the blocks.

### Native Side

Adds several new parameters to initialize the WebView. Most of these are related to the post data and editor settings.

I also commented on some native code that needs refactoring or complete removal.

## Testing

> [!NOTE]  
> To be able to test the editor please enable the feature flag `Experimental Block Editor` by going to Settings > App Settings > Debug > Feature Flags
>
> Please use the following build:
> - [iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/23499#issuecomment-2289528270)

## Media Library

**Images**

- Open the editor
- Open the inserter
- Go to the Media tab
- **Expect** to see the `Images` list item
- Go to Images
- **Expect** to see images from your media library
- Tap on an image
- **Expect** the image to be added to the post

**Openverse**

- Open the editor
- Open the inserter
- Go to the Media tab
- **Expect** to see the `Openverse` list item
- Go to Openverse
- **Expect** to see images from the Openverse directory
- Tap on an image
- **Expect** the image to be uploaded and added to the post

## Patterns

**Patterns Library**

- Open the editor
- Open the inserter
- Go to the Patterns tab
- **Expect** to see the Patterns library list
- Go to any section
- Tap on any Pattern
- **Expect** the Pattern to be added to the post

**Custom patterns**

- Open the editor
- Open the inserter
- Go to the Patterns tab
- **Expect** to see the `My patterns` item
- Tap on `My Patterns`
- **Expect** to see your patterns
- Tap on one of the patterns
- **Expect** the pattern to be added to the post

## Opening a Post

- Go to the Posts list
- Tap on an existing post
- **Expect** to see the editor with the post tile and content

## Inserter

- Open the editor
- Tap on the (+) button to open the inserter
- Scroll through the list (before some events affected the scroll)
- Tap on any block
- **Expect** to add a block by just tapping once (before you had to tap twice)

**Note**: Some hover styles were removed as well.

## Theme Styles

- Open the app
- Go to Settings > App Settings > Debug > Feature Flags
- Scroll down to the bottom and enable `Experimental Block Editor Styles`
- Open the editor
- **Expect** to see your theme styles in the editor's canvas

**Note:** There will be a bit of delay while the endpoint with the style settings fetches the data. This should be fixed once this data comes from the host app.

### Media

https://github.com/user-attachments/assets/f99e0b10-8097-409c-a7f0-dd1708351805

### Patterns

https://github.com/user-attachments/assets/f14c4436-3445-4f15-8e12-0ff88ef97389

### Opening a Post

https://github.com/user-attachments/assets/fbecc984-b7bb-4ca0-9278-f847fb6aedf2

### Inserter

https://github.com/user-attachments/assets/50c30271-75c2-4b11-b21b-64f79372958e

### Theme Styles

https://github.com/user-attachments/assets/55d9e3c4-662d-4881-92f8-5540c8aafd80


ReactApp/src/components/Editor.jsx
```diff
@@ -40,44 +41,119 @@ import '@wordpress/format-library/build-style/style.css';
...
 	function didChangeBlocks(blocks) {
-		setBlocks(blocks);
-
-		// TODO: this doesn't include everything
-		const isEmpty =
-			blocks.length === 0 ||
-			(blocks[0].name == 'core/paragraph' &&
-				blocks[0].attributes.content.trim() === '');
-		onBlocksChanged(isEmpty);
+		// setBlocks(blocks);
```
geriux, line 116: I'll work on a different PR to provide post updates to the host apps

ReactApp/src/components/hooks/use-editor-styles.js
```diff
@@ -0,0 +1,57 @@
...
+
+// eslint-disable-next-line react-refresh/only-export-components
+export const { lock, unlock } =
+	__dangerousOptInToUnstableAPIsOnlyForCoreModules(
+		'I acknowledge private features are not for use in themes or plugins and doing so will break in the next version of WordPress.',
+		'@wordpress/editor'
+	);
+const { getLayoutStyles } = unlock(blockEditorPrivateApis);
+
+// This should be exported from Core so no reimplementation is needed.
```
geriux, line 15: Or when this is part of Gutenberg and we can import it directly

ReactApp/src/components/Editor.jsx
```diff
@@ -40,44 +41,119 @@ import '@wordpress/format-library/build-style/style.css';
...
 	const [blocks, setBlocks] = useState([]);
 	const [registeredBlocks] = useState([]);
 	const [_isCodeEditorEnabled, setCodeEditorEnabled] = useState(false);
 	const titleRef = useRef();
+	const { addEntities, receiveEntityRecords } = useDispatch(coreStore);
 	const { setupEditor } = useDispatch(editorStore);
 
 	useEffect(() => {
-		setupEditor(post, [], {});
+		window.editor = editor;
```
dcalhoun, line 66: The `editor` assignment feels like unnecessary indirection to me. We might consider removing it and referencing `window.editor` elsewhere directly instead. 

I recognize `editor` was present before the proposed changes. Merely capturing my thought, we can address it later, if desired.

ReactApp/src/components/EditorToolbar.jsx
```diff
@@ -6,69 +6,69 @@ import {
...
+				<Inserter />
+
+				{isSelected && (
+					<div className="gbkit-editor-toolbar_toolbar-group">
+						<ToolbarButton
+							title={__('Open Settings')}
+							icon={cog}
+							onClick={openSettings}
+							className="gbkit-editor-toolbar_settings_icon"
+						></ToolbarButton>
```
dcalhoun: For the pursuit of brevity, we might consider self closing this element, given it has no children. 
```suggestion
						/>
```

ReactApp/src/components/hooks/use-editor-styles.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+import { store as editorStore } from '@wordpress/editor';
+import { useSelect } from '@wordpress/data';
+import { privateApis as blockEditorPrivateApis } from '@wordpress/block-editor';
+import { store as editPostStore } from '@wordpress/edit-post';
+import { useMemo } from '@wordpress/element';
+import { __dangerousOptInToUnstableAPIsOnlyForCoreModules } from '@wordpress/private-apis';
+
+// eslint-disable-next-line react-refresh/only-export-components
+export const { lock, unlock } =
```
dcalhoun: To preserve hot module reloading, we might consider adhering to this lint warning and relocating the export to a separate file.

ReactApp/src/index.css
```diff
@@ -93,36 +104,42 @@ h4 {
...
-	margin-left: 4px;
+	font-size: 34px !important;
```
dcalhoun, line 124: There are a few[ "magic numbers"](https://en.wikipedia.org/wiki/Magic_number_(programming)) in the styles currently. In a separate PR later, I posit we should make a pointed effort to explain them through variable names, design systems, code comments, etc.

ReactApp/src/index.css
```diff
@@ -15,23 +15,34 @@
...
 	line-height: 1.55;
 }
 
 /* Theme (for no-theme design) */
+* {
+	-webkit-tap-highlight-color: transparent;
+}
+
+.components-button:focus:not(:disabled) {
+	outline: 0 !important;
```
dcalhoun, line 28: At some point in the future, we should take time to consider how we might best coexist with core's styles. I.e., the use of `!important` concerns me, it is likely worthy of exploring a more sustainable override approach. 

Not a task for this PR, merely sharing the thought. Related to https://github.com/Automattic/dotcom-forge/issues/8678.

ReactApp/src/main.jsx
```diff
@@ -2,20 +2,51 @@
...
 window.GBKit = getGBKit();
 initializeApiFetch();
 
-createRoot(document.getElementById('root')).render(
-	<StrictMode>
-		<App />
-	</StrictMode>
-);
+function initializeEditor() {
+	const { themeStyles, siteApiRoot } = window.GBKit;
```
dcalhoun: Given the close proximity to the `window.GBKit` assignment, it likely has little impact, but I suggest we always rely upon the `getGBKit` utility. The consistency in approach might mitigate unexpected, cryptic errors elsewhere if `window.GBKit` is referenced directly. WDYT? 

```suggestion
	const { themeStyles, siteApiRoot } = getGBKit();
```

ReactApp/src/misc/Helpers.js
```diff
@@ -37,6 +37,22 @@ export function showBlockPicker() {
 	}
 }
 
+export function getPost() {
+	const currentPost = window?.initialPost;
```
dcalhoun: I envisioned all globals residing under a single `GBKit` namespace to avoid potential collisions. 

```js
const { initialPost: currentPost } = getGBKit();
```

Do you feel that is valuable?

ReactApp/src/misc/Helpers.js
```diff
@@ -37,6 +37,22 @@ export function showBlockPicker() {
...
+	if (currentPost) {
+		return {
+			id: currentPost.id,
+			title: { raw: decodeURIComponent(currentPost.title) },
+			content: { raw: decodeURIComponent(currentPost.content) },
+			type: currentPost?.type ?? 'post',
```
dcalhoun: Given the wrapping conditional, this optional chain is unnecessary, correct? Should we instead rely upon the OR operator? 
```suggestion
			type: currentPost.type || 'post',
```

ReactApp/src/misc/api-fetch-setup.js
```diff
@@ -92,9 +92,12 @@ function corsMiddleware(options, next) {
 function apiPathModifierMiddleware(options, next) {
 	const { siteApiNamespace } = getGBKit();
 
-	if (siteApiNamespace) {
-		// Insert the API namespace after the first two path segments. 
-		options.path = options.path.replace(/^((?:\/[\w.-]+){2})/, `$1/${siteApiNamespace}`);
+	if (siteApiNamespace && !options.path.includes(siteApiNamespace)) {
```
dcalhoun, line 95: In what context is the namespace already present?

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -36,8 +40,12 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
     private var cancellables: [AnyCancellable] = []
 
     /// Initalizes the editor with the initial content (Gutenberg).
-    public init(content: String = "", service: EditorService, siteApiRoot: String = "", siteApiNamespace: String = "", authHeader: String = "") {
+    public init(id: Int? = nil, type: String = "", title: String = "", content: String = "", service: EditorService, themeStyles: Bool = false, siteApiRoot: String = "", siteApiNamespace: String = "", authHeader: String = "") {
```
dcalhoun, line 43: I feel we are approaching an argument count threshold where we should consider an alternative pattern — builder pattern, parameter object, property setters, etc. We can address this separately, if desired.

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -136,15 +144,30 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
     }
 
     private func getEditorConfiguration() -> WKUserScript {
-        let jsCode = """
+        let escapedTitle = initialTitle.addingPercentEncoding(withAllowedCharacters: .alphanumerics)!
+        let escapedContent = _initialRawContent.addingPercentEncoding(withAllowedCharacters: .alphanumerics)!
+        let postId = id != nil ? id! : nil
+        let hasThemeStylesEnabled = themeStyles != nil ? themeStyles! : false
```
dcalhoun: I believe this might be simplified. WDYT?  
```suggestion
        let postId = id ?? nil
        let hasThemeStylesEnabled = themeStyles ?? false
```

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -136,15 +144,30 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
...
             siteApiRoot: '\(siteApiRoot)',
             siteApiNamespace: '\(siteApiNamespace)',
-            authHeader: '\(authHeader)'
+            authHeader: '\(authHeader)',
+            themeStyles: \(hasThemeStylesEnabled)
         };
         localStorage.setItem('GBKit', JSON.stringify(window.GBKit));
         "done";
         """
+        if let postId = id {
```
dcalhoun: Does `postId` serve a specific purpose or could we rely upon `id` directly?

dcalhoun (approved):
Great work! This is a lot of crucial functionality that is incredibly valuable to merge. 

Testing went relatively well for me. The two things I noted: 

1. Selecting media from Media Library does not work for me; tapping the button doesn't display anything. 
2. Media uploaded to private sites results in broken images within the editor. 

The first might be worth debugging now, the latter we could address later. 

I left several comments that I consider subjective and refactors that might be nice to have. However, I lean towards merging this sooner rather than later, as it overlaps with my own work and unlocks a lot of capabilities. We could follow up with improvements later.

Sources/GutenbergKit/Sources/EditorViewController.swift
dcalhoun: Actually, we likely do not even need `let postId = id ?? nil`. We could just pass `id`.

ReactApp/src/index.css
geriux: I agree! To be honest this whole file needs to be updated and possibly split. I think we can work on that separately addressing other issues like the `!important` usage mentioned in this [comment](https://github.com/wordpress-mobile/GutenbergKit/pull/21#discussion_r1744350860).

ReactApp/src/index.css
geriux: Definitely! I think once we have a working prototype with the needed features we should spend some time on that. Take some time to clean up and improve what we have now.

ReactApp/src/misc/Helpers.js
geriux: Let's try it! It'd be a good idea to have all of the editor params in the same namespace/localStorage 👍

ReactApp/src/misc/api-fetch-setup.js
geriux: I'll test it again and report back, but I remember I started to have issues once we merged this code and I was playing around with some endpoints, where it was appending the namespace at least twice 😅

Sources/GutenbergKit/Sources/EditorViewController.swift
geriux: Yeah, I thought about it when I kept adding more 😅 I think we could do it separately so this PR doesn't get bigger than it already is.

ReactApp/src/components/Editor.jsx
geriux: It looks like `window.editor` is not used at all, so we could just remove it.

geriux: Oh ok! I see we are setting some callbacks like `editor.setContent`. These could be also set once and use `window.editor` directly if needed.

Let's do this in a separate PR that can improve how the editor communicates with the host apps.

ReactApp/src/misc/api-fetch-setup.js
geriux: By just having the API credentials set, as soon as the editor loads with these changes I'm seeing failures to:

`https://public-api.wordpress.com/wp/v2/sites/SITE_ID/sites/SITE_ID/blocks?per_page=100&_locale=user`

It might be an issue related to how the reusable blocks are fetched, but I'm not sure if it would happen with other endpoints. I'd say to include this check. What do you think?

dcalhoun: Interesting. I'm curious as to why, but I can research further later. Yes, I agree including the check makes sense.

ReactApp/src/components/hooks/use-editor-styles.js
geriux: Updated in https://github.com/wordpress-mobile/GutenbergKit/pull/21/commits/b6560a8b2b97622ac67033ff51f1d37b3c4e3df3

ReactApp/src/main.jsx
geriux: Updated in https://github.com/wordpress-mobile/GutenbergKit/pull/21/commits/c4426ba888c964e4122551a27638ee36a3adbf8f

ReactApp/src/misc/Helpers.js
geriux: Updated in https://github.com/wordpress-mobile/GutenbergKit/pull/21/commits/54aedfb4541fb8465a6b4caa2c2c58f1d86334a3

ReactApp/src/misc/Helpers.js
geriux: Updated in https://github.com/wordpress-mobile/GutenbergKit/pull/21/commits/64665da320dd26e6a8999539340b1d73794b1442 thanks!

Sources/GutenbergKit/Sources/EditorViewController.swift
geriux: I've updated it in this commit https://github.com/wordpress-mobile/GutenbergKit/pull/21/commits/54aedfb4541fb8465a6b4caa2c2c58f1d86334a3

Sources/GutenbergKit/Sources/EditorViewController.swift
geriux: I've changed the approach in https://github.com/wordpress-mobile/GutenbergKit/pull/21/commits/54aedfb4541fb8465a6b4caa2c2c58f1d86334a3 thanks!

geriux:
> Great work! This is a lot of crucial functionality that is incredibly valuable to merge.

Thank you so much for the review and testing! 🚀 

> Testing went relatively well for me. The two things I noted:
> 
> 1. Selecting media from Media Library does not work for me; tapping the button doesn't display anything.

I forgot to mention this. Unfortunately, there's no built-in support for the Media Library within the Web Editor as it relies on the WordPress Media Library functionality.

I think we could have a hybrid approach to open the native Media Library picker. I'll create a GitHub issue for this.

> 2. Media uploaded to private sites results in broken images within the editor.

I'll create a ticket for this as well. I think currently this is solved by using the cookies which we could try to set to the WebView but it needs more investigation as using cookies might change anytime.

dcalhoun:
> I forgot to mention this. Unfortunately, there's no built-in support for the Media Library within the Web Editor as it relies on the WordPress Media Library functionality.

Aha! My confusion is a result of me reading the testing instructions too quickly and _not_ watching your example videos. 😄 I now understand the media tests were solely for the inserter _Media_ tab. 

> I'll create a ticket for this [broken media on private sites] as well. I think currently this is solved by using the cookies which we could try to set to the WebView but it needs more investigation as using cookies might change anytime.

Good to know. My understanding is that cookie authentication is a non-starter for this product moving forward. We'll likely need to further consider approaches for fetching private media. 🤔

geriux:
> Good to know. My understanding is that cookie authentication is a non-starter for this product moving forward. We'll likely need to further consider approaches for fetching private media. 🤔

Yeah, let's work on that next 👍

Sources/GutenbergKit/Sources/EditorViewController.swift
dcalhoun: Noting that we could still simplify `hasThemeStylesEnabled`, if desired.

geriux:
I'll merge this now, if there's more feedback I can happily address it in new PRs. Thanks again for the review!

Sources/GutenbergKit/Sources/EditorViewController.swift
dcalhoun: After reviewing your changes here, I now understand your original let guard clause. Feel free to revert the changes for these specific lines if you prefer the previous implementation. Either approach is fine with me.

Sources/GutenbergKit/Sources/EditorViewController.swift
geriux: Oh I missed that, I'll test that suggestion. I was trying to avoid setting a value wrapped with `Optional()`. I'll check if that handles it.

Sources/GutenbergKit/Sources/EditorViewController.swift
geriux: I think with the latest changes its ok because we are checking `id != nil` before setting the post data, and then unwrapping the value with `id!`

Sources/GutenbergKit/Sources/EditorViewController.swift
dcalhoun: Gotcha. It is very possible my suggestion is invalid. Feel free to disregard.

PR #22, opened 2024-09-06
feat: Third-party block support

dcalhoun:
## Description

Add experimental editor relying upon editor assets fetched from the remote WordPress site, enabling the ability to utilize third-party blocks. This experimental approach serves as a prototype and relies upon a new endpoint provided by a [plugin](https://github.com/wordpress-mobile/block-editor-assets-endpoint), as landing the required endpoint in Gutenberg/core [may require](https://github.com/WordPress/gutenberg/pull/64517#issuecomment-2307019003) a fair amount of effort and time.

<details><summary>Screen recording</summary>

<video src="https://github.com/user-attachments/assets/a00f2d29-db4f-4227-8ba2-a44f3c2f7455" />

</details>

Relates to:

* https://github.com/Automattic/dotcom-forge/issues/8680
* https://github.com/Automattic/dotcom-forge/issues/8682
* https://github.com/wordpress-mobile/WordPress-iOS/pull/23563.

## Testing Instructions

> [!TIP]
> Use the prototype build from the sibling [WP-iOS PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/23563#issuecomment-2339095587). 

> [!Important]
> This remote editor currently only supports self-hosted sites when authenticating with a site's admin username and [application password](https://developer.wordpress.org/rest-api/using-the-rest-api/authentication/#basic-authentication-with-application-passwords). It also requires installing and activating the [`block-editor-assets-endpoint`](https://github.com/wordpress-mobile/block-editor-assets-endpoint/releases/tag/v0.0.1-alpha.2) plugin for your self-hosted site. 

**Third-party blocks function for self-hosted sites**
1. Install and activate the [`block-editor-assets-endpoint`](https://github.com/wordpress-mobile/block-editor-assets-endpoint/releases/tag/v0.0.1-alpha.2) plugin on your self-hosted site. 
1. In the app, enable the "Experimental Block Editor" and "Experimental Block Editor Plugins" feature flags. 
2. Open a self-hosed site's post containing third-party blocks. 
3. The blocks should render as expected, editing the blocks should also succeed. 

**WPCOM-authenticated sites utilize the local editor**
1. Enable the "Experimental Block Editor" and "Experimental Block Editor Plugins" feature flags. 
2. Open a WPCOM-authenticated site's post containing third-party blocks. 
3. The blocks should render "unsupported" messages.


ReactApp/src/components/Editor.jsx
```diff
@@ -64,10 +64,9 @@ function Editor({ post }) {
 
 	useEffect(() => {
 		window.editor = editor;
-		addEntities(postTypeEntities);
 		receiveEntityRecords('postType', post.type, post);
 
-		setupEditor(post, [], []);
+		setupEditor(post, {});
```
dcalhoun, line 69: Updated to mirror the argument types found in core. Second parameter is an object, third is optional.

ReactApp/src/components/Editor.jsx
```diff
@@ -174,6 +171,7 @@ function Editor({ post }) {
 				onInput={onBlockEditorInput}
 				onChange={onBlockEditorChange}
 				settings={settings}
+				useSubRegistry={false}
```
dcalhoun, line 174: Required to ensure a singleton store is used for the block editor. Otherwise, plugins querying for blocks will find a new, empty store.

ReactApp/src/misc/api-fetch-setup.js
```diff
@@ -86,17 +86,29 @@ export function initializeApiFetch() {
...
+
 	return next(options);
 }
 
 function apiPathModifierMiddleware(options, next) {
 	const { siteApiNamespace } = getGBKit();
 
-	if (siteApiNamespace && !options.path.includes(siteApiNamespace)) {
+	if (
+		options.path &&
```
dcalhoun, line 104: It is possible to query with `options.url` rather than `options.path`, so we need to guard for that scenario.

ReactApp/src/remote.jsx
```diff
@@ -0,0 +1,135 @@
...
+		const { themeStyles, siteURL, siteApiRoot } = getGBKit();
+		const { styles, scripts } = await apiFetch({
+			url: `${siteURL}/wp-json/__experimental/wp-block-editor/v1/editor-assets`,
+		});
+		await loadAssets([...styles, ...scripts].join(''));
+
+		// Utilize remote-loaded globals rather than importing local modules
+		const { dispatch } = window.wp.data;
+		const { store: editorStore } = window.wp.editor;
+		const { store: preferencesStore } = window.wp.preferences;
```
dcalhoun, line 30: Referencing global rather than imports simplifies the Vite build config, avoiding the need to externalize this libraries, but somehow not reference them until _after_ the editor assets load.

ReactApp/src/remote.jsx
```diff
@@ -0,0 +1,135 @@
...
+	 * Load each asset in order, as they may depend upon an earlier loaded script.
+	 * Stylesheets and Inline Scripts will resolve immediately upon insertion.
+	 */
+	for (const newAsset of newAssets) {
+		await loadAsset(newAsset);
+	}
+}
+
+// Discard remote copies of localy-sourced Gutenberg packages to avoid conflicts
+const localGutenbergPackages = ['api-fetch'];
```
dcalhoun, line 80: We load the `@wordpress/api-fetch` package locally to configure its options, therefore we do not load the remote copy.

ReactApp/vite.config.js
```diff
@@ -1,18 +1,9 @@
...
 	base: '',
 	build: {
-		// Removes hashes from the filenames
-		rollupOptions: {
-			output: {
-				entryFileNames: `assets/[name].js`,
-				chunkFileNames: `assets/[name].js`,
-				assetFileNames: `assets/[name].[ext]`,
-			},
-		},
```
dcalhoun, line 15: Without filename hashes, it is near impossible to invalidate the asset cache.

ReactApp/vite.config.remote.js
```diff
@@ -0,0 +1,79 @@
...
+function externalize(id) {
+	const externalDefinition = defaultRequestToExternal(id);
+	return (
+		!!externalDefinition &&
+		!id.endsWith('.css') &&
+		!['apiFetch', 'i18n', 'url', 'hooks'].includes(
+			externalDefinition[externalDefinition.length - 1]
+		)
+	);
+}
```
dcalhoun, line 26: Externalizes all WordPress package imports that have external definitions, are not CSS, and are not the packages we use locally in our editor setup.

dcalhoun:
I discovered that if a Jetpack account connection is not active on the site, this prototype fails. It appears the `edit-post` script is not enqueued by default in the API endpoint, but declared as a dependency by Jetpack. 

So, to see the prototype work currently, one has to:

1. Install and activate the Jetpack plugin. 
2. Connect Jetpack to a WPCOM account. 
3. _Not_ use the WPCOM account to add the site to the mobile, but rely upon the site's username and application password instead. 

I'll seek a solution to this issue.

dcalhoun:
> I discovered that if a Jetpack account connection is not active on the site, this prototype fails. It appears the `edit-post` script is not enqueued by default in the API endpoint, but declared as a dependency by Jetpack.

I resolved this in [`block-editor-assets-endpoint@v0.0.1-alpha.2`](https://github.com/wordpress-mobile/block-editor-assets-endpoint/releases/tag/v0.0.1-alpha.2). Installing the latest release should allow remote editor loading with or without a Jetpack connection.

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -137,6 +141,9 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
     private func loadEditor() {
         if let editorURL = editorURL ?? ProcessInfo.processInfo.environment["GUTENBERG_EDITOR_URL"].flatMap(URL.init) {
             webView.load(URLRequest(url: editorURL))
+        } else if plugins,
+                  let editorURL = URL(string: "\(siteURL)/wp-content/plugins/block-editor-assets-endpoint/editor/remote.html") {
```
geriux: I'm still going through the PR and changes to gain the full context but can you explain a bit more why we are loading the remote `remote.html` file from the plugin? If we have this locally, can't we load the local version?

dcalhoun: I took the approach of loading a remote `remote.html` as I encountered CORS errors fetching the API endpoint. Hosting the `remote.html` file on the site's domain nullifies CORS. 

My understanding is that WordPress defaults to a lax [CORS policy](https://developer.wordpress.org/rest-api/frequently-asked-questions/#why-is-the-rest-api-not-verifying-the-incoming-origin-header-does-this-expose-my-site-to-csrf-attacks). So, we should, in theory, be able to perform the API endpoint request. 

The CORS errors were seemingly sporadic. The errors might've related to a given site's host, as I imagine different hosts could tighten CORS policies. I may have also been fighting unexpected caching that confused my testing. 

Nonetheless, I'll retest whether CORS is a legitimate issue for this feature.

dcalhoun: From testing now, there does not appear to be CORS errors when leveraging the local `remote.html` file. I updated the code to use the local copy in https://github.com/wordpress-mobile/GutenbergKit/pull/22/commits/179c3e7868212faca4b55b7c94998ed0be796c9f.

geriux (approved):
LGTM! Great work, David! 👏

I couldn’t get it to run locally, but I was able to use the testing build. ✅

Just noting that it sometimes takes a bit to load, but I know that caching or reducing loading time isn't the focus of this PR.

Once this lands we could add the new functionality in https://github.com/wordpress-mobile/GutenbergKit/pull/23

PR #23, opened 2024-09-10
Feature: Android - Add initial Editor params

geriux:
**Related PRs:**
- https://github.com/wordpress-mobile/WordPress-Android/pull/21223

This PR adds the initial editor parameters for Android, following the approach we used on iOS.

It enables the `domStorageEnabled` feature for the WebView, allowing the use of the `localStorage` API.

Unfortunately, on Android, there is nothing similar to iOS's `addUserScript` to inject the editor's configuration. Therefore, we inject the configuration as soon as we receive the first request from the WebView. In this case, it happens when the `index.html` file is requested, triggering a call to `getEditorConfiguration` to inject the configuration.

### DIFF
--- a/ReactApp/src/components/Editor.jsx
+++ b/ReactApp/src/components/Editor.jsx
@@ -8,7 +8,11 @@ import {
 	privateApis as blockEditorPrivateApis,
 } from '@wordpress/block-editor';
 import { Popover } from '@wordpress/components';
-import { getBlockTypes, unregisterBlockType } from '@wordpress/blocks';
+import {
+	getBlockTypes,
+	unregisterBlockType,
+	__unstableSerializeAndClean,
+} from '@wordpress/blocks';
 import { registerCoreBlocks } from '@wordpress/block-library';
 import { parse, serialize } from '@wordpress/blocks';
 import {
@@ -54,8 +58,12 @@ function Editor({ post }) {
 	const [blocks, setBlocks] = useState([]);
 	const [_isCodeEditorEnabled, setCodeEditorEnabled] = useState(false);
 	const titleRef = useRef();
-	const { addEntities, editEntityRecord, receiveEntityRecords } =
-		useDispatch(coreStore);
+	const {
+		addEntities,
+		editEntityRecord,
+		getEditedEntityRecord,
+		receiveEntityRecords,
+	} = useDispatch(coreStore);
 	const { getEditedPostAttribute, getEditedPostContent } =
 		useSelect(editorStore);
 	const { setEditedPost } = unlock(useDispatch(editorStore));
@@ -140,14 +148,26 @@ function Editor({ post }) {
 		return serialize(blocks); // It's used for tracking changes
 	};
 
+	editor.getContent = () => {
+		const record = getEditedEntityRecord('postType', post.type, post.id);
+		if (record) {
+			if (typeof record.content === 'function') {
+				return record.content(record);
+			} else if (record.blocks) {
+				return __unstableSerializeAndClean(record.blocks);
+			} else if (record.content) {
+				return record.content;
+			}
+		}
+	};
+
 	editor.getTitleAndContent = () => {
 		return {
 			title: getEditedPostAttribute('title'),
 			content: getEditedPostContent(),
+			contentNew: editor.getContent(),
 		};
 	};
-
-	editor.getContent = () => serialize(blocks);
 
 	editor.setCodeEditorEnabled = (enabled) => setCodeEditorEnabled(enabled);
 
--- a/ReactApp/src/components/Editor.jsx
+++ b/ReactApp/src/components/Editor.jsx
@@ -55,18 +55,14 @@ let editor = {};
 const { ExperimentalBlockCanvas: BlockCanvas } = unlock(blockEditorPrivateApis);
 
 function Editor({ post }) {
-	const [blocks, setBlocks] = useState([]);
 	const [_isCodeEditorEnabled, setCodeEditorEnabled] = useState(false);
 	const titleRef = useRef();
-	const {
-		addEntities,
-		editEntityRecord,
-		getEditedEntityRecord,
-		receiveEntityRecords,
-	} = useDispatch(coreStore);
+	const { addEntities, editEntityRecord, receiveEntityRecords } =
+		useDispatch(coreStore);
+	const { setEditedPost } = useDispatch(editorStore);
 	const { getEditedPostAttribute, getEditedPostContent } =
 		useSelect(editorStore);
-	const { setEditedPost } = unlock(useDispatch(editorStore));
+	const { getEditedEntityRecord } = useSelect(coreStore);
 
 	useEffect(() => {
 		window.editor = editor;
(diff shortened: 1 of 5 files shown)


ReactApp/src/components/Editor.jsx
```diff
@@ -59,15 +59,15 @@ function Editor({ post }) {
...
+	const { addEntities, receiveEntityRecords } = useDispatch(coreStore);
 
 	useEffect(() => {
 		window.editor = editor;
 		addEntities(postTypeEntities);
-		receiveEntityRecords('postType', post.type, post);
+		receiveEntityRecords('postType', post.type, post, undefined, true);
 
-		setupEditor(post, [], []);
+		setupEditor(post, {});
```
geriux: This change was done in https://github.com/wordpress-mobile/GutenbergKit/pull/22 I'll update with `trunk` once it is merged.

ReactApp/src/components/Editor.jsx
```diff
@@ -168,25 +174,31 @@ function Editor({ post }) {
...
-			</BlockEditorProvider>
-		</div>
+		hasLoadedPost && (
+			<div className="editor__container">
+				<BlockEditorProvider
+					value={postBlocks}
+					onInput={onBlockEditorInput}
+					onChange={onBlockEditorChange}
+					settings={settings}
+					useSubRegistry={false}
```
geriux: This change was done in https://github.com/wordpress-mobile/GutenbergKit/pull/22 I'll update with `trunk` once it is merged.

nbradbury:
@geriux I'm confused by the testing steps in this PR. It states to use #21223 to test this PR?

dcalhoun:
> @geriux I'm confused by the testing steps in this PR. It states to use #21223 to test this PR?

@nbradbury because Gutenberg — particularly the mobile variant — is a "feature" that is largely only accessible in the WordPress mobile app, it is common practice to have multiple PRs for landing Gutenberg features. One PR for the changes to Gutenberg, one for integrating those changes into the host app (the WordPress mobile app). Because of this, testing instructions are often only listed in one of the two related PRs and a prototype build on the host app PR is utilized for testing. 

I.e., performing the testing instructions on https://github.com/wordpress-mobile/WordPress-Android/pull/21223 using the prototype build found in https://github.com/wordpress-mobile/WordPress-Android/pull/21223#issuecomment-2341311367 should suffice.

nbradbury:
@dcalhoun Thanks for the clarification! Apparently COVID left me stupid because I overlooked that this is a Gutenberg PR, mistakenly thinking it was a WPAndroid one :)

nbradbury:
@geriux I deleted my comments on that WPAndroid PR and will re-test using the prototype build.

nbradbury:
@geriux I tested using the prototype build and it's looking very promising, but I did run into a few issues:

* When I try to add an image from my device I get an error toast stating, "Cannot read properties of undefined" and the image doesn't get added.
* When I share an image from Google Photos with WPAndroid, it correctly creates a new post but the image doesn't get added.
* Tapping to add an image from my media library does nothing. Is that supposed to be working?
* The keyboard randomly loses focus. For example, I type "/" add choose to add a quote and although the soft keyboard remains visible, typing does nothing until I tap the editor.
* Creating a page from a layout doesn't always work. Sometimes the layout isn't used and only an empty image block appears.

ReactApp/src/components/Editor.jsx
```diff
@@ -101,28 +108,46 @@ function Editor({ post }) {
...
+			) {
+				onEditorContentChanged();
+				postTitleRef.current = title;
+				postContentRef.current = content;
+			}
+		});
+
+		return () => {
+			unsubscribe();
+		};
```
dcalhoun: How are content changes communicated to the host app from the React Native version? My impression is that we do so via the editor's autosave event — editor triggers autosave → communicate autosave to host app → host app requests title/content over the bridge. I presume that flow occur far less frequently than the store change. 

Am I correctly interpreting this code updates the host app on every store change? Are there potential performance implications to this approach?

dcalhoun: If we retain the store subscription, we might simplify it with either removing the unnecessary anonymous function or returning subscribe altogether. WDYT? 

<details><summary>Remove unnecessary anonymous function</summary>

```suggestion
		const unsubscribe = subscribe(() => {
			const { title, content } = editor.getTitleAndContent();
			if (
				title !== postTitleRef.current ||
				content !== postContentRef.current
			) {
				onEditorContentChanged();
				postTitleRef.current = title;
				postContentRef.current = content;
			}
		});

		return unsubscribe;
		};
```

</details>

<details><summary>Return the subscribe return value</summary>

```suggestion
		return subscribe(() => {
			const { title, content } = editor.getTitleAndContent();
			if (
				title !== postTitleRef.current ||
				content !== postContentRef.current
			) {
				onEditorContentChanged();
				postTitleRef.current = title;
				postContentRef.current = content;
			}
		});
```

</details>

ReactApp/src/misc/api-fetch-setup.js
```diff
@@ -141,3 +142,27 @@ function filterEndpointsMiddleware(options, next) {
...
+			options.body.forEach((value, key) => {
+				if (key === 'post' && value === '-1') {
+					return;
+				}
+				newFormData.append(key, value);
+			});
+
+			options.body = newFormData;
+		}
+	}
```
dcalhoun: I didn't test this suggested code, but would it make sense to short this to avoid nesting and iterating? 

```suggestion
	if (
		options.path.startsWith('/wp/v2/media') &&
		options.method === 'POST' &&
		options.body instanceof FormData &&
		options.body.get('post') === '-1'
	) {
		options.body.delete('post');
	}

```

dcalhoun (commented):
This was quite the undertaking! Thank you for getting this operating. 👏🏻 

Similar to @nbradbury, I encountered a few issues. 

* Focus is lost while typing — it appears as though text entry is captured and then inserted once focused is regained. I presume this may relate to the [intentional blurring](https://github.com/wordpress-mobile/GutenbergKit/blob/7f3f9409de9531d7ed8d69e63365dd7dfe9bafcd/Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt#L254C29-L258) that is present currently. 
* There is often an unexpected cache of post content — e.g., a newly opened post includes content from a previous post or a new page layout is absent due to cached empty content. 
* I was only infrequently able to successfully upload media, and only for the Gallery block.  For all other media blocks — Image, File, Audio, etc — tapping _Upload_ did not display any UI for selecting media. With the Gallery block, it would also not respond to tapping the _Upload_ button or fail to upload with an invalid JSON response error. Tapping the _Upload_ button when replacing media within a Gallery also did not respond to taps. 

I tested the [pr21223-57ed1f5](https://github.com/wordpress-mobile/WordPress-Android/pull/21223#issuecomment-2341311367) prototype build on a Samsung Galaxy S20 running Android 13. 

<details><summary>Lost focus</summary>


https://github.com/user-attachments/assets/92f5488b-e7be-4960-8168-fcadeac1125c

</details>

<details><summary>Cached content</summary>

https://github.com/user-attachments/assets/1cf1039d-44cc-4cd2-a37d-5835ca1ccbce

</details>

<details><summary>Gallery upload error</summary>

android-upload-error

</details>

jkmassel (approved):
Adding an approval here to unblock things, but @geriux will address issues if needed!

geriux:
Hey there @nbradbury @dcalhoun thank you for testing and the feedback! 

I'm working to work on all of the feedback in a separate PR so we can unblock some changes for iOS.

Once this is merged I'll focus on the caching and loading issues on Android.

PR #24, opened 2024-09-16
Fix - EditorProvider and private APIs

geriux:
This PR addresses an issue where a removed experimental API breaks the editor in the latest Gutenberg version.

It now uses the `EditorProvider` exported from `@wordpress/editor` However, the main issue is that the media library tab no longer loads because the `inserterMediaCategories` data is missing.

This occurs because the data is marked as private, and the `EditorProvider` component removes it using the [BlockEditorProvider](https://github.com/WordPress/gutenberg/blob/afbb659ed61a07e0b8e98cdd57e64363d1d62910/packages/editor/src/components/provider/index.js#L394), where the prop `stripExperimentalSettings` is [set to](https://github.com/WordPress/gutenberg/blob/afbb659ed61a07e0b8e98cdd57e64363d1d62910/packages/block-editor/src/components/provider/index.js#L60) true.

In Gutenberg, the editor uses `ExperimentalEditorProvider` with `ExperimentalBlockEditorProvider`, but these are not publicly exported.

For now, we can merge this to have a working editor and then look for a long-term solution, likely by attempting to export the other provider as a private API until it is no longer marked as experimental.

Additionally, this PR updates the WordPress packages to their latest versions, adding support for multi-select.

A new middleware is introduced in the API Fetch setup to avoid fetching post content when the editor loads. We are already passing the data, and fetching it again can slow down the editor on slower connections and cause issues where content might flicker or reload.

Regarding post loading, for empty posts, a post ID is expected when the editor loads. For drafts, the Web editor generates a post ID by auto-saving. Since this functionality is not available, a `-1` post ID is set to ensure the editor loads correctly.

### To test

> [!NOTE]  
> You can use the following [WordPress iOS build](https://github.com/wordpress-mobile/WordPress-iOS/pull/23591#issuecomment-2346606120) to test these changes
> **Precondition:** Enable the feature flag `Experimental Block Editor`

### Test case 1 - The editor loads an existing post

- Open the app
- Go to Posts
- Tap on an existing post
- **Expect** the post to load with both the title and the content

### Test case 2 - The editor loads with a new post

- Open the app
- Open the editor
- **Expect** to see the title and content placeholders


ReactApp/src/components/Editor.jsx
```diff
@@ -48,25 +48,19 @@ import { unlock } from './lock-unlock';
...
 	const titleRef = useRef();
 	const { addEntities, receiveEntityRecords } = useDispatch(coreStore);
-	const { setupEditor } = useDispatch(editorStore);
 
 	useEffect(() => {
 		window.editor = editor;
+		addEntities(postTypeEntities);
 		receiveEntityRecords('postType', post.type, post);
 
-		setupEditor(post, {});
```
dcalhoun, line 69: Will you help me understand why this `setupEditor` call is removed? Does its removal relate to the addition of `addEntities`?

dcalhoun (approved):
Testing succeeded for me. 🚀

ReactApp/src/components/Editor.jsx
geriux: Sure! I've removed it because it is called within the [`EditorProvider`](https://github.com/WordPress/gutenberg/blob/256fa3f93e7ef70625232ade855a0dbb36e03599/packages/editor/src/components/provider/index.js#L267) component.

PR #25, opened 2024-09-17
build: Improve source map accuracy

dcalhoun:
## Description

Leverage the seemingly well-established [`magic-string`](https://github.com/rich-harris/magic-string) package to ensure accurate source maps, improving the debugging experience. 

| Before | After |
| - | - |
| before | after |

## Testing Instructions

Not completely necessary for this developer-centric feature, but the following showcases the difference. 

1. Open the remote editor in Chrome. 
2. Open the developer tools and open the `Editor.jsx` file in the Source tab. 
3. Add a breakpoint. 
4. Refresh the page. 
5. Verify the developer tools displays the accurate inline values (e.g., variable assignment values).


geriux (approved):
Awesome! I was able to test it using the remote editor 🚀

PR #26, opened 2024-09-20
fix: Reinstate editor singleton registry

dcalhoun:
## Description
Reinstate disabling the editor sub-registry so that a singleton registry is
used. This allows third-party blocks to query the block editor state.

Additionally, remove what appears to be a redundant invocation of
`getEditorSettings` that may have been erroneously left in place during
https://github.com/wordpress-mobile/GutenbergKit/pull/24. It appears the `EditorProvider` already [invokes](https://github.com/WordPress/gutenberg/blob/219065c3be5881a68a8667fda9c04c7b7f546763/packages/editor/src/components/provider/index.js#L178) `getEditorSettings`.

## Testing Instructions

Verify the editor — local or remote — launches without errors.

Verify Jetpack AI queries that require block content succeed; e.g., requesting
Jetpack AI modify an existing block.

### DIFF
--- a/ReactApp/src/components/Editor.jsx
+++ b/ReactApp/src/components/Editor.jsx
@@ -145,7 +145,11 @@ function Editor({ post }) {
 	return (
 		hasLoadedPost && (
 			<div className="editor__container">
-				<EditorProvider post={currentPost} settings={settings}>
+				<EditorProvider
+					post={currentPost}
+					settings={settings}
+					useSubRegistry={false}
+				>
 					<BlockCanvas
 						shouldIframe={false}
 						height="auto"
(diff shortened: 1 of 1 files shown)


ReactApp/src/components/Editor.jsx
```diff
@@ -132,13 +128,12 @@ function Editor({ post }) {
 
 	const settings = useMemo(
 		() => ({
-			...editorSettings,
```
dcalhoun, line 135: @geriux as described in the pull request description, I believe this is [redundant](https://github.com/WordPress/gutenberg/blob/219065c3be5881a68a8667fda9c04c7b7f546763/packages/editor/src/components/provider/index.js#L178). Even worse, it creates infinite render loops when `useSubRegistry` is `false`. Will you please help me verify I am not introducing regressions with this change?

geriux: Thank you for addressing this! I've confirmed it does not introduce any regressions 🚀

geriux (approved):

PR #29, opened 2024-10-11
fix: Stabilize Android media uploads

dcalhoun:
## Description

Relates to https://github.com/wordpress-mobile/WordPress-Android/pull/21223. 

Fix various instabilities in Android media uploads. Synchronize Gradle versions
with the WordPress-Android project.

## Testing Instructions

See https://github.com/wordpress-mobile/WordPress-Android/pull/21223.

### DIFF
--- a/Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
+++ b/Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
@@ -22,7 +22,6 @@ import android.webkit.WebViewClient
 import androidx.webkit.WebViewAssetLoader
 import androidx.webkit.WebViewAssetLoader.AssetsPathHandler
 import org.json.JSONObject
-import java.lang.ref.WeakReference
 
 const val ASSET_URL = "https://appassets.androidplatform.net/assets/index.html"
 
@@ -49,16 +48,16 @@ class GutenbergView : WebView {
     var editorDidBecomeAvailable: ((GutenbergView) -> Unit)? = null
     var filePathCallback: ValueCallback<Array<Uri?>?>? = null
     val pickImageRequestCode = 1
-    private var onFileChooserRequested: WeakReference<((Intent, Int) -> Unit)?>? = null
-    private var contentChangeListener: WeakReference<ContentChangeListener>? = null
+    private var onFileChooserRequested: ((Intent, Int) -> Unit)? = null
+    private var contentChangeListener: ContentChangeListener? = null
     private var editorDidBecomeAvailableListener: EditorAvailableListener? = null
 
     fun setContentChangeListener(listener: ContentChangeListener) {
-        contentChangeListener = WeakReference(listener)
+        contentChangeListener = listener
     }
 
     fun setOnFileChooserRequestedListener(listener: (Intent, Int) -> Unit) {
-        onFileChooserRequested = WeakReference(listener)
+        onFileChooserRequested = listener
     }
 
     fun setEditorDidBecomeAvailable(listener: EditorAvailableListener?) {
@@ -145,7 +144,7 @@ class GutenbergView : WebView {
                     intent.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, true)
                 }
 
-                onFileChooserRequested?.get()?.let { callback ->
+                onFileChooserRequested?.let { callback ->
                     handler.post {
                         callback(Intent.createChooser(intent, "Select Files"), pickImageRequestCode)
                     }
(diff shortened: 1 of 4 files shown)


Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -21,22 +21,19 @@ import android.webkit.WebViewClient
...
     private var didFireEditorLoaded = false
-    private var hasSetEditorConfig = false
     private var assetLoader = WebViewAssetLoader.Builder()
         .addPathHandler("/assets/", AssetsPathHandler(this.context))
         .build()
     private var initialTitle: String = ""
     private var type: String = ""
     private var id: Int? = null
-    private var themeStyles: Boolean? = null
+    private var themeStyles: Boolean = false
```
dcalhoun, line 36: Defaulting to `false` felt simpler than a nullable value.

Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -45,19 +42,19 @@ class GutenbergView : WebView {
     private var lastUpdatedContent = ""
 
     private val handler = Handler(Looper.getMainLooper())
-    var editorDidBecomeAvailable: ((GutenbergView) -> Unit)? = null
+    private var editorDidBecomeAvailable: ((GutenbergView) -> Unit)? = null
     var filePathCallback: ValueCallback<Array<Uri?>?>? = null
     val pickImageRequestCode = 1
-    private var onFileChooserRequested: WeakReference<((Intent, Int) -> Unit)?>? = null
```
dcalhoun, line 51: The weak references seemingly led to unexpectedly discarded event handlers, which led to the "Upload" button sporadically failing to operate. Given we null out these values during view destruction, it feels safe to remove the weak reference.

Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -94,15 +91,6 @@ class GutenbergView : WebView {
...
                 request: WebResourceRequest?
             ): WebResourceResponse? {
-                if (!hasSetEditorConfig && withConfig) {
-                    handler.post {
-                        var editorInitialConfig = getEditorConfiguration()
-                        view?.evaluateJavascript(editorInitialConfig, null)
-
-                    }
-                    hasSetEditorConfig = true
-                }
```
dcalhoun, line 104: There is a race condition that results in this approach not consistently load the global values before the HTML page loaded. 

Unfortunately, Android does not provide a method similar to iOS' `addUserScript`. So, this "push" approach was replaced with a "pull" approach on Android to ensure the value is available before the first WebView requests are made.

Demo-Android/gradle/libs.versions.toml
```diff
@@ -1,5 +1,5 @@
 [versions]
-agp = "8.4.2"
+agp = "8.5.1"
```
dcalhoun, line 2: The AGP and Gradle versions were updated to mirror the WordPress-Android project.

Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
nbradbury: Good catch and good call 👍

ReactApp/src/misc/store.js
```diff
@@ -9,12 +9,22 @@ export function getGBKit() {
...
 	}
 
-	const emptyObject = {};
+	// Android relies upon "pulling" the GBKit object from the native host, as it
+	// does not provide a way to inject JavaScript prior to the WebView loading.
+	if (window.editorDelegate) {
+		try {
+			return JSON.parse(window.editorDelegate.getEditorConfiguration());
+		} catch (error) {
+			console.error('Failed parsing GBKit from editorDelegate:', error);
```
nbradbury, line 18: If an error occurs here, is there a way to let the user know something went wrong? And does the message sent to `console.error` get added to the app log? Apologies if those are basic questions, but I have little experience here.

nbradbury (approved):
I left a minor comment that isn't an issue, more of an item for my curiosity. Let's :shipit:!

ReactApp/src/misc/store.js
dcalhoun: Good question. In the event his is triggered—given the fallback value of an empty object—the editor will display normally. However, most actions will undoubtedly fail, creating a confusing experience. 

Unfortunately, it is unlikely a user would be recover from an error of this nature themselves, but we could communicate configuration failed and direct them to contact support/community forums. I captured this opportunity in https://github.com/Automattic/dotcom-forge/issues/9453.

PR #31, opened 2024-10-18
Fix Editor Focus issues

jkmassel:
I originally thought that the editor text selection issues were released to the WebView not being the first responder on iOS, but that's not the case – it seems this issue has been addressed upstream, so pulling in the latest version of the dependencies corrects it.

To test:

1. Run this project from `trunk` on-device and paste a large amount of text. Note that it's difficult to select a small portion of text, and attempting to use "Select All" is difficult to invoke. 
2. Run this PR on-device and try the test again. It should now work how you expect.


ReactApp/package.json
```diff
@@ -13,31 +13,31 @@
...
-		"@wordpress/block-library": "^9.7.0",
-		"@wordpress/core-data": "^7.7.0",
-		"@wordpress/edit-post": "^8.7.0",
-		"@wordpress/editor": "^14.7.0",
-		"@wordpress/element": "^6.7.0",
-		"@wordpress/format-library": "^5.7.0",
-		"@wordpress/preferences": "^4.7.0",
-		"@wordpress/private-apis": "^1.7.0",
-		"react": "^18.3.1"
+		"@monaco-editor/react": "~> 4.6",
```
dcalhoun: What is the rationale for changing to the `~>` syntax? 

I could be mistaken, but I don't believe it is a [valid npm version syntax](https://semver.npmjs.com). I recommend retaining the caret (`^`) syntax for this particular use case; it's the de facto for JavaScript ecosystem and is a relatively stable approach. If we encounter issues with breaking changes, we could remove the range altogether—i.e., `4.6.0` rather than `^4.6.0`. 

WDYT?

dcalhoun (approved):
Good find! The updated dependencies do appear to resolve the focus issue. Thank you!

ReactApp/package.json
jkmassel: Ah yep, I accidentally mixed some Ruby in my JS 😳 

Fixed in 512456d

PR #33, opened 2024-10-21
Add iOS Request Interception

jkmassel:
Related:

- CMM-235-linear-issue
- https://github.com/wordpress-mobile/WordPress-Android/pull/21331

Ref CMM-235.

### DIFF
--- a/Sources/GutenbergKit/Sources/EditorViewController.swift
+++ b/Sources/GutenbergKit/Sources/EditorViewController.swift
@@ -132,6 +132,8 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
     }
 
     private func setUpEditor() {
+        registerRequestInterceptor()
+
         let webViewConfiguration = webView.configuration
         let userContentController = webViewConfiguration.userContentController
         let editorInitialConfig = getEditorConfiguration()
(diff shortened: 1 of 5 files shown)


Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -368,6 +370,64 @@ private final class GutenbergEditorController: NSObject, WKNavigationDelegate, W
...
+    override class func canInit(with request: URLRequest) -> Bool {
+
+        // We don't want to interfere with loading the editor JS
+        guard request.url?.host != "localhost" else {
+            debugPrint("Not intercepting \(request)")
+            return false
+        }
+
+        // We care about WordPress.com resources – let's modify those if needed
+        if request.url?.host?.contains("wordpress.com") == true {
```
dcalhoun, line 383: Eventually, we should hoist this conditional into the host app, as we want to avoid such references in Gutenberg.

Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -368,6 +370,64 @@ private final class GutenbergEditorController: NSObject, WKNavigationDelegate, W
     }
 }
 
-private extension WKWebView {
+class InterceptionProtocol: URLProtocol {
+    override class func canInit(with request: URLRequest) -> Bool {
+
+        // We don't want to interfere with loading the editor JS
+        guard request.url?.host != "localhost" else {
```
dcalhoun, line 377: The development server can also be an IP when testing with a physical device. Maybe we can check against the `GUTENBERG_EDITOR_URL` environment variable as well?

dcalhoun (commented):
You mentioned being unhappy, which aspects of this approach cause that?

dcalhoun:
Superseded by https://github.com/wordpress-mobile/GutenbergKit/pull/144.

PR #30, opened 2024-10-16
Add Android Request Interception

jkmassel:
This initial implementation allows us to inject authentication at runtime if needed.

You can test it by making the demo app activity conform to `GutenbergRequestInterceptor`, then adding:
```

    override fun interceptRequest(request: WebResourceRequest): WebResourceRequest {

        if (request.url.host.isNullOrEmpty()) {
            return request
        }

        if (!request.url.host!!.contains("wordpress.com")) {
            return request
        }

        request.requestHeaders["Authorization"] =
            "Bearer [redacted – replace this string]"

        return request
    }
```

Once you've added a bearer token, copy the URL to an image on a private blog, and insert a new image block with that URL. The editor will now display it.

PR #34, opened 2024-10-23
fix: Rely upon the WebView for completing non-intercepted requests

dcalhoun:
## Related

- https://github.com/Automattic/dotcom-forge/issues/9446
- https://github.com/wordpress-mobile/WordPress-Android/pull/21331
- https://github.com/wordpress-mobile/GutenbergKit/pull/30

## Description

Previously, a manual request was created for _all_ requests. This was
unnecessary and resulted in header values being unexpectedly discarded.

## Testing Instructions

> [!Tip]
> Test using the WPAndroid [prototype build](https://github.com/wordpress-mobile/WordPress-Android/pull/21331#issuecomment-2430588801). 

1. Open the editor for a private WPCOM site.
1. Add an Image block.
1. Upload an image for the block.
1. Verify the image successfully uploads and renders within the editor.
1. Close and re-open the editor.
1. Verify the image successfully renders in the editor.

### DIFF
--- a/Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergRequestInterceptor.kt
+++ b/Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergRequestInterceptor.kt
@@ -5,11 +5,11 @@ import android.webkit.WebResourceResponse
 import android.webkit.WebView
 
 public interface GutenbergRequestInterceptor {
-    fun modifyRequest(view: WebView, request: WebResourceRequest): WebResourceResponse?
+    fun modifyRequest(request: WebResourceRequest): WebResourceResponse?
 }
 
 class DefaultGutenbergRequestInterceptor: GutenbergRequestInterceptor {
-    override fun modifyRequest(view: WebView, request: WebResourceRequest): WebResourceResponse? {
+    override fun modifyRequest(request: WebResourceRequest): WebResourceResponse? {
         return null
     }
 }
--- a/Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergRequestInterceptor.kt
+++ b/Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergRequestInterceptor.kt
@@ -1,13 +1,15 @@
 package org.wordpress.gutenberg
 
 import android.webkit.WebResourceRequest
+import android.webkit.WebResourceResponse
+import android.webkit.WebView
 
 public interface GutenbergRequestInterceptor {
-    fun interceptRequest(request: WebResourceRequest): WebResourceRequest
+    fun modifyRequest(view: WebView, request: WebResourceRequest): WebResourceResponse?
 }
 
 class DefaultGutenbergRequestInterceptor: GutenbergRequestInterceptor {
-    override fun interceptRequest(request: WebResourceRequest): WebResourceRequest {
-        return request
+    override fun modifyRequest(view: WebView, request: WebResourceRequest): WebResourceResponse? {
+        return null
     }
 }
--- a/Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergRequestInterceptor.kt
+++ b/Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergRequestInterceptor.kt
@@ -2,14 +2,18 @@ package org.wordpress.gutenberg
 
 import android.webkit.WebResourceRequest
 import android.webkit.WebResourceResponse
-import android.webkit.WebView
 
 public interface GutenbergRequestInterceptor {
-    fun modifyRequest(request: WebResourceRequest): WebResourceResponse?
+    fun canIntercept(request: WebResourceRequest): Boolean
+    fun handleRequest(request: WebResourceRequest): WebResourceResponse?
 }
 
 class DefaultGutenbergRequestInterceptor: GutenbergRequestInterceptor {
-    override fun modifyRequest(request: WebResourceRequest): WebResourceResponse? {
+    override fun canIntercept(request: WebResourceRequest): Boolean {
+        return false
+    }
+
+    override fun handleRequest(request: WebResourceRequest): WebResourceResponse? {
         return null
     }
 }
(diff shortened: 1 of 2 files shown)


Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergRequestInterceptor.kt
```diff
@@ -1,13 +1,15 @@
 package org.wordpress.gutenberg
 
 import android.webkit.WebResourceRequest
+import android.webkit.WebResourceResponse
+import android.webkit.WebView
 
 public interface GutenbergRequestInterceptor {
-    fun interceptRequest(request: WebResourceRequest): WebResourceRequest
+    fun modifyRequest(view: WebView, request: WebResourceRequest): WebResourceResponse?
```
jkmassel: It's not clear to me where `view` is used – is this just an iOS-style "delegates should receive a reference to the object they're operating on in case there's more than one of them" thing? 

Or did I miss something? 😁

Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -99,38 +91,18 @@ class GutenbergView : WebView {
...
             override fun shouldInterceptRequest(
-                view: WebView?,
-                request: WebResourceRequest?
+                view: WebView,
+                request: WebResourceRequest
             ): WebResourceResponse? {
-                if (request?.url == null) {
-                    return super.shouldInterceptRequest(view, request)
-                } else if(request.url.host?.contains("appassets.androidplatform.net") == true) {
+                val requestResponse = requestInterceptor.modifyRequest(view, request)
```
jkmassel: The old design made it impossible for an mis-configured interceptor to cause the request to be run more than once – this change loses that.

I really love the idea of having the Interceptor perform the HTTP request, that part is great, but I wonder if we can get the simplicity back?

Perhaps the interceptor could have another method like `canIntercept(request) -> Bool`?

WDYT?

jkmassel (changes requested):

Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
dcalhoun: Unfortunately, I do not follow how the proposed changes introduce the possibility of running a request more than once. You may need to elaborate further for me. Are you able to provide an example? 

I'm not opposed to reinstating the protection you describe, but it is difficult for me to imagine how an additional `canIntercept` method might achieve that. Presumably, one could also misconfigure _that_ method and achieve the same duplicative runs. 

I hoisted the performing the HTTP request to the host app because I was unable to mutate what appeared to be read-only request headers in Java (maybe I overlooked something). Additionally, hoisting the networking logic to the host app also provides additional flexibility that the host app needs—e.g., mapping the resource URL to a proxy for private Atomic sites.

Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergRequestInterceptor.kt
dcalhoun: The inclusion of `view` was solely mirroring the function signature of [`shouldInterceptRequest`](https://developer.android.com/reference/android/webkit/WebViewClient#shouldInterceptRequest(android.webkit.WebView,%20android.webkit.WebResourceRequest))—i.e., delegating the entirely of the hook capability to the delegate. 

The `view` parameter is likely unnecessary for this particular use case and could be removed to simplify the function and limit its abilities.

dcalhoun: Removed in 946b8c24e4e55e42304fa0821d27307915ef9cb8.

Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
dcalhoun: Considering it further, my latest interpretation is that you mean the error handling that once resided in GutenbergKit ensured that the WebView completed the request if the host app's request interception failed. I don't consider that preventing duplicate runs, but an attempt to provide a fallback. Let me know if I am misinterpreting the situation. 

That said, my understanding is that the `WebResourceRequest` headers and URL cannot be mutated in the host app. Your original example in the description of https://github.com/wordpress-mobile/GutenbergKit/pull/30 appears to showcase it is possible, but I'm experiencing that it is not in the host app's Java file. Maybe there is a limitation in Java that is not present in Kotlin? 

If retaining the previous structure is important, I suppose the host app could manually build a new `WebResourceRequest` and return that?

jkmassel: > the error handling that once resided in GutenbergKit ensured that the WebView completed the request if the host app's request interception failed

Yeah, if `requestInterceptor.modifyRequest(request)` returns null, the request will be run again in the `WebView` (and almost certainly fail).

My thinking is that we could:

1. Have something that should return pretty much instantly to say "can we handle this specific request" then
2. Have something that tries the actual request (and can only return a response – there'd be no way to return `null`)

dcalhoun: > 1. Have something that should return pretty much instantly to say "can we handle this specific request" then

Rather than a separate hook, I used the return value of `null` to achieve this, similar to how [`shouldInterceptRequest`](https://developer.android.com/reference/android/webkit/WebViewClient#shouldInterceptRequest(android.webkit.WebView,%20android.webkit.WebResourceRequest)) functions. Very much like my [previous mirroring](https://github.com/wordpress-mobile/GutenbergKit/pull/34#discussion_r1813399339) of its method signature. 

If you'd prefer to split the two functions—"can I intercept this request?" and "modify this request"—that is fine, I will refactor the code. 

I will share that I do not necessarily perceive the value in doing so. The original implementation still [returned `null`](https://github.com/wordpress-mobile/GutenbergKit/blob/8ed393e2b0bbc770c430f28f533d3994bc3ee53d/Demo-Android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt#L128-L132) in the event of a failure "handling" the request, which means the WebView would try the request and likely fail.

dcalhoun: Refactored to match the iOS implementation structure in 189ca417f1ca130c579ef70f7006a6a67ae6b842.

jkmassel (approved):
This looks solid, thanks!

PR #48, opened 2024-12-11
feat: Open editor canvas links

dcalhoun:
## What?

Open embedded links outside of the editor WebView—i.e., in the OS web browser
instead.

## Why?

Fixes https://github.com/Automattic/dotcom-forge/issues/9828.

This approach mimics the web editor experience. It also prevents odd UI states
and errors originating from navigating forward and backwards in the single
WebView that displays the editor.

## How?

Leverage the `shouldOverrideUrlLoading` hook to handle URL opening in a new
browser intent.

## Testing Instructions

1. Add a link within Paragraph block.
1. Tap the link preview in the popover to navigate to the URL.
1. Verify the URL opens in the OS browser, not the editor in-app WebView.
1. Navigate back to the editor by pressing the hardware back button.
1. Verify the editor does not crash.

N/A, no user-facing UI changes.

## Screenshots or screencast 
N/A, no user-facing UI changes.

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
@@ -112,6 +112,15 @@ class GutenbergView : WebView {
 
                 return super.shouldInterceptRequest(view, request)
             }
+
+            override fun shouldOverrideUrlLoading(view: WebView?, url: String?): Boolean {
+                url?.let {
+                    val intent = Intent(Intent.ACTION_VIEW, Uri.parse(it))
+                    intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                    view?.context?.startActivity(intent)
+                }
+                return true
+            }
         }
 
         this.webChromeClient = object : WebChromeClient() {
--- a/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
@@ -250,7 +250,7 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
     private func openMediaLibrary(_ config: OpenMediaLibraryAction) {
         delegate?.editor(self, didRequestMediaFromSiteMediaLibrary: config)
     }
-    
+
     public func setMediaUploadAttachment(_ media: String) {
         evaluate("editor.setMediaUploadAttachment(\(media));")
     }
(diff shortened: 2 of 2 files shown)


android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -112,6 +112,15 @@ class GutenbergView : WebView {
...
             }
+
+            override fun shouldOverrideUrlLoading(view: WebView?, url: String?): Boolean {
+                url?.let {
+                    val intent = Intent(Intent.ACTION_VIEW, Uri.parse(it))
+                    intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                    view?.context?.startActivity(intent)
+                }
+                return true
+            }
```
dcalhoun, line 123: I do not perceive any issues with this approach, but please let me know if I am naive.

dcalhoun:
@nbradbury I noted this breaks the Demo app included in this repository. I'll fix that before marking this "ready for review" again.  Sorry for the noise!

dcalhoun:
@nbradbury never mind. It seems it was an unfortunate build cache. This is ready for review.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
nbradbury: This is fine.

nbradbury:
@dcalhoun I'm not sure why, but using a local build of GutenbergKit in WPAndroid isn't working (the app locks up). What do you recommend as another way to test this PR?

dcalhoun:
@nbradbury that is surprising to hear. When using a [local copy](https://github.com/wordpress-mobile/WordPress-Android/blob/f537145e8d18fc908f54966ef308d5c0e168e5d9/local-builds.gradle-example#L17) of GutenbergKit for integration to WPAndroid, you can take one of two approaches: 

1. Run `make build` in GutenbergKit generating a new build output before building WPAndroid. 
2. Run `make dev-server` in GutenbergKit, update the view to [load the server](https://github.com/wordpress-mobile/GutenbergKit/blob/f84ff324a1443e4cbc96f1a290ec216a0416b353/android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt#L178-L190) rather than the build output, and then build WPAndroid. 

Did you use either of these approaches?

nbradbury:
> Did you use either of these approaches?

I did not, but I just tried each command and they both failed with " Cannot read property '@wordpress/api-fetch' of undefined." I've attached the log but it's not very helpful. Any ideas?

[2024-12-12T17_36_23_376Z-debug.log](https://github.com/user-attachments/files/18115578/2024-12-12T17_36_23_376Z-debug.log)

dcalhoun:
@nbradbury ah, OK. An aside: I'm actively updating the README to further clarify how one tests the app in a local development environment in https://github.com/wordpress-mobile/GutenbergKit/pull/49. 

Your shared log appears to show that your development machine is leveraging Node.js v10, which is very old and past EOL. The GutenbergKit project expects v20. Can you verify the Node.js version used for the build process with the following commands? 

```
cd path/to/GutenbergKit/ 
node --version
```

If it is not Node.js v20, I suggest using your apparent `nvm` installation to install the latest Node.js v20.

nbradbury (approved):
@dcalhoun Thanks for the assist - I'm really not skilled at the command line. I was able to use `make build` and use a local GutenbergKit install, and opening links in the browser works as expected. :shipit:

PR #52, opened 2024-12-18
fix: Address disruptive UI bugs

dcalhoun:
## What?

Address styling bugs that disrupt the UX. This reintroduces some hover and focus styles included from the web editor, this was done as the one-off overrides resulted in inconsistent styling and regressions elsewhere[^1]. We should consider modifying and improving button styles in a more robust manner at a later time. 

It's worth noting that this does not fix all style issues, only those I deemed particularly disruptive. 

* Increased editor toolbar button tap target size. 
* Fixed illegible buttons caused from lingering hover styles. 
* Fixed misaligned inserter tab. 
* Fixed inconsistencies in theme styles application—e.g., text size, text color, horizontal root padding. 
* Fixed erroneous default browser styles on `fieldset` elements in block settings—e.g., Image block border radius. 
* Refactored stylesheets to align with Gutenberg project practices—e.g., leveraging Sass, co-locating related styles and markup. 

[^1]: E.g., tapping the buttons in an Image block's placeholder could leave the button styles in odd states, with missing borders or illegible text. 

## Why?

Fixes https://github.com/Automattic/dotcom-forge/issues/8678. Fixes https://github.com/Automattic/dotcom-forge/issues/9943. 

The bugs degrade the UX and erode user trust in the product.

## How?

Refactor to further utilize Gutenberg project components and styling practices,
benefiting from well-tested UI and modularizing custom styles.

## Testing Instructions

Navigate the editor, noting any unexpected styles, layout issues, or bugs.

Navigate the editor, noting any unexpected styles, layout issues, or bugs.

## Screenshots or screencast 
There are numerous visual fixes, too many to capture in screenshots.


src/index.scss
dcalhoun, line 1: The lines removed were either relocated to a specific component folder or removed due to problematic hover style overrides.

kean:
> Increased editor toolbar button tap target size.

The buttons initially appeared too large, but after testing on a physical device, I thought they were just the right size.

> Fixed illegible buttons caused from lingering hover styles.

Confirmed as fixed.

> Fixed misaligned inserter tab.

Also fixed 👏 

There is some misalignment with popovers (no right margin):



> Refactored stylesheets to align with Gutenberg project practices

Some styles appear to be lost. For example, menus now have small buttons again. I completely recognize that initial implementation was not sustainable to say the least 😅 

In terms of the styling in general, I don't think we need to over-complicate it, but it needs some changes as the current styles are not designed for mobile. It can be as simple as:

- Use Apple's color palette. For example, separators and borders should be light gray. The standard Gutenberg black borders are too harsh.
- Add ~8pt corner radius to menus, and anywhere they make sense.
- Use standard body fonts (17pt San Francisco) for all menus, labels, etc.

This will get us 90% to where it feels nice and native.

 For reference, the original design looked more like this:

screenshot-2024-07-01-at-11 13 36e280afam-1

kean (approved):
LGTM, minus the missing styles (see the previous comment)

dcalhoun:
Thank you for your testing, review, and thoughts, @kean. I really appreciate it. 🙇🏻‍♂️ 

> Some styles appear to be lost. For example, menus now have small buttons again. 

Will you share a specific example of a "small button" please? That will help ensure I do not misunderstand your discovery. 

> Use Apple's color palette. For example, separators and borders should be light gray. The standard Gutenberg black borders are too harsh.

It appears I introduced darker borders by wrapping our editor toolbar with `Toolbar`. The dropdown menus in the web editor header toolbar do _not_ have the dark borders, but utilize the lighter gray. I'll investigate a solution. 

That said, at this time, I am not in complete agreement that applying Apple's color palette is the most appropriate solution—at least not everywhere, and not without a robust solution that avoids regressions from CSS overrides. 

1. **We should consider the Android platform.** We could accomplish with platform-specific, top-level classes on the root element for diverging styles, but it's worth debating if that is worth the maintenance burden. 
1. **There is the "uncanny valley" problem.** Attempts to shed Gutenberg's unique web styles and align with Apple's could result in degraded UX where we mimic native UI, but do not deliver the same quality with our web UI. In some ways, a perception or expectation issue. 
1. **Simple CSS overrides are brittle.** Part of the focus of these proposed changes is reverting overrides to ensure the UI is at least consistent within itself, albeit inconsistent with the host platform. Long term, if we truly want to pursue unique platform-aligned styles, we should likely build custom web components for the mobile platform—i.e., avoid the pattern of reusing web UI components and altering them with CSS. 

> Add ~8pt corner radius to menus, and anywhere they make sense.

I'm not inherently opposed to this, but it relates to what I shared above regarding border colors and platform alignment. There are instances where border radii were modified, but introduced regressions—e.g., mismatched border radii between parent and child. 

> Use standard body fonts (17pt San Francisco) for all menus, labels, etc.

Will you share a specific example that you believe causes problems currently please? I do not want to misinterpret your statement.

kean:
> Attempts to shed Gutenberg's unique web styles and align with Apple's could result in degraded UX where we mimic native.

Sorry, I wasn't entirely clear as that was not my suggestion. My point is that Gutenberg isn't optimized for mobile devices, and it's our job to do it. I think it should retain its unique WordPress look and feel. It's really the only option.

The primary point is that text should be legible. Mobile requires much larger fonts than desktop. On macOS, body (most text) uses size 13pt. The equivalent on iOS is 17pt. It's pretty clear that the current fonts are too small if you contrast it with the native components:

Screenshot 2024-12-20 at 9 24 42 AM

For Gutenberg menus, I would probably go a little bit smaller with 16pt as it typically provides a lot of options. But, ideally, these need to be grouped and hidden in submenus on mobile where applicable.

I also advocate for using light gray colors where applicable for borders and adding a corner radius on iOS. It will make it feel more like part of the app. Rounded corners are quintessentially Apple, and it will go a long way to make it feel more like also part of the platform. I hope it won't be complicated from the technical point.

dcalhoun:
Before merging, I pushed a few changes improving legibility and tap targets. We can iterate further in future pull requests. 

* Fixed the dark gray borders found in the editor toolbar dropdown menus. 
* Reduced the editor toolbar button tap targets from 48pt to 46pt. 
* Increased the baseline text for interactive buttons from 13pt to 17pt. 
* Increased the minimum tap target height for menu list and auto-complete items from 36pt to 42pt.

kean:
Awesome! Let's ship it in 25.6 :shipit:

PR #55, opened 2025-01-08
feat: Editor history navigation

dcalhoun:
Related:

* https://github.com/Automattic/dotcom-forge/issues/9968
* https://github.com/wordpress-mobile/WordPress-iOS/pull/23961
* https://github.com/wordpress-mobile/WordPress-Android/pull/21570

## What?

Enable navigating forwards and backwards through the editor history.

## Why?

Undo and redo actions are common, helpful features expected by users.

## How?

Expose the availability of undo/redo actions to the host app, alongside methods
to invoke the relevant actions.

## Testing Instructions

1. Insert and edit various content in the editor.
1. Tap the undo and redo buttons.
1. Verify editor history is accurately replayed.

Perform the testing instructions with a screen reader.

## Screenshots or screencast 

| iOS | Android |
| - | - |
| Simulator Screenshot - iPhone 16 Pro - 2025-01-08 at 12 06 54 | Screenshot_20250109_102658_Jetpack Pre-Alpha |


android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -113,8 +118,9 @@ class GutenbergView : WebView {
                 return super.shouldInterceptRequest(view, request)
             }
 
-            override fun shouldOverrideUrlLoading(view: WebView?, url: String?): Boolean {
-                url?.let {
+            override fun shouldOverrideUrlLoading(view: WebView?, request: WebResourceRequest): Boolean {
+                val url = request.url.toString()
+                url.let {
```
dcalhoun, line 123: Refactored to address deprecated method warning.

nbradbury:
@dcalhoun Testing the Android app, I noticed that undo is one character at a time. This isn't an uncommon approach (in fact, this is how GitHub handles undo), but some people expect undoing a word at a time or even a sentence at a time. Is this planned for the future?

**Update:** Ah, it appears it only does the one character a time undo in the title. In the body, it works as I expected. This is probably fine.

https://github.com/user-attachments/assets/1f380ae7-3bdb-40d1-a132-490d21c61869

nbradbury:
@dcalhoun Image undo is a little wonky. If I add an image then undo it, it returns to showing me the "image box" (not sure what to call it) where I can choose a different image. I guess that makes sense - it's possible the user chose the wrong image - but if I undo that, the editor simply shows the **/** I typed to add a block. Is that intentional? I do see it works this way for other blocks, too, but I have to type **/** again and backspace to be able to choose a block.


https://github.com/user-attachments/assets/caa80322-f96c-4d47-b4c4-3c41ea6c24ed

dcalhoun:
@nbradbury thank you for the thorough testing! 🙇🏻‍♂️ 

> I noticed that undo is one character at a time

I've noted the same behavior. As you mention in your update, it does appear to occur more readily in the post title field—not sure why that is the case. Also, this behavior occurs in the web editor as well, so it is not specific to GutenbergKit's implementation. If a change is desired, we'll need to address it in Gutenberg. That is definitely possible, but I consider it outside of the scope of this work. 

> Image undo is a little wonky. If I add an image then undo it, it returns to showing me the "image box" (not sure what to call it) where I can choose a different image. I guess that makes sense - it's possible the user chose the wrong image - but if I undo that, the editor simply shows the **/** I typed to add a block. Is that intentional?

Similar to the pace of undo for characters vs words, this behavior matches the web editor. Its correctness is debatable, but I suggest we address it in Gutenberg in a separate scope of work.

nbradbury:
> If a change is desired, we'll need to address it in Gutenberg. That is definitely possible, but I consider it outside of the scope of this work ... Its correctness is debatable, but I suggest we address it in Gutenberg in a separate scope of work.

Agreed. I should've taken the time to test on the web before commenting here.

kean (approved):

PR #60, opened 2025-01-14
feat: Toggle code editor

dcalhoun:
Related:

* https://github.com/Automattic/dotcom-forge/issues/10117
* https://github.com/wordpress-mobile/WordPress-Android/pull/21582
* https://github.com/wordpress-mobile/WordPress-iOS/pull/23979

## What?

Enable editing post content mark up with the plain text editor.

## Why?

Mirror functionality in the web editor, and address an error thrown in the iOS
app when pressing the "Code Editor" button.

## How?

Add necessary bridge logic to communicate the toggling of the code editor.
Expand the editor to include conditional logic for both the visual and text
editors.

## Testing Instructions

Verify the code editor is available, allowing creating and saving edits to be
made in both the visual or code editor.

The editor mode can be toggled by tapping the more button (three vertical dots)
in the top-right corner, then tapping Code Editor or HTML mode on iOS and
Android respectively.

Ensure the menu items are accessible via screen reader.

## Screenshots or screencast 
| iOS | Android |
| - | - |
| Screenshot 2025-01-14 at 5 10 44 PM | Screenshot_20250114_171048_Jetpack Pre-Alpha |


nbradbury:
This isn't specific to this PR, but I wanted to point out that dark mode isn't honored in the editor (in either visual or code).

dark

Also not specific to this PR, if I enter visual mode and add something like `<B>BOLD</B>`, when I return to visual mode I get this message that I'm trying to use "the deprecated code editor." I know this isn't something we have control over, but it might be worth passing along that this isn't very user-friendly (I would think most users have no clue what the "deprecated code editor" means).

deprecated

dcalhoun:
> This isn't specific to this PR, but I wanted to point out that dark mode isn't honored in the editor (in either visual or code).

@nbradbury thanks for noting this. The mobile editor has long resided in an odd place when it comes to dark mode support, see p9ugOq-45B-p2 and pc8Hgn-Ch-p2 for the most recent discussion. 

There are few relevant pieces:

1. Editor UI or "chrome"
2. Code editor
3. Visual editor—with and without theme styles applied

The first two, in theory, _could_ universally apply dark mode styles.  The third needs to conditionally apply dark mode styles or respect a theme's styles. All of them venture into the complex topic of embracing or overriding how Gutenberg core approaches—or ignores—dark mode styles. My perception is that there is little-to-no "dark mode" in Gutenberg core at this time. 

> [...] if I enter visual mode and add something like `<B>BOLD</B>`, when I return to visual mode I get this message that I'm trying to use "the deprecated code editor." [...] that this isn't very user-friendly (I would think most users have no clue what the "deprecated code editor" means).

Agreed, this is a confusing experience. At minimum, there is room to improve the copy to provide a clearer, more user-friendly message that is applicable to all contexts—web browser, embedded web view, etc.

I'm honestly uncertain of the intent of the code editor or who/what purpose it serves. As you noted, you cannot simply write (valid) HTML and have the visual editor render it straightforwardly. One can really only write _block mark up_ in the code editor.

nbradbury:
@dcalhoun Functionality-wise this looks good to me, but I'll wait for @kean to do a code review before approving.

kean (approved):
I tested the basic scenarios, and it worked well. Some feedback:

- The title block takes up too much space.
- The font and the line height for code could be a notch smaller. The code is more verbose than the content, and I would optimize it to fit more stuff on the screen.
- Can we get rid of the border? It takes up some valuable horizontal space.

And as I suggested earlier, nice to have: some basic syntax highlighting for HTML tags if it doesn't take much time, and a way to toggle line wrap mode.

dcalhoun:
@kean thank you for the review and feedback. 

> * The title block takes up too much space.
> * The font and the line height for code could be a notch smaller. The code is more verbose than the content, and I would optimize it to fit more stuff on the screen.

I reduced the font size and line height in 500f8b4cee301ef6fef6fc6c24226129d7c3e16c. 

<details><summary>Screenshot: Smaller font size and line height</summary>
<p>

Screenshot 2025-01-16 at 12 40 38


</p>
</details> 

> * Can we get rid of the border? It takes up some valuable horizontal space.

I'd rather not, at least for now. The post title field is self-scrollable, while the post content relies upon body scrolling due to its computed, fixed height. We'd need to build custom components to change this, which we might consider later. 

Because of the aforementioned scrolling behavior, removing the post title field's border then creates an odd experience of overflow-cropped content without a bounding border. Removing merely the post content field's border appears odd, personally. 

Additionally, overriding the component styles with CSS is a brittle approach. I'd prefer to limit the modifications we are making when possible. 

> And as I suggested earlier, nice to have: some basic syntax highlighting for HTML tags if it doesn't take much time, and a way to toggle line wrap mode.

Agreed. I consider them nice-to-haves that we could revisit in the future, ideally in core Gutenberg.

PR #61, opened 2025-01-18
feat: Dismiss keyboard button

dcalhoun:
Related:

- https://github.com/Automattic/dotcom-forge/issues/8673
- https://github.com/wordpress-mobile/WordPress-Android/pull/21602
- https://github.com/wordpress-mobile/WordPress-iOS/pull/23988

## What?

Add a button dedicated to dismissing the virtual keyboard.

## Why?

An explicit button for dismissing the keyboard allows users to view the
entirety of the post content without the virtual keyboard blocking a
portion of the content.

## How?

Add a block toolbar button that is conditionally displayed depending on whether
the virtual keyboard visibility. As there is little-to-no `VirtualKeyboard` API
support in browsers, we rely upon focus and blur events to provide a best guess.

## Testing Instructions

1. Insert or select a text block.
1. Tap the dismiss keyboard button.
1. Verify the virtual keyboard closes and the block is deselected.
1. Insert or select a non-text block.
1. Verify the dismiss keyboard button is not displayed.

Verify the dismiss keyboard button has a helpful label.

## Screenshots or screencast 
| iOS | Android |
| - | - |
| ios-dismiss-keyboard | android-dismiss-keyboard |


nbradbury:
@dcalhoun This seems to be working great, with one exception: if I dismiss the keyboard using the Back arrow in the bottom nav bar, the new keyboard button continues to appear until I tap on it:

https://github.com/user-attachments/assets/c1c5690c-1a36-42ea-a143-50ca83a8c3b0

Also, given that the bottom Back arrow dismisses the keyboard, is this new button necessary? Perhaps we have this new button in case the user doesn't realize the back arrow closes the keyboard?

dcalhoun:
> if I dismiss the keyboard using the Back arrow in the bottom nav bar, the new keyboard button continues to appear until I tap on it:

> Also, given that the bottom Back arrow dismisses the keyboard, is this new button necessary? Perhaps we have this new button in case the user doesn't realize the back arrow closes the keyboard?

@nbradbury ah, good catch on the hardware back button oddity. I implemented the dismiss keyboard button on Android because it was present in Gutenberg Mobile. However, you make a good point that it is somewhat superfluous given how the hardware back button dismisses the virtual keyboard. I suppose the button is far more useful on iOS, which lacks a hardware back button. 

Given Android's back button and the current shortcoming of the dedicated dismiss keyboard button lingering, it might make more sense to forgo the dedicated button on Android. WDYT? 

I will ask one question: Is the hardware back button's presence and behavior universal? I.e., will it always be available for users on various devices?

nbradbury:
> Is the hardware back button's presence and behavior universal? I.e., will it always be available for users on various devices?

Given the number of Android manufacturers and devices, I'd be hard-pressed to answer that with 100% certainty. But I do note that on both my Google and Samsung devices, the back button is present and dismisses the keyboard.

> it might make more sense to forgo the dedicated button on Android. WDYT?

I agree.

dcalhoun:
> Given the number of Android manufacturers and devices, I'd be hard-pressed to answer that with 100% certainty. But I do note that on both my Google and Samsung devices, the back button is present and dismisses the keyboard.

Fair enough. 😅 

From a bit of research, it appears Android has a [Compatibility Definition Document](https://source.android.com/docs/compatibility/15/android-15-cdd#723_navigation_keys) where it outlines requirements for Android device compatibility, including navigation functions. It seems back function buttons are recommended, but not required; it is recognized there may be valid reasons for excluding back function buttons, albeit for particular circumstances. 

> The [Home](http://developer.android.com/reference/android/view/KeyEvent.html#%60KEYCODE_HOME%60), [Recents](http://developer.android.com/reference/android/view/KeyEvent.html#%60KEYCODE_APP_SWITCH%60), and [Back](http://developer.android.com/reference/android/view/KeyEvent.html#%60KEYCODE_BACK%60) functions typically provided via an interaction with a dedicated physical button or a distinct portion of the touch screen, are essential to the Android navigation paradigm and therefore, device implementations:

Given the aforementioned Android suggestion and the shortcoming of the proposed dismiss keyboard button implementation—which would likely require explicit bridge communication to address—my thought is we should forgo the dedicated dismiss keyboard button. I implemented an Android exclusion.

kean:
I tested in on the device, and it worked pretty well.

I'm also can't say I'm completely sure about its value on iOS, as it's not without the downsides: it takes away some valuable horizontal space and makes it a bit harder to reach the "More" button. Previously, you could just scroll to the right and knew it was always the last item on the edge of the screen. It now requires a bit more precision to tap. 

It's uncommon for iOS apps to have a "Dismiss Keyboard" button. The common approach is to use the interactive [keyboardDismissMode](https://developer.apple.com/documentation/uikit/uiscrollview/keyboarddismissmode-swift.property). I haven't tested if it works with `WKWebView`, but I assume it does. The other common option is to dismiss the keyboard when you tap in an empty area where there are no controls.

One similar example is the Notes app. It uses the interactive dismiss mode. I'm not also not entirely sure how important it is to be able to dismiss a keyboard in the first place. You are pretty much always editing.

dcalhoun:
> I'm also can't say I'm completely sure about its value on iOS, as it's not without the downsides: it takes away some valuable horizontal space and makes it a bit harder to reach the "More" button. 

@kean I perceive the usage of horizontal space as very small downside, one of many trade-offs involved in designing UI for a complex and robust text editor. The usage of the horizontal space for the button was done in the name of improving the overall editing experience. 

> It's uncommon for iOS apps to have a "Dismiss Keyboard" button. The common approach is to use the interactive [keyboardDismissMode](https://developer.apple.com/documentation/uikit/uiscrollview/keyboarddismissmode-swift.property). I haven't tested if it works with `WKWebView`, but I assume it does. The other common option is to dismiss the keyboard when you tap in an empty area where there are no controls.

Unfortunately, I do not believe `keyboardDismissMode` works with `WKWebView`, but I'd love to be proven wrong. An alternative is relying upon JavaScript scroll events to dismiss the keyboard. This is definitely possible, but likely would take decent effort to achieve an experience that feels natural, and not "uncanny valley"-like. I'm open to this, but figured a dedicated button was a more reasonable solution, one that aligns with the current editing experience. 

Yes, tapping non-interactive/focusable areas dismisses the keyboard, but those types of areas are limited and not obvious in the block editor. 

I believe these nuanced aspects are why mobile Safari displays a "Done" button above the keyboard when focusing interactive elements. 

> I'm not also not entirely sure how important it is to be able to dismiss a keyboard in the first place. You are pretty much always editing.

Personally, I place a lot of value on dismissing the keyboard in the editor. I frequently read and re-read my written word, many times over. Scrolling and reading with the keyboard open is cumbersome IMO.

kean (approved):
> Unfortunately, I do not believe keyboardDismissMode works with WKWebView, but I'd love to be proven wrong.

It's an opt-in feature. You can enable it like so:

```swift
self.webView.scrollView.keyboardDismissMode = .interactive
```

I got it to work, but because of the way the toolbar is implemented, it does look odd. It's not nearly as bad when you do it quickly on a device.

https://github.com/user-attachments/assets/8dd5d1f7-8996-4de8-9083-733c806a4a54

> Personally, I place a lot of value on dismissing the keyboard in the editor.

I'll leave it up to; just wanted to share my view. I'd enable interactive dismiss because, as a user, you expect this gesture to work everywhere across the OS. If it doesn't look ideal, I would probably overlook it. 

I don't see a lot of harm taking away the horizontal space – just wanted to point it out.

dcalhoun:
@kean I appreciate the feedback. 

My opinion is that dismissing the keyboard through interactive swiping is a better experience than a dedicated button. I opened https://github.com/wordpress-mobile/GutenbergKit/pull/63 to achieve that. Thank you for the help.

PR #91, opened 2025-02-12
feat: Default block appender

### LINKED ISSUE #77: Allowing inserting a Paragraph block by tapping after the last block in the block canvas

The web editor provides a few easy ways to insert the default block (Paragraph) after existing content—e.g., the default block appender at then of an empty Paragraph, hovering between blocks. Gutenberg Mobile provided the ability to tap the empty space beneath the last block in the block list to append the default block. 

GutenbergKit does not provide such a mechanism, requiring the user to open the inserter to select a block. 

We mimic Gutenberg Mobile's approach, allowing a user to insert a default block by tapping the space beneath the last block.

dcalhoun:
## What?

Allow tapping beneath the last block to quickly inserter the default
block—a Paragraph block.

## Why?

Fix #77. Mimics similar abilities found in the web editor and Gutenberg Mobile.
Improve the writing flow, allowing users to append additional blocks without
opening the block inserter.

## How?

Add a hidden, dedicated button for appending a default block.

## Testing Instructions

1. Open the editor.
1. Insert a block.
1. Tap the space beneath the last block.
1. Verify an empty Paragraph block is appended at the end of the content.

Perform the above testing steps with a screen reader.

## Screenshots or screencast 
N/A, no visible changes.


package.json
dcalhoun, line 1: All of the dependency additions merely capture existing, undeclared dependencies. Their absence does not currently cause issue because they are transitive dependencies for other declared dependencies, but relying on that is brittle. These changes ensure all project dependencies are declared.

dcalhoun:
@kean sending a friendly message to ensure this review request is not overlooked. Thank you!

kean:
Hey, I tested in a simulator (may not be ideal representation) and it is definitely an improvement and make the editor feel more responsive, but I noticed a couple of things:

- If the last block is a paragraph, it still inserts a new paragraph block, even if the last paragraph is empty
- It doesn't always work depending on where you tap

https://github.com/user-attachments/assets/780a869a-147b-4698-bff9-980bde0bc095

I often use Notes as a reference because I believe it has nearly flawless text editing. In Notes, when the keyboard is dismissed and you tap below the content, if the last section is a paragraph, the cursor automatically moves to the end of that paragraph. This allows you to either continue adding text or press "Return" to create a new paragraph. If the last "block" isn't a paragraph, only then it seems to start a new paragraph. I think it's probably ideal.

kean:
(Potentially related) I'd drop the "Type / to choose a block", especially if you have no hardware keyboard connected – simply showing a cursor would be much nicer. We could leverage the newly added [TipKit](https://github.com/wordpress-mobile/WordPress-iOS/pull/23648) or something similar web-based to teach the user about "/" once and when they connect the keyboard.

dcalhoun:
> I noticed a couple of things

Your proposed improvements make sense. The PR matches the current Gutenberg Mobile implementation. WDYT about capturing the improvements in a separate issue and merging this as-is? 

> (Potentially related) I'd drop the "Type / to choose a block", especially if you have no hardware keyboard connected – simply showing a cursor would be much nicer.

I'm open to discussing this further, but, as you referenced, consider it tangentially related and would prefer addressing it separately.

kean:
Sounds good. I opened two issues: https://github.com/wordpress-mobile/GutenbergKit/issues/98, https://github.com/wordpress-mobile/GutenbergKit/issues/99.

kean (approved):

Issue #79, opened 2025-02-07
Draft posts containing media are always marked as unsaved on iOS


dcalhoun:
## Description

On **iOS only**, if a **draft** post contains a **media** block, the post remains marked as "unsaved" even after saving the draft. 

## Step-by-step reproduction instructions

1. Create a new post in the editor. 
2. Add an Image block. 
3. Tap the more menu (three-horizontal dots) in the top-right corner. 
4. Tap _Save Draft_. 
5. Exit the editor after the draft saves. 

## Expected behaviour

The editor closes without prompting to save edits. 

## Actual behaviour

The editor prompts to save edits. 

## WordPress information

-   WordPress version: 
-   Gutenberg version: 
-   Are all plugins except Gutenberg deactivated? 
-   Are you using a default theme (e.g. Twenty Twenty-One)? 

## Device information

-   Device:  iPhone SE
-   Operating system:  iOS 18.2
-   WordPress app version:  25.7

dcalhoun:
It seems the REST API modifies `img` HTML adding a space character before the end of the tag. This results in the host WordPress-iOS app comparing the latest REST API content (w/ space) to the latest local Gutenberg content (w/o space) and determining the content changed. 

The added space occurs on the web's `wp-admin/post.php` as well. One can observe it there by: 

1. Create a new post in WP Admin.  
2. Add an Image with media. 
3. Save the post. 
4. Note the HTML in the "Code editor." 
5. Refresh the page. 
6. Review the updated HTML in the "Code editor." 

I'm unsure as to why the space is added, but it's problematic for our current approach to monitoring changes in WordPress-iOS. 

```diff
diff --git a/test.html b/test.html
index ada518fbf5..5246bea63e 100644
--- a/test.html
+++ b/test.html
@@ -1,3 +1,3 @@
 <!-- wp:image {"id":327,"sizeSlug":"large","linkDestination":"none"} -->
-<figure class="wp-block-image size-large"><img src="https://[REDACTED]/wp-content/uploads/2025/02/img_0342.jpeg?w=1024" alt="" class="wp-image-327"/></figure>
+<figure class="wp-block-image size-large"><img src="https://[REDACTED]/wp-content/uploads/2025/02/img_0342.jpeg?w=1024" alt="" class="wp-image-327" /></figure>
 <!-- /wp:image -->
\ No newline at end of file

```

kean:
[...]

PR #104, opened 2025-02-20
fix: Improve autosave and unsaved changes detection

dcalhoun:
Related:
- https://github.com/wordpress-mobile/GutenbergKit/issues/79
- https://github.com/wordpress-mobile/WordPress-iOS/pull/24102
- https://github.com/wordpress-mobile/WordPress-Android/pull/21698

## What?

Improve autosave and change detection logic, aligning more closely with the
approach of Gutenberg Mobile.

## Why?

Previously, host apps determined whether content changes were present by
comparing the content received from GutenbergKit to the latest server-provided
post content. This is problematic as the server makes slight modifications to
content markup, creating perpetually changed content.

## How?

To address this, we need GutenbergKit to communicate changed content. We
accomplish by sending a new `changed` parameter alongside the title and content.
Rather than updating the post and content refs on every content change, we now
do so at the time at which the host app requests the latest content. This
ensures the host and GutenbergKit synchronize their latest perception of
changes.

We also now rely upon `AutosaveMonitor` to throttle how often we request the
host app persist the latest changes, rather than a generic store subscription.

## Testing Instructions

See sibling PRs for platform-specific testing instructions:

- https://github.com/wordpress-mobile/WordPress-iOS/pull/24102
- https://github.com/wordpress-mobile/WordPress-Android/pull/21698

N/A, no user-facing changes.

## Screenshots or screencast 
N/A, no user-facing changes.


src/components/editor/use-editor-setup.js
```diff
@@ -24,20 +21,11 @@ export function useEditorSetup( post ) {
 		receiveEntityRecords( 'postType', post.type, post );
 
 		setupEditor( post, {} );
-		registerCoreBlocks();
```
dcalhoun, line 27: This was relocated to the entry file as the blocks must be registered before we parse the raw post content provided by the host app. 

https://github.com/wordpress-mobile/GutenbergKit/blob/8ccd1fb448d6691fa1f1ba940657dd241c04790c/src/components/editor/use-host-bridge.js#L28

src/components/editor/use-host-bridge.js
```diff
@@ -26,6 +22,12 @@ export function useHostBridge( post, editorRef ) {
 		[ editEntityRecord, post.id, post.type ]
 	);
 
+	const postTitleRef = useRef( post.title.raw );
+	const postContentRef = useRef( null );
+	if ( postContentRef.current === null ) {
+		postContentRef.current = serialize( parse( post.content.raw || '' ) );
+	}
```
dcalhoun, line 29: Explicitly [selectively instantiated](https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents) to avoid unnecessary performance costs.

src/components/editor/use-host-bridge.js
```diff
@@ -48,10 +50,18 @@ export function useHostBridge( post, editorRef ) {
...
+			const title = getEditedPostAttribute( 'title' );
+			const content = getEditedPostContent();
+			const changed =
+				title !== postTitleRef.current ||
+				content !== postContentRef.current;
+
+			if ( changed ) {
+				postTitleRef.current = title;
+				postContentRef.current = content;
+			}
```
dcalhoun, line 62: Selectively updating these refs allows us to accurately communicate a "content has changed" status to the host app, which is important for WordPress-iOS' implementation.

src/components/editor/use-host-bridge.js
```diff
@@ -87,23 +97,6 @@ export function useHostBridge( post, editorRef ) {
...
-			if (
-				title !== postTitleRef.current ||
-				content !== postContentRef.current
-			) {
-				onEditorContentChanged();
-				postTitleRef.current = title;
-				postContentRef.current = content;
-			}
-		} );
-	}, [] );
```
dcalhoun, line 106: This generic subscription is replaced by the `AutosaveMonitor`, which provides a more robust implementation, including intervals and dirty checks.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -321,15 +319,25 @@ class GutenbergView : WebView {
...
                     lastUpdatedTitle = jsonObject.getString("title")
                     lastUpdatedContent = jsonObject.getString("content")
+                    changed = jsonObject.getBoolean("changed")
                 } catch (e: JSONException) {
                     Log.e("GutenbergView", "Received invalid JSON from editor.getTitleAndContent")
                 }
 
-                callback.onResult(lastUpdatedTitle ?: "", lastUpdatedContent ?: originalContent)
+                val title = lastUpdatedTitle ?: ""
+                val content = if (changed) {
```
kean, line 335: TIL `if/else` is an expression in JS 👀

kean (approved):
Approved as it's an improvement. I would, however, considering implementing it in a slightly different way: 

- The app initialize the editor with the initial title and content
- The editor keeps track of whether the post has changes (title/content) and communicates it via a new property `EditorState/hasChanges`.
- If the app receives an update with `hasChanges` set to `true`, it only then triggers the autosave logic to put data on disk. Autosave happens infrequently and only as a crash-protection measure. This is the only time the app needs to get the latest content from the editor.
- The app never changes the title or content on its side. If it needs to, it uses one of the editor bridge APIs to do so (e.g. when restoring from a previous revision)

This setup will minimize the need to pass serialized content via the bridge and establish the source of truth for the content – the editor. I might be missing something, but it seems sound.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
dcalhoun: This is Kotlin. 😃

dcalhoun:
> I would, however, considering implementing it in a slightly different way:
> * The app initialize the editor with the initial title and content
> [...]
> * The app never changes the title or content on its side. If it needs to, it uses one of the editor bridge APIs to do so (e.g. when restoring from a previous revision)

My perception is that our approach accomplishes these two aspect currently. Please correct me if I'm wrong. 

> * The editor keeps track of whether the post has changes (title/content) and communicates it via a new property `EditorState/hasChanges`.
> * If the app receives an update with `hasChanges` set to `true`, it only then triggers the autosave logic to put data on disk. Autosave happens infrequently and only as a crash-protection measure. This is the only time the app needs to get the latest content from the editor.

Given the `AutosaveMonitor` already [limits its calls](https://github.com/WordPress/gutenberg/blob/db7153125f6dcbd7135ac12af3adc6e56f96109f/packages/editor/src/components/autosave-monitor/index.js#L39-L42) of `onEditorContentChange` to when changes exist, in a way, we already achieve this; we just do not rely upon a dedicated state variable in the host app. I.e., `onEditorContentChange` → host autosave is only ever triggered when content was changed, which then leads to the host requesting serialized content from the library. 

As `onEditorContentChange` and `getTitleAndContent` are two separate loops running at different times, the new `changed` parameter ensures library communicates the latest state at the time the host is seeking to retrieve/persist the latest content. 

> This setup will minimize the need to pass serialized content via the bridge and establish the source of truth for the content – the editor. I might be missing something, but it seems sound.

I agree avoiding unnecessary bridge communication is a worthwhile target. However, it's unclear to me whether the proposed changes would reduce communication. That said, I am very open to refactors to reduce complexity. I'd just like to understand the proposal better—if you are willing to open pull requests expressing the idea, even better.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
kean: Ahhh right.

kean:
> I agree avoiding unnecessary bridge communication is a worthwhile target. However, it's unclear to me whether the proposed changes would reduce communication.

One of the reasons would be to ensure that the "Publish/Save" button updates (nearly) instantly after you make a change, otherwise it feels a bit slow. It's, of course, not the end of the world.

In terms of the performance, my assumption is that serializing and sending the entire post content on every keystone is not feasible, especially with larger posts. Perhaps, there is an acceptable setting that balances performance and responsiveness. Unfortunately, I'm not sure how to measure it, but I'm suggesting a potential design where no serialization is needed*, so it never becomes a bottleneck. The logic that responds to changes and determines if the content is different from the initial content will live entirely in the JS land. 

*it will only be needed once every ~10 seconds for autosave

dcalhoun:
> One of the reasons would be to ensure that the "Publish/Save" button updates (nearly) instantly after you make a change, otherwise it feels a bit slow.

I agree, this would be nice. I'll note that Gutenberg Mobile (GBM) does not perform this way at the moment. 

> [...] I'm suggesting a potential design where no serialization is needed*, so it never becomes a bottleneck.

I understood the _intent_; I fail to understand _how_ your proposal's execution differs from what I [shared](https://github.com/wordpress-mobile/GutenbergKit/pull/104#issuecomment-2683007110), specifically regarding performance. 

My perception now is that the critical piece you suggest is immediately updating UI based on the `onEditorContentChange` event and only applying an interval to the persistence act. Currently, both UI updates (Update/Publish button state) and persistence (autosave) are performed on an interval managed in the host app—current is [seven seconds](https://github.com/wordpress-mobile/WordPress-iOS/blob/fb4d598eccbfc9f5576cd7d5296dffcf668b835c/WordPress/Classes/ViewRelated/NewGutenberg/NewGutenbergViewController.swift#L327), the proposal is [one second](https://github.com/wordpress-mobile/WordPress-iOS/blob/fb4d598eccbfc9f5576cd7d5296dffcf668b835c/WordPress/Classes/ViewRelated/Post/Utils/Autosaver.swift#L25) (mirroring GBM). 

I believe accomplishing immediate updates for the UI involves a larger refactor of code untouched in the proposed changes. Namely, the host manages the [enablement](https://github.com/wordpress-mobile/WordPress-iOS/blob/fb4d598eccbfc9f5576cd7d5296dffcf668b835c/WordPress/Classes/ViewRelated/Post/PostEditorState.swift#L198) of the Publish/Update button via [changes](https://github.com/wordpress-mobile/WordPress-iOS/blob/fb4d598eccbfc9f5576cd7d5296dffcf668b835c/WordPress/Classes/ViewRelated/NewGutenberg/NewGutenbergViewController.swift#L54) to the [post `changes` attribute](https://github.com/wordpress-mobile/WordPress-iOS/blob/fb4d598eccbfc9f5576cd7d5296dffcf668b835c/WordPress/Classes/ViewRelated/Post/PostEditor.swift#L82). Updating the post `changes` attribute requires the latest content from the editor. We'd need to replace the UI's reliance on the `post` object altogether.

kean:
> We'd need to replace the UI's reliance on the post object altogether.

Yeah, the way it uses `changes` is neat. The "Publish" button does need to track not only the content but also the post settings changes.

> the proposal is [one second](https://github.com/wordpress-mobile/WordPress-iOS/blob/fb4d598eccbfc9f5576cd7d5296dffcf668b835c/WordPress/Classes/ViewRelated/Post/Utils/Autosaver.swift#L25) (mirroring GBM).

Yeah, maybe it's OK. In terms of UX, I'd say it feels fine. Performance – not sure how to measure. I think you convinced me that your version is ultimately better.

PR #107, opened 2025-03-05
Add iOS support for custom bundles

jkmassel:
## Testing Instructions

Verify that the solution works with sites that are
- [ ] located under a subdirectory: `https://example.com/blog`
- [ ] using a non-English locale for the whole site and/or the logged-in user.


ios/Demo-iOS/Sources/EditorDownloadView.swift
```diff
@@ -0,0 +1,67 @@
...
+
+        func download() {
+            Task {
+
+                let url = URL(string: siteUrl)!
+                    .appendingPathComponent("wp-json")
+                    .appendingPathComponent("__experimental")
+                    .appendingPathComponent("wp-block-editor")
+                    .appendingPathComponent("v1")
+                    .appendingPathComponent( "editor-assets" )
```
dcalhoun, line 30: To simplify using different endpoints—e.g., it varies based on environment—we might use this path as a default if a full URL is not provided.

ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
```diff
@@ -20,4 +20,12 @@ public struct EditorConfiguration {
         self.title = title
         self.content = content
     }
+
+    var manifestURL: URL {
+        URL(string: siteApiRoot)!
+            .appendingPathComponent("__experimental")
+            .appendingPathComponent("wp-block-editor")
+            .appendingPathComponent("v1")
+            .appendingPathComponent( "editor-assets" )
```
dcalhoun, line 35: To simplify using different endpoints—e.g., it varies based on environment—we might use this path as a default if a full URL is not provided.

ios/Sources/GutenbergKit/Sources/EditorLibrary.swift
```diff
@@ -0,0 +1,223 @@
...
+        var html =
+"""
+<!doctype html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
+        <title>Gutenberg</title>
+        [!--- Scripts ---]
+        [!--- Styles  ---]
```
dcalhoun, line 158: After inserting the fetched editor assets, we need to insert the editor UI scripts and styles as well. The "remote" editor has a separate bundle from the bundled editor. Without this, the editor is never instantiated and displayed. 

How might we ensure the [asset tags](https://github.com/wordpress-mobile/GutenbergKit/blob/947c5315198eea96e3e8c32fcc6afa59ef242638/ios/Sources/GutenbergKit/Gutenberg/remote.html#L10-L11) and corresponding asset files are available and loaded?

dcalhoun: An aside: Once this new bundle feature is in place, we might one day consider aligning the asset loading strategy for both the bundled and remote editors to operate in the same way, where WordPress packages are [externalized](https://github.com/wordpress-mobile/GutenbergKit/blob/947c5315198eea96e3e8c32fcc6afa59ef242638/vite.config.remote.js#L36-L119) for both. 

The primary differences is that the former bundles both the WordPress packages and editor UI into a single JavaScript file, the latter relies upon WordPress packages available in `window.wp` globals and its bundle only includes editor UI. See pbArwn-6yv-p2. 

Aligning them might reduce the likelihood of bugs originating from build complexity or environment differences. I don't think aligning them is necessary for merging this functionality.

dcalhoun: It is likely unnecessary, but we might consider using valid HTML comments to avoid a situation where these strings are accidentally displayed in the WebView. I.e., `<!--- Scripts --->` rather than `[!--- Scripts ---]`.

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -17,13 +18,15 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
     public weak var delegate: EditorViewControllerDelegate?
 
     /// A custom URL for the editor.
```
dcalhoun, line 21:
```suggestion
```

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -106,15 +109,16 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
     }
 
     private func loadEditor() {
-        if let editorURL = editorURL ?? ProcessInfo.processInfo.environment["GUTENBERG_EDITOR_URL"].flatMap(URL.init) {
+        // If you want to load something that's not a manifest, specify the environment variable
+        if let editorURL = ProcessInfo.processInfo.environment["GUTENBERG_EDITOR_URL"].flatMap(URL.init) {
```
dcalhoun, line 123: Unlikely a priority to start, but I imagine we'll need a way to load a fetched asset set/manifest _and_ a local running server for the editor UI bundle. That would allow iterating on the editor UI while relying upon site-specific WordPress packages.

dcalhoun (commented):
Very cool! Thank you for exploring this. It is incredibly helpful. 

The bundled editor seems to function great. Expectedly, the remote/site-specific editor does not load, currently.

I left a few comments, I'd love your thoughts on them.

ios/Demo-iOS/Sources/EditorDownloadView.swift
jkmassel: This can/should be improved a bunch. Because it's the demo app, I just whipped up something quick.

In the real world, we'd probably use https://github.com/Automattic/wordpress-rs to find this endpoint in a programmatic way by scanning the `wp-json` endpoint for its entry. If it's not there, we'd assume the site doesn't support editor assets and default back to the bundled editor.

I think ... LMK if this doesn't seem like a good plan?

ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
jkmassel: It would be far better to inject this, I think.

We already have it in the demo app, so I think it'd just be a matter of passing it in here?

ios/Sources/GutenbergKit/Sources/EditorLibrary.swift
jkmassel: >It is likely unnecessary, but we might consider using valid HTML comments to avoid a situation where these strings are accidentally displayed in the WebView. I.e.,  rather than [!--- Scripts ---].

Yeah this is probably best

jkmassel: > How might we ensure the [asset tags](https://github.com/wordpress-mobile/GutenbergKit/blob/947c5315198eea96e3e8c32fcc6afa59ef242638/ios/Sources/GutenbergKit/Gutenberg/remote.html#L10-L11) and corresponding asset files are available and loaded?

I think pretty much all of these use the same format (we add `type="module" crossorigin` to scripts and `rel="stylesheet" crossorigin` to styles), so we could probably build this up from the raw asset URLs ourselves?

> where WordPress packages are [externalized](https://github.com/wordpress-mobile/GutenbergKit/blob/947c5315198eea96e3e8c32fcc6afa59ef242638/vite.config.remote.js#L36-L119) for both

Thanks for the explanation – I don't think I was familiar with it before. I think treating everything as though it's a remote editor will definitely avoid nasty surprises later. It should still be pretty predictable for the bundled editor (all of the "remote" things are still on-disk), so there doesn't seem to be much downside AFAICT.

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
jkmassel: Ah yeah I didn't consider this case. If you have a fetched asset set, is there any local editor code running? I think I assumed that the remote server's code would completely replace anything running on a dev server

ios/Sources/GutenbergKit/Sources/EditorLibrary.swift
```diff
@@ -0,0 +1,222 @@
...
+
+        let assets = try await withThrowingTaskGroup(of: CachedAsset.self) { group in
+            let progress = Progress(totalUnitCount: Int64(manifest.scripts.count + manifest.styles.count))
+            callback?(progress)
+
+            var assets: [CachedAsset] = []
+
+            for script in manifest.scripts {
+                group.addTask {
+                    let destination = manifest.rootDirectory.appendingPathComponent(script.path)
```
crazytonyli, line 123: Verify if this approach works with sites that are located under a subdirectory: `http://example.com/blog`

dcalhoun:
Superseded by #148.

PR #50, opened 2024-12-17
fix: Replace unsupported editor architecture

dcalhoun:
## What?

Replace our current editor architecture with the approach recommended in the Gutenberg Framework [documentation](https://wordpress.org/gutenberg-framework/docs/intro).

## Why?

Fixes https://github.com/Automattic/dotcom-forge/issues/10026. 

After merging https://github.com/wordpress-mobile/GutenbergKit/pull/24, several regressions were introduced. Namely, the block toolbar would unexpectedly empty when moving the mouse cursor after typing into a rich text field.

Additionally, the Patterns and Media tabs in the block inserter have not worked since the previous change, as it no longer had access to private APIs enabling their operation.

## How?

Replace our usage of `EditorProvider` with `BlockEditor` provider, as well as reintroduce the necessary supporting code for managing the post content state.

## Testing Instructions

**1 - Formatting options remain present within the block toolbar**
1. Start the development server. 
1. Open the editor in your desktop browser.
1. Insert a Paragraph block.
1. Type a couple of words into the block. 
1. Move your mouse cursor.
1. Verify the block toolbar remains populated with rich text formatting
   options—e.g., bold, italic, etc.

**2 - The Patterns/Media tab populate with content from the current site**
1. Start the development server. 
1. Open the editor within the WordPress mobile app.
1. Open the block inserter.
1. Tap the Patterns/Media tab.
1. Verify the tab is populated with content from the current site.

### Accessibility Testing Instructions

N/A, no user-facing changes.

## Screenshots or screencast 
N/A, no user-facing changes.

PR #108, opened 2025-03-05
refactor: Avoid private useBlockEditorSettings API

dcalhoun:
## What?

Remove use of `@wordpress/block-editor`'s private, non-exported `useBlockEditorSettings` function. This reintroduces changes originally implemented in https://github.com/wordpress-mobile/GutenbergKit/pull/24, but extended to address a disappearing block toolbar and empty Patterns/Media inserter tabs—effectively reverting https://github.com/wordpress-mobile/GutenbergKit/pull/50. 

## Why?

First, using private APIs is inadvisable and not a sound, long-term solution. 

Second, we were actually patching `@wordpress/block-editor` to export `useBlockEditorSettings` as a private API. While patching "works" for the editor bundled with the app, we cannot patch the "remote"/site-specific editor—i.e., one that relies upon WordPress packages sourced from a user's site (see pbArwn-763-p2). 

## How?

First, remove the `@wordpress/block-editor` patch. 

Second, we now rely upon `@wordpress/editor`'s `EditorProvider` rather than `@wordpress/block-editor`'s `BlockEditorProvider`. The positive is that this approach negates our need to access `useBlockEditorSettings` as its usage is now managed by the `EditorProvider`. The negative is that the `EditorProvider` strips other experimental features—e.g., populating content for the Patterns/Media inserter tabs—by default, which leads to the next change... 

Third, we introduce a `@wordpress/block-editor` patch disabling `stripExperimentalSettings`. This means the Patterns/Media tabs now function as desired in the bundled editor. As noted previously, this patch cannot be applied to the remote editor, meaning the tabs do not populate in that environment. We are trading one patch for another, but the outcome is the remote editor, at least, functions, albeit without a few features requiring patches. 

Lastly, since we can no longer set `hasFixedToolbar` on the `BlockEditorProvider`, we rely upon setting `fixedToolbar` in the preferences store.

## Testing Instructions

> [!tip]
> Installable prototypes are available for testing:
> - https://github.com/wordpress-mobile/WordPress-iOS/pull/24151
> - https://github.com/wordpress-mobile/WordPress-Android/pull/21734

**1 - Formatting options remain present within the block toolbar**
1. Start the development server. 
1. Open the editor in your desktop browser.
1. Insert a Paragraph block.
1. Type a couple of words into the block. 
1. Move your mouse cursor.
1. Verify the block toolbar remains populated with rich text formatting
   options—e.g., bold, italic, etc.

**2 - The Patterns/Media tab populate with content from the current site**
1. Open the editor within the WordPress mobile app.
1. Open the block inserter.
1. Tap the Patterns/Media tab.
1. Verify the tab is populated with content from the current site.

N/A, no user-facing changes. 

## Screenshots or screencast 
N/A, no user-facing changes.


patches/@wordpress+block-editor+14.11.0.patch
```diff
@@ -10,3 +10,16 @@ index aa2e9a5..f2d8481 100644
...
+diff --git a/node_modules/@wordpress/block-editor/build-module/components/provider/index.js b/node_modules/@wordpress/block-editor/build-module/components/provider/index.js
+index e89fa49..0f14ae8 100644
+--- a/node_modules/@wordpress/block-editor/build-module/components/provider/index.js
++++ b/node_modules/@wordpress/block-editor/build-module/components/provider/index.js
+@@ -104,7 +104,7 @@ export const ExperimentalBlockEditorProvider = withRegistryProvider(props => {
+ export const BlockEditorProvider = props => {
+   return /*#__PURE__*/_jsx(ExperimentalBlockEditorProvider, {
+     ...props,
+-    stripExperimentalSettings: true,
++    stripExperimentalSettings: false,
```
dcalhoun, line 22: Patch manually enables populating the Patterns and Media tabs within the inserter. As outlined in the PR description, this is a workaround that works for the local editor, but not the remote editor. 

Ideally, we seek the Gutenberg project exposing the APIs we need to avoid this patch.

src/index.jsx
```diff
@@ -39,7 +39,11 @@ function initializeEditor() {
 			dispatch( editorStore ).updateEditorSettings( editorSettings );
 		} );
 
-	dispatch( preferencesStore ).setDefaults( 'core/edit-post', {
+	const preferenceDispatch = dispatch( preferencesStore );
+	preferenceDispatch.setDefaults( 'core', {
+		fixedToolbar: true,
+	} );
```
dcalhoun, line 45: Since we can no longer set `hasFixedToolbar` on the `BlockEditorProvider`, we must rely upon the preferences store.

src/utils/bridge.js
```diff
@@ -165,18 +165,21 @@ export function getPost() {
...
 
 	// Since we don't use the auto-save functionality, draft posts need to have an ID.
 	// We assign a temporary ID of -1.
 	return {
+		id: -1,
 		type: 'post',
 		status: 'auto-draft',
-		id: -1,
+		title: { raw: '' },
+		content: { raw: '' },
```
dcalhoun, line 182: Ensures draft posts do not throw exceptions when referencing values on undefined objects—this only occurred in development environments.

dcalhoun:
@kean would you mind reviewing this when you have time please?

kean (approved):
I smoke tested the changes – looks good.

PR #113, opened 2025-04-01
feat: Filter block types with allow list

dcalhoun:
Related: 

- CMM-224-linear-issue
- https://github.com/Automattic/jetpack/pull/42835

## What?

Filter the available block types via an allow list fetched from the API
endpoint.

## Why?

Fix CMM-224. Used for selectively enabling stable block types and/or removing
unnecessary or problematic assets.

## How?

Given block types are registered via the scripts loaded onto the page, we must
unregister each block type not within the allow list.

## Testing Instructions

This small, conditional-wrapped change is likely safe to merge without additional testing. For posterity, the following are the explicit steps for testing:

1. In the Jetpack mobile app, enable the `Experimental block editor plugins` feature flag. 
2. Smoke test the editor and Jetpack blocks. The REST API endpoint currently allows all of the same block types as before, so there are no changes to observe. 

N/A, no navigation changes.

## Screenshots or screencast 
N/A, no visual changes.

### DIFF
--- a/src/remote.jsx
+++ b/src/remote.jsx
@@ -33,7 +33,11 @@ async function initalizeRemoteEditor() {
 		const { themeStyles, hideTitle, siteApiRoot, siteApiNamespace } =
 			getGBKit();
 		// TODO: Load editor assets within the host app
-		const { styles, scripts } = await apiFetch( {
+		const {
+			styles,
+			scripts,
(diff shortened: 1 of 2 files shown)


jkmassel (approved):
Code makes sense, so I think this is ok to land

PR #114, opened 2025-04-05
refactor: Host provides editor settings

dcalhoun:
Related: 

* CMM-200-linear-issue
* https://github.com/wordpress-mobile/GutenbergKit/issues/105
* https://github.com/wordpress-mobile/WordPress-iOS/pull/24417
* https://github.com/wordpress-mobile/WordPress-Android/pull/21791

## What?

Fetch and cache editor settings before providing them to the editor.

## Why?

It allows improving offline support and performance. Ref CMM-200. Fix #105. 

## How?

Remove editor settings fetch, and replace it with an editor configuration option. 

## Testing Instructions

See testing instructions for sibling PRs: 

* https://github.com/wordpress-mobile/WordPress-iOS/pull/24417
* https://github.com/wordpress-mobile/WordPress-Android/pull/21791

### Accessibility Testing Instructions

N/A, no user-facing changes. 

## Screenshots or screencast 
N/A, no user-facing changes.

PR #117, opened 2025-04-16
Suggestions for #114

jkmassel:
## What?
I had a suggestion for #114 that was longer than would make sense in the suggestions box

## Why?
Prevents invalid states in the editor view controller


dcalhoun (approved):
I like the changes; solid improvements. Thank you for taking time to open a PR.

dcalhoun:
Also noting: I smoke tested both the bundled and site-specific editors—I did not encounter any issues.

PR #126, opened 2025-05-12
perf: Avoid unnecessary editor assets requests

dcalhoun:
## What?

Rely upon a cache to avoid unnecessary network requests during editor initialization. 

## Why?

We invoke `loadEditorAssets` twice—once for i18n modules, and then again for all other `@wordpress` modules. The second invocation is performed immediately after the first, and there is no benefit for requesting a new response. 

## How?

Cache the API endpoint response in memory. Long-term, we plan to cache the response and the assets themselves, which is explored in #107. 

## Testing Instructions

1. Test while using a build of the WordPress app with the Gutenberg/experimental editor `_plugins` debug flag enabled. 
2. Monitor your testing device's network activity. 
3. Open the remote or site-specific editor. 
4. Verify a single request to the editor assets endpoint is performed. 

N/A, no user-facing changes. 

## Screenshots or screencast 
N/A, no user-facing changes.

### DIFF
--- a/src/utils/remote-editor.js
+++ b/src/utils/remote-editor.js
@@ -10,6 +10,12 @@ import { getGBKit } from './bridge';
 import { error } from './logger';
 
 /**
+ * Cache for editor assets to avoid unnecessary network requests
+ * @type {Object|null}
+ */
+let editorAssetsCache = null;
+
+/**
  * @typedef {Object} EditorAssetConfig
  *
  * @property {string[]} allowedBlockTypes Array of allowed block types provided by the API.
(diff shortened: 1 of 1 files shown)


src/utils/remote-editor.js
```diff
@@ -9,11 +9,16 @@ import apiFetch from '@wordpress/api-fetch';
...
+ * Cache for editor assets to avoid unnecessary network requests
+ * @type {Object|null}
+ */
+let editorAssetsCache = null;
+
 /**
  * @typedef {Object} EditorAssetConfig
  *
  * @property {string[]} allowedBlockTypes Array of allowed block types provided by the API.
- * @property {Object}   wpDependencies    WordPress dependencies, empty when allowedPackages is provided.
```
dcalhoun, line 16: This was removed in a separate commit as its usage was removed entirely in an earlier commit.

dcalhoun:
Rebased to exclude the now merged `feat/internationalization` branch.

kean (approved):

PR #135, opened 2025-05-16
fix: Scrolling along edges breaks text editing

dcalhoun:
## What?

Fix perpetually disabled text blocks after scrolling the editor by swiping
outside of the block canvas container.

## Why?

Fix CMM-356. The inability to edit text is unexpected.

## How?

Existing Gutenberg `preventFocusCapture` logic relies upon `pointerdown`
and `pointerup` events for temporarily disabling all text blocks
`contenteditable` status. This is done to mitigate issues with block
selection for `flex` elements.

However, `pointerup` events are not always triggered on touch devices
when a touch turns into a swipe. When you scroll with your finger
outside of the block canvas, the `pointerup` callback is never invoked.
Scrolling twice or more leads to a stale `value` in the callback, and a
state where `contenteditable` is perpetually disabled.

## Testing Instructions

1. Open a post with a few text blocks.
1. Scroll by swiping alongside the edge of the display—i.e., outside of the
   block canvas that contains the blocks.
1. Perform the previous step at least once more.
1. Attempt to focus a text block and modify its content.

N/A, no navigation changes.

## Screenshots or screencast 
N/A, no visual changes.


nbradbury (approved):
Fix is good!  I was able to reproduce the problem in `trunk` but not with this branch :shipit:

PR #140, opened 2025-05-20
fix: Embed renders

dcalhoun:
## What?

Fix broken embed rendering.

## Why?

Embeds unexpectedly launched the iOS browser rather than rendering.
Additionally, embed sizing was incorrect due to missing `body` class and
wrapping elements.

Fix CMM-287. 

## How?

Update iOS WebView request handling logic to only open link-activation requests
in the OS browser.

Add required `body` class for responsive embeds. See https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-support/#responsive-embedded-content.

Add host-specific logic for removing wrapping elements from embed renders.

## Testing Instructions

> [!Tip]
> When testing iOS, use the prototype build: https://github.com/wordpress-mobile/WordPress-iOS/pull/24556

**1: Embeds can be created**
1. Insert an Embed block. 
1. Paste a YouTube URL (or other embed URL) into the input.
1. Verify the embed renders as expected.

**2: Opening a post with an embed renders the embed**
1. Open a post with an existing embed. 
2. Verify the embed renders as expected, the OS browser is not launched. 

N/A, no navigation changes.

## Screenshots or screencast 
N/A, no visual changes.

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
@@ -387,22 +387,13 @@ private final class GutenbergEditorController: NSObject, WKNavigationDelegate, W
             return
         }
 
-        if url.isFileURL || // Local editor file
-            url.scheme == "about" || // Empty request
-            url.scheme == "blob" || // Blob URL, used by inserter iframes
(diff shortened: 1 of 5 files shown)


nbradbury (approved):
Tested fine! :shipit:

PR #142, opened 2025-05-27
fix: Prevent inline image crash

dcalhoun:
## What?

Mitigate native host crashes stemming from the absence of a `multiple`
parameter passed to `openMediaLibrary`. 

## Why?

Fix CMM-375. Fix JETPACK-IOS-1F5E.

The absence of this parameter conflicted with expected argument types in the
host app. It appears different contexts may exclude this value—e.g., when
inserting an inline image within a Paragraph block.

## How?

Set a default `mutliple` value.

## Testing Instructions

> [!Tip]
> Test using the [WP-iOS prototype build](https://github.com/wordpress-mobile/WordPress-iOS/pull/24568#issuecomment-2913144985). 

1. Insert a Paragraph block.
1. Open the text formatting menu (tap the downward chevron in the block
   toolbar).
1. Tap _Inline image_.
1. Verify the app does not crash and one can insert an image via the media picker.

N/A, no navigation changes.

## Screenshots or screencast 
N/A, no visual changes.

### DIFF
--- a/src/components/editor/test/use-media-upload.test.jsx
+++ b/src/components/editor/test/use-media-upload.test.jsx
@@ -124,4 +124,21 @@ describe( 'useMediaUpload', () => {
 
 		expect( openMediaLibrary ).toHaveBeenCalled();
 	} );
+
+	it( 'should always provide a multiple argument to the openMediaLibrary callback', () => {
+		let FilterCallback;
+		addFilter.mockImplementation( ( name, namespace, callback ) => {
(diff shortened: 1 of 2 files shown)


dcalhoun:
@nbradbury whenever you have time, would you be willing to review this? Thank you! 🙇🏻‍♂️

nbradbury (approved):
I reproduced the crash in the existing build, and it's fixed in the prototype build :shipit:

PR #92, opened 2025-02-12
[Build] Update Gradle to 8.12.1 and AGP to 8.8.1

ParaskP7:
Issue: [apps-infra-plans#174](https://github.com/Automattic/apps-infra-plans/issues/174)
Required By (`for local builds`): [WPAndroid##21680](https://github.com/wordpress-mobile/WordPress-Android/pull/21680)

## What?

This PR updates Gradle to [8.12.1](https://docs.gradle.org/8.12.1/release-notes.html) and AGP to [8.8.1](https://developer.android.com/build/releases/gradle-plugin#8-8-1) ([context](https://github.com/wordpress-mobile/WordPress-Android/pull/21515#issuecomment-2545405811)).

---

## Why?

This Gradle and AGP update is required by [WPAndroid##21680](https://github.com/wordpress-mobile/WordPress-Android/pull/21680) to avoid breaking local development abilities as it would severely deteriorate productivity ([context](https://github.com/wordpress-mobile/WordPress-Android/pull/21515#issuecomment-2541439990)).

---

## How?

Updating Gradle and AGP versions that match the [WPAndroid##21680](https://github.com/wordpress-mobile/WordPress-Android/pull/21680) PR related Gradle and AGP versions.

---

## Testing Instructions

1. Verify that all the CI checks are successful.
2. Smoke test the example `app` and verify everything is working as expected[^1].
3. Follow the (`3.`) testing instructions on [WPAndroid##21680](https://github.com/wordpress-mobile/WordPress-Android/pull/21680) to test both apps, Jetpack and WordPress, using the composite builds.
4. Try to test for build time increases both, locally and on CI.
5. Locally, in addition to the debug variant, assemble/lint the release variant, and verify everything is working as expected.
    - Run: `./gradlew assembleRelease`
    - Run: `./gradlew lintRelease`

[^1]: I tried following the [README.md#Android](https://github.com/wordpress-mobile/GutenbergKit?tab=readme-ov-file#android) instructions, starting the development server with the example `app` and all, but I am not sure I am doing it the right way. I get an app being launched with a black screen, which then points me to the browser (opening `Chrome`), where I can then add title/description and do things with this GBKit. Not sure if that's how the example `app` is supposed to be tested locally (please guide me). PS: On `android/local.properties` -> `GUTENBERG_EDITOR_URL`, I am using the network URL that the development server points to. 🤔

---

## Screenshots or screencast  `N/A`

PR #143, opened 2025-05-28
Revert to AGP 8.7.3

nbradbury:
As a companion to [this WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/21917), this PR reverts AGP to v8.7.3. Please refer to that WPAndroid PR for details.

### DIFF
--- a/android/gradle/libs.versions.toml
+++ b/android/gradle/libs.versions.toml
@@ -1,5 +1,5 @@
 [versions]
-agp = "8.8.1"
+agp = "8.7.3"
 kotlin = "1.9.0"
 coreKtx = "1.13.1"
 junit = "4.13.2"


dcalhoun (approved):
I tested running the following and all functioned as expected: 

* The GBK Demo app. 
* WP-Android with a GBK composite build. 
* WP-Android with a GBK binary depedency. 
* `./gradlew assembleRelease`.
* `./gradlew lintRelease`

I note that the [previous Gradle upgrade](https://github.com/wordpress-mobile/GutenbergKit/pull/92) included additional file changes:

* `android/gradle/wrapper/gradle-wrapper.jar`
* `android/gradle/wrapper/gradle-wrapper.properties`
* `android/gradlew`
* `android/gradlew.bat`

I presume we do not need to make similar changes, but @ParaskP7 will you confirm?

ParaskP7:
> I presume we do not need to make similar changes, but @ParaskP7 will you confirm?

You're right @nbradbury , no need to do any Gradle up/downgrade with this change, downgrading and thus touching AGP only should be all we need from this.

PR #146, opened 2025-05-30
Add Configuration Builder (and tests)

jkmassel:
## What?
Android uses a configuration builder which is much cleaner around mutability. This PR adds a configuration builder to iOS as well.

## Why?
The configuration has concerning mutability – it should never change after it's created because it's injected at the start. Once it starts being consumed, changes to it don't apply elsewhere, so by making it immutable it's harder to use it wrong.

## How?
Builder pattern

## Testing Instructions
Covered by new unit tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,9 @@ lint-js: npm-dependencies
 test-js: npm-dependencies
 	npm run test -- run
 
+lint-swift:
+	swift package plugin swiftlint
+
 local-android-library: build
 	echo "--- :android: Building Library"
 	./android/gradlew -p ./android :gutenberg:publishToMavenLocal -exclude-task prepareToPublishToS3
@@ -59,7 +62,7 @@ build-swift-package: build
 	$(call XCODEBUILD_CMD, build)
 
 test-swift-package: build
-	$(call XCODEBUILD_CMD, test)
+	swift test
 
 release:
 	@echo "--- :rocket: Starting GutenbergKit Release Process"
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ build-swift-package: build
 	$(call XCODEBUILD_CMD, build)
 
 test-swift-package: build
-	swift test
+	$(call XCODEBUILD_CMD, test)
 
 release:
 	@echo "--- :rocket: Starting GutenbergKit Release Process"
--- a/ios/Demo-iOS/Sources/ContentView.swift
+++ b/ios/Demo-iOS/Sources/ContentView.swift
@@ -66,20 +66,20 @@ struct ContentView: View {
 private extension EditorConfiguration {
 
     static var template: Self {
-        var configuration = EditorConfiguration.default
-
-        #warning("1. Update the property values below")
+        #warning("1. Update the siteURL and authHeader values below")
         #warning("2. Install the Jetpack plugin to the site")
-        configuration.siteURL = "https://modify-me.com"
-        configuration.authHeader = "Insert the Authorization header value here"
+        let siteUrl: String = "https://modify-me.com"
+        let authHeader: String = "Insert the Authorization header value here"
+        let siteApiRoot: String = "\(siteUrl)/wp-json/"
 
-        // DO NOT CHANGE the properties below
-        configuration.siteApiRoot = "\(configuration.siteURL)/wp-json/"
-        configuration.editorAssetsEndpoint = URL(string: configuration.siteApiRoot)!.appendingPathComponent("wpcom/v2/editor-assets")
-        // The `plugins: true` is necessary for the editor to use 'remote.html'
-        configuration.plugins = true
+        let configuration = EditorConfigurationBuilder()
+            .setSiteUrl(siteUrl)
+            .setAuthHeader(authHeader)
+            .setSiteApiRoot(siteApiRoot)
+            .setEditorAssetsEndpoint(URL(string: siteApiRoot)!.appendingPathComponent("wpcom/v2/editor-assets"))
+            .setShouldUsePlugins(true)
 
-        return configuration
+        return configuration.build()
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -59,4 +62,4 @@ build-swift-package: build
 	$(call XCODEBUILD_CMD, build)
 
 test-swift-package: build
-	$(call XCODEBUILD_CMD, test)
```
dcalhoun: Will this have a negative impact on CI checks? Or does the device used not matter for the tests?

dcalhoun (commented):
At a high level, this makes sense to me. Solid improvements—thank you! Testing the Demo app worked as expected. 

This is likely already understood, but building WP-iOS with these changes fails currently. Before merging this, we should open a sibling PR updating the WP-iOS implementation. 

That said, I wanted to address the following: 

> The configuration has concerning mutability – it should never change after it's created because it's injected at the start. Once it starts being consumed, changes to it don't apply elsewhere, so by making it immutable it's harder to use it wrong.

It's not entirely true that mutations do not have an impact at the moment, because we explicitly rely upon [mutating](https://github.com/wordpress-mobile/GutenbergKit/blob/f26928ac3a15ad619dd4f6ebf02402e385edce6f/ios/Sources/GutenbergKit/Sources/EditorViewController.swift#L239-L241) the configuration to set the [editor settings](https://github.com/wordpress-mobile/WordPress-iOS/blob/121f472ed898d046880d4979345fe2348f94e04e/WordPress/Classes/ViewRelated/NewGutenberg/NewGutenbergViewController.swift#L381-L386) and start the editor after the host app [fetches the settings](https://github.com/wordpress-mobile/WordPress-iOS/blob/121f472ed898d046880d4979345fe2348f94e04e/WordPress/Classes/ViewRelated/NewGutenberg/NewGutenbergViewController.swift#L352-L375). 

I am quite confident that code could/should be refactored to be less complex. There are a couple of nuances to consider: 

1. We ["warm up" the editor](https://github.com/wordpress-mobile/WordPress-iOS/blob/121f472ed898d046880d4979345fe2348f94e04e/WordPress/Classes/ViewRelated/Blog/My%20Site/MySiteViewController.swift#L173) on the _My Site_ tab with an empty configuration; 
2. [Within `init`](https://github.com/wordpress-mobile/WordPress-iOS/blob/121f472ed898d046880d4979345fe2348f94e04e/WordPress/Classes/ViewRelated/NewGutenberg/NewGutenbergViewController.swift#L154-L184), we initialize the editor configuration and view controller; 
3. And [within `viewDidLoad`](https://github.com/wordpress-mobile/WordPress-iOS/blob/121f472ed898d046880d4979345fe2348f94e04e/WordPress/Classes/ViewRelated/NewGutenberg/NewGutenbergViewController.swift#L208-L212), we display an activity indicator and fetch the editor settings. 

These nuances could likely be refactored to embrace immutable configuration while retaining the necessary loading sequence.

jkmassel:
@dcalhoun – I broadly agree with your comment. IMHO having two methods on the editor like `init` and `updateConfiguration` both of which take an immutable configuration would make a ton of sense, and would make it clear that simply updating the configuration object doesn't cause the underlying editor to change in any way.

I'm happy to leave this PR with you, feel free to do with it what you think is best!

Makefile
dcalhoun: I reverted this in https://github.com/wordpress-mobile/GutenbergKit/pull/146/commits/26c3afa3e7a9b9db7f5bce3069d09cad3e5c2aad. It seems this was originally changed when making the test suite pass on macOS (https://github.com/wordpress-mobile/GutenbergKit/pull/146/commits/7dca3ef282d22b10253932efaee4a4dcb7552ce3), which makes sense, but I'm uncertain why we'd remove the specific device for CI runs.

dcalhoun:
@crazytonyli in addition to help with https://github.com/wordpress-mobile/WordPress-iOS/pull/24662#discussion_r2195212119, I welcome your review of these changes now that I've rebased it atop the latest `trunk` branch. Thank you!

crazytonyli (approved):

dcalhoun:
@jkmassel unfortunately, this is a breaking change. Would you be willing to resolve the [incompatible types](https://github.com/wordpress-mobile/WordPress-iOS/pull/24662#discussion_r2195212119) and finish the WP-iOS integration? Alternatively, we could revert this change for the time being.

dcalhoun:
@jkmassel I'll also note that GBK cookie configuration was removed from iOS in https://github.com/wordpress-mobile/GutenbergKit/pull/157. I see that https://github.com/wordpress-mobile/GutenbergKit/pull/146/commits/fc28a0ab89d20d0e1c6c1e37e2beeb45515d73ad re-added iOS cookie configuration. 

We should either remove cookie configuration again or finish reinstating it—ensuring we [set them for the WebView](https://github.com/wordpress-mobile/GutenbergKit/commit/5ec3e9b437879b504cfed7a25ef9c60d05bf1344#diff-7d1c3b45b49105e10e9dc4ac403350f6223668df8bcd7c7d287a3eb202494a2aL40-L45) and updating the WP-iOS integration. 

Currently, WP-iOS relies upon the [shared global cookie jar](https://github.com/wordpress-mobile/WordPress-iOS/blob/248cd1c619f3289ea63ce40376f04b2004742138/WordPress/Classes/ViewRelated/NewGutenberg/NewGutenbergViewController.swift#L393-L413) used by the WebView. GBK's cookie logic clears that out when launching its own WebView. We'd need to update WP-iOS to address this.

dcalhoun:
@jkmassel it looks like this also reinstated the WebView globals removed in https://github.com/wordpress-mobile/GutenbergKit/pull/172. I suggest we remove that too. It is now unused.

PR #150, opened 2025-06-25
fix: Localize remote editor

dcalhoun:
## What?

Fix localization of strings within the remote editor.

## Why?

Close CMM-504. Close CMM-547. Localization worked in the local editor, but not in the remote editor.

## How?

The remote editor relies upon accessing `@wordpress` modules via the
`window.wp` global. Most of these references are handled by
externalization performed by Vite. However, the `@wordpress/i18n` module
is a dependency of `@wordpress/api-fetch`, which we do not want to
externalize. So, the `import` of `@wordpress/i18n` is not correctly
externalized for the remote editor.

As a workaround, we check for and utilize the global if it is available.
Long term, we should align the local and remote editor architecture so
that they load in the same manner. At which point, we could remove the
conditional and always reference the global.

Additionally, I added a guard to referencing the Node.js `process` to avoid
intermittent server errors.

## Testing Instructions

1. Start the `make dev-server-remote` command.
2. Open the iOS demo app and set the `GUTENBERG_EDITOR_REMOTE_URL` environment
   variable to `http://localhost:5174/remote.html`.
3. Apply the below example diff to the iOS demo app, replacing the `<YOUR-SIMPLE-SITE-NAME>`
   and `<REDACTED>` placeholders with the actual site name and auth header.
4. Run the app.
5. Open the editor and verify that the UI strings are in Spanish.

<details><summary>Example Diff</summary>

```diff
diff --git a/ios/Demo-iOS/Gutenberg.xcodeproj/xcshareddata/xcschemes/Gutenberg.xcscheme b/ios/Demo-iOS/Gutenberg.xcodeproj/xcshareddata/xcschemes/Gutenberg.xcscheme
index 1a352df..b6fd3e5 100644
--- a/ios/Demo-iOS/Gutenberg.xcodeproj/xcshareddata/xcschemes/Gutenberg.xcscheme
+++ b/ios/Demo-iOS/Gutenberg.xcodeproj/xcshareddata/xcschemes/Gutenberg.xcscheme
@@ -54,12 +54,12 @@
          <EnvironmentVariable
             key = "GUTENBERG_EDITOR_URL"
             value = "http://localhost:5173/"
-            isEnabled = "NO">
+            isEnabled = "YES">
          </EnvironmentVariable>
          <EnvironmentVariable
             key = "GUTENBERG_EDITOR_REMOTE_URL"
             value = "http://localhost:5174/remote.html"
-            isEnabled = "NO">
+            isEnabled = "YES">
          </EnvironmentVariable>
       </EnvironmentVariables>
    </LaunchAction>
diff --git a/ios/Demo-iOS/Sources/EditorView.swift b/ios/Demo-iOS/Sources/EditorView.swift
index f1f3394..b71ce84 100644
--- a/ios/Demo-iOS/Sources/EditorView.swift
+++ b/ios/Demo-iOS/Sources/EditorView.swift
@@ -72,7 +72,13 @@ private struct _EditorView: UIViewControllerRepresentable {
     var editorURL: URL?
 
     func makeUIViewController(context: Context) -> EditorViewController {
-        let viewController = EditorViewController()
+        var config = EditorConfiguration()
+        config.locale = "es"
+        config.plugins = true
+        config.siteApiNamespace = ["sites/<YOUR-SIMPLE-SITE-NAME>.wordpress.com"]
+        config.siteApiRoot = "https://public-api.wordpress.com/"
+        config.authHeader = "Bearer <YOUR-AUTH-TOKEN>"
+        let viewController = EditorViewController(configuration: config)
         viewController.editorURL = editorURL
         if #available(iOS 16.4, *) {
             viewController.webView.isInspectable = true

```

</details>

N/A, no visual changes. 

## Screenshots or screencast 
N/A, no visual changes.

### DIFF
--- a/vite.config.remote.js
+++ b/vite.config.remote.js
@@ -19,38 +19,6 @@ export default defineConfig( {
 		rollupOptions: {
 			input: resolve( __dirname, 'src/remote.html' ),
 			external: externalize,
-			output: {
-				// Manual chunks are necessary to prevent circular dependencies, some of
-				// which originate from Vite's injected helpers.
-				//
-				// See:
-				// - https://github.com/vitejs/vite/issues/18551
-				// - https://github.com/vitejs/vite/issues/13952
-				// - https://github.com/vitejs/vite/issues/5189#issuecomment-2175410148
-				manualChunks( id ) {
-					const VITE_COMMON_MODULES = [
-						'vite/preload-helper',
-						'vite/modulepreload-polyfill',
-						'vite/dynamic-import-helper',
-						'commonjsHelpers',
-						'commonjs-dynamic-modules',
-						'__vite-browser-external',
-					];
-					if (
-						VITE_COMMON_MODULES.some( ( m ) => id.includes( m ) )
-					) {
-						return 'vite-helpers';
-					}
-
-					if ( id.includes( 'src/utils/bridge.js' ) ) {
-						return 'bridge';
-					}
-
-					if ( id.includes( 'src/utils/logger.js' ) ) {
-						return 'logger';
-					}
-				},
-			},
 		},
 		target: 'esnext',
 	},
--- a/vite.config.remote.js
+++ b/vite.config.remote.js
@@ -67,8 +67,8 @@ function wordPressExternals() {
 
 				if (
 					! externalDefinition ||
-					/@wordpress\/(api-fetch|i18n|url)/.test( id ) ||
-					/@wordpress\/(api-fetch|i18n|url)/.test( module )
+					/@wordpress\/(api-fetch|url|hooks)/.test( id ) ||
+					/@wordpress\/(api-fetch|url|hooks)/.test( module )
 				) {
 					continue; // Exclude the module from externalization
 				}
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -23,32 +23,33 @@ import './index.scss';
 import EditorLoadError from './components/editor-load-error';
 import { error } from './utils/logger';
 
-try {
-	await awaitGBKitGlobal();
-	initializeApiFetch();
-	await configureLocale();
+// Rely upon promises rather than async/await to avoid timeouts caused by
+// circular dependencies. Addressing the circular dependencies is quite
+// challenging due to Vite's preload helpers and bugs in `manualChunks`
+// configuration.
+//
+// See:
+// - https://github.com/vitejs/vite/issues/18551
+// - https://github.com/vitejs/vite/issues/13952
+// - https://github.com/vitejs/vite/issues/5189#issuecomment-2175410148
+awaitGBKitGlobal()
+	.then( initializeApiAndLocale )
+	.then( importEditor )
+	.then( initializeEditor )
+	.catch( handleError );
 
-	// Ensure the correct translation strings are used by postponing the import
-	// of `@wordpress` packages until after the locale is set.
-	//
-	// @TODO: A circular dependency prevents the use of async/await. Ideally, we
-	// address the circular dependency using Rollup's `manualChunks`. However, a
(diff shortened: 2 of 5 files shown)


crazytonyli:
Does the `configuration.locale` only affect the hard-coded strings in the Gutenberg library? Does Gutenberg pass it to the site API requests, so that potentially the block name (for example, "Jetpack AI Assistant") can be localized? Also, why is the "Add a block" title and the tabs below don't use the right language?

dcalhoun:
> Does the `configuration.locale` only affect the hard-coded strings in the Gutenberg library? 

Yes, only strings within the Gutenberg project are localized at this time. We have issues for expanding l10n. 

> Does Gutenberg pass it to the site API requests, so that potentially the block name (for example, "Jetpack AI Assistant") can be localized? 

No, it does not. If this is possible, this is a great idea for localizing plugin-provided strings. 

> Also, why is the "Add a block" title and the tabs below don't use the right language?

This is surprising. I am unable to reproduce this in with either the Demo app or WP-iOS. 🤔 

Do you note any other missing strings—block toolbar menu items, block settings, etc? I presume your test is using the steps I provided? Do you experience the issue when testing with WP-iOS?

crazytonyli:
> Do you note any other missing strings—block toolbar menu items, block settings, etc?

The block is still in English, even though the block inserter is mostly Spanish.



> I presume your test is using the steps I provided?

Yes, I double checked. I followed the steps. Except, my dev server address is at http://localhost:5173/remote.html, the 5174 port does not load for me.

> Do you experience the issue when testing with WP-iOS?

I didn't test WP-iOS with changes in this PR.

vite.config.remote.js
```diff
@@ -99,8 +67,8 @@ function wordPressExternals() {
 
 				if (
 					! externalDefinition ||
-					/@wordpress\/(api-fetch|i18n|url)/.test( id ) ||
-					/@wordpress\/(api-fetch|i18n|url)/.test( module )
```
dcalhoun, line 103: `i18n` was removed to ensure the `localization` module correctly references the `window.wp.i18n` global for the remote editor.

dcalhoun:
> Yes, I double checked. I followed the steps. Except, my dev server address is at http://localhost:5173/remote.html, the 5174 port does not load for me.

Good to know what testing steps you following, thank you. The `5174` port is solely because I generally run both the local and remote editor servers side-by-side. 

> I didn't test WP-iOS with changes in this PR.

Right, I was requesting that you attempt testing in WP-iOS as I am unable to reproduce the issue in both GBK and WP-iOS currently. 

That said, I pushed additional refactors to further align the l10n logic across the local and remote editors, and hope that maybe this also resolves the issue you observe. Would you please test and let me know? Thanks! 🙇🏻‍♂️

crazytonyli:
From the WP-iOS app, I can see the translated "Add a block" text. However, I still see the English text in the Demo-iOS app

crazytonyli (approved):

dcalhoun:
> From the WP-iOS app, I can see the translated "Add a block" text. However, I still see the English text in the Demo-iOS app 

@crazytonyli glad to hear the translated strings display as expected within the WP-iOS app. I'll merge this with that positive testing outcome.  

Given you and I experience different outcomes in the Demo app, it seems to be an environment configuration issue. Does the issue resolve if you run `make build`, which downloads the translation strings?

crazytonyli:
> Does the issue resolve if you run `make build`

I just tried it again, and it resolved the issue. However, I'm pretty sure I tried both ways (`make build` and `make dev-server-remote`) in previous tests too...

dcalhoun:
It makes sense to me that it might solve the problem (missing translations downloaded), but it is strange that the same approach did not work previously... 🤔

PR #153, opened 2025-07-09
Android Asset Caching Implementation

oguzkocer:
## Summary

This PR implements the Android counterpart for iOS editor asset caching (PR #148), enabling performance improvements for remote editor functionality by caching JavaScript, CSS, and source map files locally.

### Key Components

1. **`CachedAssetRequestInterceptor`** - Intercepts WebView requests for cacheable assets
2. **`EditorAssetsLibrary`** - Manages HTTP operations, caching, and cleanup
3. **`EditorConfiguration`** - Extended with caching-related properties
4. **JavaScript Bridge** - Updated to support Android asset fetching

### Architecture Decisions

- **Direct Request Interception**: Uses Android's native WebView request interception instead of URL scheme modification (unlike iOS)
- **Non-blocking Approach**: Serves cached assets immediately while caching new ones in background to prevent ANR
- **Time-based Cleanup**: Removes cache files older than 7 days, leveraging asset versioning for natural cleanup
- **No HTML Parsing**: Simplified implementation that doesn't require parsing HTML or JSON manifests

### Caching Strategy

- **Cacheable Assets**: `.js`, `.css`, `.js.map` files from configured hosts
- **Storage Location**: `context.filesDir/editor-caches/{site-name}` (matches iOS structure)
- **Lifecycle Management**: Proper cleanup on WebView destruction
- **Thread Safety**: All network operations on background threads with proper coroutine management

### Configuration

```kotlin
val config = EditorConfiguration.builder()
    .setEnableAssetCaching(true)
    .setCachedAssetHosts(setOf("wp.com", "wordpress.com"))
    .setEditorAssetsEndpoint("https://example.com/wp-json/wpcom/v2/editor-assets") // Optional
    .build()
```

### URL Construction

Matches iOS implementation:
- Uses `editorAssetsEndpoint` if configured
- Falls back to `{siteApiRoot}wpcom/v2/editor-assets`
- Consistent behavior across platforms


dcalhoun:
Thank you! This caching will have a big impact on performance and stability. 

While testing this, I discovered I recently shipped a regression to WPCOM Simple sites that broke loading of Jetpack blocks in GBK. During your own testing, if you ever successfully loaded the editor for a WPCOM Simple site and found no Jetpack blocks, this regression was likely the cause. I corrected this in https://github.com/Automattic/jetpack/pull/44274. 

## Successful testing

After fixing the regression, **I successfully loaded Jetpack blocks using the caching mechanism—both WPCOM Simple and Atomic sites**. I used the following patch for WP-Android to ensure the correct configuration. 

<details><summary>WP-Android cache configuration diff</summary>

```diff
diff --git a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.kt b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.kt
index 135861253d6..6c9e98c611d 100644
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.kt
@@ -2520,6 +2520,7 @@ class EditPostActivity : BaseAppCompatActivity(), EditorFragmentActivity, Editor
             )
 
             val postType = if (editPostRepository.isPage) "page" else "post"
+            // TODO: Update this to set the correct siteApiRoot for self-hosted sites
             val siteApiRoot = if (isWpCom) "https://public-api.wordpress.com/" else ""
             val authToken = accountStore.accessToken
             val authHeader = "Bearer $authToken"
@@ -2533,6 +2534,7 @@ class EditPostActivity : BaseAppCompatActivity(), EditorFragmentActivity, Editor
                 "postType" to postType,
                 "postTitle" to editPostRepository.getPost()?.title,
                 "postContent" to editPostRepository.getPost()?.content,
+                "siteURL" to siteModel.url,
                 "siteApiRoot" to siteApiRoot,
                 "namespaceExcludedPaths" to arrayOf("/wpcom/v2/following/recommendations", "/wpcom/v2/following/mine"),
                 "authHeader" to authHeader,
diff --git a/libs/editor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergKitEditorFragment.java b/libs/editor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergKitEditorFragment.java
index 4686f4d3b79..5f41b0ccd50 100644
--- a/libs/editor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergKitEditorFragment.java
+++ b/libs/editor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergKitEditorFragment.java
@@ -41,6 +41,7 @@ import org.wordpress.android.util.AppLog;
 import org.wordpress.android.util.AppLog.T;
 import org.wordpress.android.util.PermissionUtils;
 import org.wordpress.android.util.ProfilingUtils;
+import org.wordpress.android.util.UrlUtils;
 import org.wordpress.android.util.helpers.MediaFile;
 import org.wordpress.android.util.helpers.MediaGallery;
 import org.wordpress.aztec.IHistoryListener;
@@ -58,6 +59,7 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 
 import static org.wordpress.gutenberg.Media.createMediaUsingMimeType;
@@ -565,13 +567,19 @@ public class GutenbergKitEditorFragment extends EditorFragmentAbstract implement
             postId = -1;
         }
 
+        var siteURL = (String) mSettings.get("siteURL");
+        var siteApiRoot = (String) mSettings.get("siteApiRoot");
+        var siteApiNamespace = (String[]) mSettings.get("siteApiNamespace");
+        // TODO: Update to set correct endpoint for self-hosted sites
+        var editorAssetsEndpoint = siteApiRoot + "wpcom/v2/" + siteApiNamespace[0] + "/editor-assets";
+
         EditorConfiguration config = new EditorConfiguration.Builder()
                 .setTitle((String) mSettings.get("postTitle"))
                 .setContent((String) mSettings.get("postContent"))
                 .setPostId(postId)
                 .setPostType((String) mSettings.get("postType"))
                 .setThemeStyles((Boolean) mSettings.get("themeStyles"))
-                .setPlugins((Boolean) mSettings.get("plugins"))
+                .setPlugins((Boolean) true)
                 .setSiteApiRoot((String) mSettings.get("siteApiRoot"))
                 .setSiteApiNamespace((String[]) mSettings.get("siteApiNamespace"))
                 .setNamespaceExcludedPaths((String[]) mSettings.get("namespaceExcludedPaths"))
@@ -579,6 +587,9 @@ public class GutenbergKitEditorFragment extends EditorFragmentAbstract implement
                 .setWebViewGlobals((List<WebViewGlobal>) mSettings.get("webViewGlobals"))
                 .setEditorSettings(editorSettings)
                 .setLocale((String) mSettings.get("locale"))
+                .setEditorAssetsEndpoint(editorAssetsEndpoint)
+                .setCachedAssetHosts(Set.of("s0.wp.com", UrlUtils.getHost(siteURL)))
+                .setEnableAssetCaching(true)
                 .build();
 
         mGutenbergView.start(config);

```

</details>

> While testing this in WPAndroid, I noticed that [we may not be using the correct `siteApiRoot`](https://github.com/wordpress-mobile/WordPress-Android/blob/abff1d5cf68e83e6e8684eb718370c8b1c2bbac8/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.kt#L2523) [...]

You are correct. In WP-iOS, we set the [correct API root for self-hosted sites](https://github.com/wordpress-mobile/WordPress-iOS/blob/6af4f4f3bf91c46dc073badf3d56431e95695fb8/WordPress/Classes/ViewRelated/NewGutenberg/NewGutenbergViewController.swift#L845-L847); we need to implement the same in WP-Android now that application passwords are available. 

> Does self-hosted sites support /editor-assets in any form? If so, do we have a static url to provide the assets?

They do, but it requires the Jetpack plugin to be installed and activated, as the editor assets endpoint is not yet available in WordPress Core. The endpoint URL would resemble `https://example.com/wp-json/wpcom/v2/editor-assets`. 

## Suggestions

We might consider implementing similar navigation as the iOS GBK Demo app, where one can test either the bundled editor a [configured remote editor](https://github.com/wordpress-mobile/GutenbergKit/blob/9cb742b38cd5dc1ea242ff943e44a6dfbf8d7d41/ios/Demo-iOS/Sources/ContentView.swift#L71-L74). 

<details><summary>iOS GBK Demo app navigation</summary>
<p>


https://github.com/user-attachments/assets/b3d3be7b-855c-4bc3-bc7a-de113d8ddc14



</p>
</details> 

Given my success testing the current changes, I **suggest we move forward refining the code in preparation for review**. Please let me know if there are any aspects with which you would like my help.

oguzkocer:
> We might consider implementing similar navigation as the iOS GBK Demo app, where one can test either the bundled editor a [configured remote editor](https://github.com/wordpress-mobile/GutenbergKit/blob/9cb742b38cd5dc1ea242ff943e44a6dfbf8d7d41/ios/Demo-iOS/Sources/ContentView.swift#L71-L74).

This suggestion has been implemented in #155 with a minor follow up fix in #156.

oguzkocer:
@dcalhoun This is now ready for review. The force-push was about me trying to fix the lint error, so there are no changes to the existing commits from before your first review.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/CachedAssetRequestInterceptor.kt
```diff
@@ -0,0 +1,91 @@
...
+    override fun handleRequest(request: WebResourceRequest): WebResourceResponse? {
+        val url = request.url?.toString() ?: return null
+
+        try {
+            // Handle manifest endpoint requests
+            if (url.contains("/wpcom/v2/") && url.contains("/editor-assets")) {
+                Log.d(TAG, "Manifest request detected: $url - letting WebView handle normally")
+                // Let WebView make the request normally, we'll cache assets as they're requested
+                return null
+            }
```
dcalhoun, line 44: Why do we intercept this request, but then ultimately let the WebView handle it? Should we merely not [intercept it](https://github.com/wordpress-mobile/GutenbergKit/pull/153/files#diff-6825429178dd961e20714e221f5996b7243fed5d37e14efb0387928409908ee9R21-R24)?

android/Gutenberg/src/main/java/org/wordpress/gutenberg/EditorConfiguration.kt
```diff
@@ -86,6 +89,9 @@ open class EditorConfiguration constructor(
         private var editorSettings: String? = null
         private var locale: String? = "en"
         private var cookies: Map<String, String> = mapOf()
+        private var enableAssetCaching: Boolean = false
+        private var cachedAssetHosts: Set<String> = emptySet()
```
dcalhoun, line 93: The iOS implementation lacks these options. Do you believe we should add them to iOS?

dcalhoun (approved):

android/Gutenberg/src/main/java/org/wordpress/gutenberg/EditorConfiguration.kt
oguzkocer: I am not sure about `cachedAssetHosts`, but I think `enableAssetCaching` is useful. It's much easier to turn off a flag for debugging than to have to do an included build, or publish a different version of the library to disable caching.

What do you think @crazytonyli?

android/Gutenberg/src/main/java/org/wordpress/gutenberg/CachedAssetRequestInterceptor.kt
oguzkocer: This one is interesting and I asked the same question to Claude as to why it picked this implementation. It didn't give the most satisfactory answer, so I tried to identify the differences and ended up keeping it:

1. These requests will never hit the subsequent calls for `getCachedAsset` & `cacheAssetInBackground` resulting in some performance benefits.
2. It makes it very obvious that this is a special url and should never be part of the caching mechanism.

I think the second point is mostly a nice to have, but the first point sold me on it.

dcalhoun: Alright. Thank you for elaborating.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/EditorConfiguration.kt
crazytonyli: I believe we should use the same flag for "asset caching" and "is using remote editor", because we'd always want to cache assets for remote editors.

There is also an `editorAssetsEndpoint` property, which can be used as a flag. Asset caching is only possible if that property is set.

dcalhoun: @crazytonyli I had similar thoughts regarding using `editorAssetsEndpoint` as both the toggle for plugins and the configuration for fetching remote editor assets. My only concern is naming it in a way that clearly communicates its intent and impact. Should we keep `editorAssetsEndpoint` or name it something else—`remoteEditorAssetsEndpoint`, `pluginAssetsEndpoint`, etc? It returns both plugin scripts and core WP scripts. 

My understanding is that `editorAssetsEndpoint` is currently not required as both platforms have a fallback value of `<site-host>/wp-json/wpcom/v2/editor-assets`. I am in favor of removing the fallback and requiring a configured path. This would allow use to remove the [`wpcom` reference](https://github.com/search?q=repo%3Awordpress-mobile%2FGutenbergKit%20wpcom&type=code) from this codebase that we hope to merge into the community-centric Gutenberg project.

oguzkocer: > I believe we should use the same flag for "asset caching" and "is using remote editor", because we'd always want to cache assets for remote editors.

That may be true for production, but I don't agree with it for debugging. It's a good idea to have a way to turn something off without having to change the source code.

---

One improvement that might address our concerns is to group caching-related properties into a single container object. Currently, caching arguments are scattered as separate parameters, making their relationships unclear without prior knowledge.

With a container approach, we can:
- Use constructors that automatically set sensible defaults
- Set some properties based on others internally
- Reduce the number of constructor arguments

```
class CacheConfig {
  constructor(endpoint) {
    this.enabled = true;
    this.endpoint = endpoint;
  }
}
```

PR #154, opened 2025-07-14
Change editor assets to the cache directory

crazytonyli:
See the inline comments for the reason for this change.

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorAssetsLibrary.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorAssetsLibrary.swift
@@ -128,7 +128,7 @@ private extension EditorConfiguration {
 
         return FileManager.default
             // Use the app cache directory to prevent editor assets from being backed up to iCloud.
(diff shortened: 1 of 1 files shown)


ios/Sources/GutenbergKit/Sources/EditorAssetsLibrary.swift
```diff
@@ -127,7 +127,9 @@ private extension EditorConfiguration {
         }
 
         return FileManager.default
-            .urls(for: .applicationSupportDirectory, in:.userDomainMask)
+            // Use the app cache directory to prevent editor assets from being backed up to iCloud.
+            // Set `isExcludedFromBackup = ture` if this directory is changed in the future.
```
dcalhoun:
```suggestion
            // Set `isExcludedFromBackup = true` if this directory is changed in the future.
```

dcalhoun (approved):
Changes make sense to me. Left one suggestion for fixing a typo.

PR #155, opened 2025-07-17
Add multi-site support to Android demo app

oguzkocer:
Similar to the iOS demo app, allows users to add and manage multiple WordPress sites alongside the bundled editor.

## Changes

- **Multi-site configuration**: Users can add remote WordPress sites via FAB button
- **Authentication flow**: Implements WordPress Application Passwords authentication with auto-discovery
- **Site management**: Tap to launch the editor, long-press to delete sites
- **Persistent storage**: Added sites are saved and restored between app launches

## Testing Instructions

1. **Add remote site**: Tap the FAB (+) button and enter a WordPress site URL
2. **Authentication**: App will auto-discover the site's API endpoints, then open a web view for login
3. **Login process**: Enter WordPress credentials, authorize the app, and you'll be redirected back
4. **Site management**: New site appears in the list. Tap to launch the editor, long-press to delete the site
5. **Bundled editor**: Original bundled editor remains available at the top of the list

**Requirements**: WordPress sites must have the Application Passwords feature enabled (WordPress 5.6+ or Jetpack plugin).

### DIFF
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -43,7 +43,7 @@ dependencies {
     implementation(libs.androidx.activity)
     implementation(libs.androidx.constraintlayout)
     implementation(libs.androidx.webkit)
-    implementation("androidx.recyclerview:recyclerview:1.3.2")
+    implementation(libs.androidx.recyclerview)
     implementation(project(":Gutenberg"))
     testImplementation(libs.junit)
     androidTestImplementation(libs.androidx.junit)
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -9,7 +9,7 @@ android {
 
     defaultConfig {
         applicationId = "com.example.gutenbergkit"
-        minSdk = 22
+        minSdk = 24
         targetSdk = 34
         versionCode = 1
         versionName = "1.0"
@@ -44,6 +44,7 @@ dependencies {
     implementation(libs.androidx.constraintlayout)
     implementation(libs.androidx.webkit)
     implementation(libs.androidx.recyclerview)
+    implementation(libs.wordpress.rs.android)
     implementation(project(":Gutenberg"))
     testImplementation(libs.junit)
     androidTestImplementation(libs.androidx.junit)
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -43,6 +43,7 @@ dependencies {
     implementation(libs.androidx.activity)
     implementation(libs.androidx.constraintlayout)
     implementation(libs.androidx.webkit)
+    implementation("androidx.recyclerview:recyclerview:1.3.2")
     implementation(project(":Gutenberg"))
     testImplementation(libs.junit)
     androidTestImplementation(libs.androidx.junit)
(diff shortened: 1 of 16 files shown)


android/app/src/main/java/com/example/gutenbergkit/MainActivity.kt
```diff
@@ -41,6 +78,88 @@ class MainActivity : AppCompatActivity() {
             .setCookies(emptyMap())
             .build()
 
-        gbView.start(config)
+    private fun createRemoteConfiguration(config: ConfigurationItem.RemoteEditor): EditorConfiguration =
+        createCommonConfigurationBuilder()
+            .setPlugins(true) // Enable plugins for remote editor
+            .setSiteURL(config.siteUrl)
+            .setSiteApiRoot(config.siteApiRoot)
+            .setSiteApiNamespace(arrayOf("wp/v2"))
```
dcalhoun, line 86: Setting a namespace is applicable for WPCOM sites communicating through the WPCOM API. We do not need to set a namespace for the default `wp/v2`. 
```suggestion
            .setSiteApiNamespace(arrayOf())
```

android/app/build.gradle.kts
```diff
@@ -9,7 +9,7 @@ android {
 
     defaultConfig {
         applicationId = "com.example.gutenbergkit"
-        minSdk = 22
+        minSdk = 24
```
dcalhoun, line 12: It seems that bumping to 24 broke loading the development server from an IP address on the development machine, that it may include tighter default security. 

It seems that one now has to add an explicit entry for the IP in the network security configuration. Is there a better way to allow the development server to load? 

https://github.com/wordpress-mobile/GutenbergKit/blob/6ae54c92a05445a93c370d569debc581c9ec455f/android/app/src/main/res/xml/network_security_config.xml#L3-L6

dcalhoun (approved):
Thank you! The test plan succeeds for me. I successfully loaded both the bundled and remote editors. 

I left a few comments for your consideration, but they are not blocking.

dcalhoun:
I forgot one thing: I noticed the WebViews appear to linger after closing the editor (visible in Chrome's `chrome://inspect` inspector). Is that expected? 
lingering-web-views

android/app/build.gradle.kts
oguzkocer: This is not something I am familiar with :( I've made this bump because `wordpress-rs` requires it, although I don't remember why.

Unless this is a major issue for you, I think it's fine to go with a temporary solution like the one you mentioned above. If it's major, I can look into it.

dcalhoun: While annoying, it's not a blocker for my workflow; I can work around it. 

The bigger concern is someone less familiar with the project encountering failed development server loads without knowledge of this workaround. Also, we don't want to commit one-off IP addresses to this configuration file. I'll try to find a solution soon.

oguzkocer: OK, please let me know before spending any significant time on this. I'll double check that we need `minSdk = 24` for `wordpress-rs`.

If you want, we can remove `wordpress-rs`, but I find having a proper authentication flow easier to work with than always having to figure out the authentication token, updating a config document etc. Having said that, I don't plan to work in this repo too much, so whatever works best for you is fine with me.

PR #167, opened 2025-08-20
Forward original failed HTTP response to web view

crazytonyli:
The web inspector can now show more helpful error messages when certain assets fail to load.

| Before | After |
| ------ | ----- |
|  |  |

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorAssetsLibrary.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorAssetsLibrary.swift
@@ -81,7 +81,7 @@ public actor EditorAssetsLibrary {
     }
 
     /// Fetches one asset (JavaScript or stylesheet) and caches its content on the device.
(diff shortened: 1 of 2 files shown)


dcalhoun (commented):
Thank you for working to address! 

When testing the GBK Demo app, I noted styles were broken. When inspecting the network activity, it appears the assets are not served from the `gbk-cache-` location. I imagine these two issues are related, and likely cause script load issues as well.

| Before | After |
| - | - |
| Image | Image |

crazytonyli:
@dcalhoun Good catch. I have fixed the issue in the latest commit.

dcalhoun (approved):
Thank you for adding visibility into the failure details. 

I now see the failures relate to mapping files. I only see this when loading a WP.com site. I'm not sure what to make of it just yet. 

Image

Merging this as an improvement. We can debug the map failures separately.

PR #172, opened 2025-08-24
refactor: Remove unnecessary WebView global configuration

dcalhoun:
## What?

Remove the configuration option for setting global values in the WebView. 

## Why?

This logic is no longer used by the host WordPress apps. Instead, globals like this are set from loading JavaScript files into the WebView, rather than configuring globals via GutenbergKit configuration.

## How?

Remove the entire WebView global configuration option. 

## Testing Instructions

Smoke test the editor. 

N/A, no user-facing changes. 

## Screenshots or screencast 
N/A, no user-facing changes.


nbradbury (approved):
The related WPAndroid PRs tested fine so :shipit:

PR #175, opened 2025-09-03
fix: Prevent multiple Android WebView starts

dcalhoun:
## What?

Prevent the WebView loading a URL multiple times during initialization.

## Why?

In service of https://github.com/wordpress-mobile/WordPress-Android/pull/22185. 

This intermittently led to an invisible editor. The editor was hidden during
the second load, but not revealed again after the second load, as the reveal
logic is only run once.

## How?

- Remove the rudimentary preloading from `createAndPreloadWebView`.
- Repurpose `warmup` utility to load WebView and fetch editor assets, similar to
  the approach found for iOS.
- De-duplicate editor load listeners.

## Testing Instructions

See https://github.com/wordpress-mobile/WordPress-Android/pull/22185. 

N/A, no user-facing changes.

## Screenshots or screencast 
N/A, no user-facing changes.


android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -45,7 +45,6 @@ class GutenbergView : WebView {
     private var configuration: EditorConfiguration = EditorConfiguration.builder().build()
 
     private val handler = Handler(Looper.getMainLooper())
-    private var editorDidBecomeAvailable: ((GutenbergView) -> Unit)? = null
```
dcalhoun, line 48: `editorDidBecomeAvailable` was, at some point, duplicated by `editorDidBecomeAvailableListener`. We rely upon the latter, so I removed the former.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -631,7 +615,6 @@ object GutenbergWebViewPool {
     private fun createAndPreloadWebView(context: Context): GutenbergView {
         val webView = GutenbergView(context)
         webView.initializeWebView()
-        webView.loadUrl(ASSET_URL)
```
dcalhoun, line 634: This rudimentary URL loading was problematic. First, it did not take into consideration the editor configuration, often resulting in loading the incorrect URL. Second, its loading triggered the editor reveal logic, the editor was then reverted to an invisible state when once `start` was invoked by the host app.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -584,64 +576,75 @@ class GutenbergView : WebView {
...
+            webView.start(configuration)
+            warmupWebView = webView
+
+            // Schedule cleanup after assets are loaded
+            warmupHandler = Handler(Looper.getMainLooper())
+            warmupRunnable = Runnable {
+                cleanupWarmup()
             }
+            warmupHandler?.postDelayed(warmupRunnable!!, ASSET_LOADING_TIMEOUT_MS)
         }
```
dcalhoun, line 607: Repurposed this unused `warmup` utility with an implementation that matches that found for iOS. We rely upon a hidden WebView to perform all the editor requests, allowing the caching mechanism to occur as it normally would—both the manifest and individual assets.

adalpari (approved):
LGTM and works as expected!

PR #177, opened 2025-09-10
Extract WebBridge to be reused in SwiftUI.WebView

crazytonyli:
This is a quick prototype of creating a SwiftUI version of `EditorViewController`, using the new `SwiftUI.WebView` API in iOS 26. See the new `GutenbergEditor` SwiftUI view.

The idea is to extract the code that interacts with the Gutenberg library web view out of `EditorViewController`, which then can be reused by the `GutenbergEditor` SwiftUI view. `WebBridge` is the beginning of that.

At the moment, `WebBridge` only contains code that configures the Gutenberg editor library. Later, we can add more API to execute JavaScript functions within the web view.

You can try the new SwiftUI view via the demo app, by turning on the "Use SwiftUI WebView" toggle.

## Potential next steps

I don't think we _need_ to have a SwiftUI view at the moment, but I think it's worthwhile extracting some Gutenberg editor specific code out of the `EditorViewController`, which would put the library into a good position to create a SwiftUI view when needed.

So, I guess that could be the next steps? Moving some more code out of `EditorViewController`?


dcalhoun:
Merely closing outdated PRs to reduce noise. Please feel free to restore this branch and/or reopen the PR as desired. Thanks! 🙇🏻‍♂️

PR #178, opened 2025-09-10
Decouple Swiftsoup

jkmassel:
## What?
Moves the `Swiftsoup` dependency out into its own package that can be conditionally used.

## Why?
The library current requires that its hosting app adopt `Swiftsoup`. It's only used for a small corner of the functionality where regex might not properly parse the data we get back from the server. 

## How?
Creates a new `AssetManifestParser` package that can be included (or not). By default, the implementation now uses a regex-based approach, but warns that it's not recommended.

## Testing Instructions
I've added unit tests for both the default and new parsers that cover the parsing better than it was before.

I'd recommend testing it against a site using the assets plugin.

n/a

## Screenshots or screencast
n/a

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
@@ -38,6 +38,9 @@ public struct EditorConfiguration {
     // Cookies
     public let cookies: [HTTPCookie]
 
+    // Parses URLs out of the Asset Manifest HTML
+    public var assetManifestParser: EditorAssetManifestParser
+
     /// Deliberately non-public – consumers should use `EditorConfigurationBuilder` to construct a configuration
     init(
         title: String,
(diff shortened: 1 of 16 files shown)


ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
```diff
@@ -112,7 +119,7 @@ public struct EditorConfiguration {
         return String(data: jsonData, encoding: .utf8) ?? "undefined"
     }
 
-    public static let `default` = EditorConfigurationBuilder().build()
+    public static var `default` = EditorConfigurationBuilder().build()
```
dcalhoun, line 122: Why the change from `let` to `var`?

dcalhoun (commented):
The intent and changes seem sound from review. Unfortunately, the implementation does not work from my testing. The cached assets are never used when inspecting network activity of the WebView. Do you experience this as well? 

I tested by [configuring](https://github.com/wordpress-mobile/GutenbergKit/blob/601a9a88e5774e13e40999b556f7e5da6cc42715/ios/Demo-iOS/Sources/ContentView.swift#L71-L72) the Demo app. I could not test with WP-iOS due to the current incompatibility with WP-iOS with the merge of https://github.com/wordpress-mobile/GutenbergKit/pull/146. 

| Trunk | Changes |
| - | - |
| Image | Image |

dcalhoun:
Merely closing outdated PRs to reduce noise. Please feel free to restore this branch and/or reopen the PR as desired. Thanks! 🙇🏻‍♂️

PR #182, opened 2025-09-29
fix: Force visual editor mode to ensure consistency with native host UI

dcalhoun:
## What?

Avoid mismatches between native UI and the editor's mode. Ensure the native host
receives an `onEditorLoaded` event.

## Why?

Fix CMM-795. 

GutenbergKit provides no way to configure the initial editor mode. When
GutenbergKit persists a different mode in its local storage, this leads
to a mismatch in native UI and the editor mode. Defaulting to visual avoids
any mismatches.

Additionally, forcing visual mode on initialization mitigates the fact
that the current logic does not trigger the `onEditorLoaded` event when
the text editor loads. This led to the native host never receiving the
event. The result was an indefinite loading indicator until the user
toggled the editor mode.

## How?

Update the preference store on initialization.

## Testing Instructions

> [!tip]
> Testing using the prototype build from https://github.com/wordpress-mobile/WordPress-iOS/pull/24905. 

1. Open the editor.
1. Toggle the mode to _Code Editor_.
1. Close the editor.
1. Open the editor.
1. Verify the editor visual mode loads.

N/A, no navigation changes.

## Screenshots or screencast 
N/A, no visual changes.

### DIFF
--- a/src/utils/editor.jsx
+++ b/src/utils/editor.jsx
@@ -35,6 +35,11 @@ export function initializeEditor( { allowedBlockTypes } = {} ) {
 		themeStyles,
 	} );
 
+	// Force visual mode on initialization to ensure consistency with native UI,
+	// which defaults to visual mode. In the future, we could allow the native
+	// host app to configure the initial editor mode.
+	preferenceDispatch.set( 'core', 'editorMode', 'visual' );
+
 	registerCoreBlocks();
 	unregisterDisallowedBlocks( allowedBlockTypes );
 	const post = getPost();
--- a/src/utils/editor.test.jsx
+++ b/src/utils/editor.test.jsx
@@ -1,0 +1,154 @@
+/**
+ * External dependencies
+ */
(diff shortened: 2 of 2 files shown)


kean:
> When GutenbergKit persists a different mode in its local storage, this leads to a mismatch in native UI and the editor mode. 

I was not able to reproduce the issue described in https://linear.app/a8c/issue/CMM-795/closing-gutenbergkit-while-using-the-code-editor-causes-infinite-load. I used the following steps:

- Create a new post
- Switch to "Code Editor"
- Save draft
- Re-open the editor

I see the visual editor open.

I just wanted to ensure this change is necessary before approving (the changes look good).

dcalhoun:
@kean be sure to use the "remote" (`plugins` configuration) editor. Sorry for not communicating that in testing steps. Let me know if that does not allow reproduction.

kean (approved):
Ah, gotcha

PR #186, opened 2025-10-06
feat: Send modal dialog state to host app

### LINKED ISSUE #129: Improve the Android hardware back navigation experience

The Android platform includes the ability to navigate backwards via a hardware button or swipe gesture. The user expectation is that performing this action dismisses modals within the editor. Currently it exits the editor entirely. 

We'll need to track modal visibility in the native logic and/or communicate back navigation events to the WebView. There is likely pros and cons to either approach. I'm guessing the latter is the best approach; we'd need to relinquish backwards navigation back to the native host if now modal is visible.

dcalhoun:
Related:

- https://github.com/wordpress-mobile/WordPress-iOS/pull/24917
- https://github.com/wordpress-mobile/WordPress-Android/pull/22260

## What?

Send web modal dialog state to the host app.

## Why?

Close CMM-822. Close CMM-444. Close #129. 

Enable host apps to respond to a web modal dialog opening or closing. E.g., it
may be prudent to disable surrounding native UI while a web modal dialog is
open.

## How?

Add bridge events for opening and closing modal dialogs.

## Testing Instructions

> [!important]
> When testing the iOS Demo app, you may need to run `make build` or use the development server before running the Demo app. 

### 1. Web modal dialogs disable native header navigation
1. Open the Demo app.
1. Open a modal dialog—e.g., block inserter, block inspector.
1. Verify the native header navigation is disabled.
1. Close the modal dialog.
1. Verify the native header navigation is enabled.

### 2. Android hardware back button dismisses web dialogs
1. Open the Demo app.
1. Open a modal dialog—e.g., block inserter, block inspector.
1. Press the hardware back button. 
1. Verify the web dialog is dismissed. 
1. Press the hardware back button. 
1. Verify the editor is dismissed. 

Perform the testing via a screen reader.

## Screenshots or screencast 
N/A

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorViewControllerDelegate.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorViewControllerDelegate.swift
@@ -40,6 +40,16 @@ public protocol EditorViewControllerDelegate: AnyObject {
     ///
     /// - parameter type: The type of autocompleter that was triggered (e.g., "plus-symbol", "at-symbol").
     func editor(_ viewController: EditorViewController, didTriggerAutocompleter type: String)
+
+    /// Notifies the client that a modal dialog has opened in the web editor.
+    ///
+    /// - parameter dialogType: The type of modal dialog that opened (e.g., "block-inserter", "media-library").
+    func editor(_ viewController: EditorViewController, didOpenModalDialog dialogType: String)
+
+    /// Notifies the client that a modal dialog has closed in the web editor.
+    ///
+    /// - parameter dialogType: The type of modal dialog that closed (e.g., "block-inserter", "media-library").
+    func editor(_ viewController: EditorViewController, didCloseModalDialog dialogType: String)
 }
 #endif
 
--- a/ios/Sources/GutenbergKit/Sources/EditorViewControllerDelegate.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorViewControllerDelegate.swift
@@ -1,6 +1,7 @@
 import Foundation
 
 #if canImport(UIKit)
+@MainActor
 public protocol EditorViewControllerDelegate: AnyObject {
     /// Called when the editor loads.
     func editorDidLoad(_ viewContoller: EditorViewController)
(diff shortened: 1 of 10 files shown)


ios/Sources/GutenbergKit/Sources/EditorViewControllerDelegate.swift
```diff
@@ -1,6 +1,7 @@
 import Foundation
 
 #if canImport(UIKit)
+@MainActor
```
dcalhoun, line 4: Added in https://github.com/wordpress-mobile/GutenbergKit/pull/186/commits/bd8f3c08d493ee2fcee53bb9efeb7f4cf5e2bf17 to address conformance warning.

adalpari (approved):
LGTM! Tested here: https://github.com/wordpress-mobile/WordPress-Android/pull/22260

src/components/editor/use-host-bridge.js
```diff
@@ -93,6 +93,21 @@ export function useHostBridge( post, editorRef ) {
 			switchEditorMode( mode );
 		};
 
+		window.editor.dismissTopModal = () => {
+			const target =
+				editorRef.current.ownerDocument.activeElement || document.body;
+			target.dispatchEvent(
```
kean, line 99: What does this line do?

kean (approved):

src/components/editor/use-host-bridge.js
dcalhoun: It retrieves the relative document's active element, or falls back to the body. Subsequent lines then trigger a <kbd>Escape</kbd> event on the element to dismiss active modals. 

The reason this line uses [`ownerDocument`](https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument) is because the Gutenberg project contains a [lint rule discouraging use of `document.activeElement`](https://github.com/WordPress/gutenberg/blob/5992373408d388c6cd9a190b495bd5318cbfdbbb/packages/eslint-plugin/rules/no-global-active-element.js). This is discouraged as avoiding globals is a general good practice and, more importantly, due to Gutenberg relying upon `iframe`s, where querying `document` may return an unexpected value.

PR #191, opened 2025-10-08
feature: Auto-focus when creating a new post

kean:
## What?
The editor will now automatically focus when you open create a new post. This is the behavior you expect on iOS, and it matches Gutenberg Mobile.

This change is part of https://github.com/wordpress-mobile/GutenbergKit/pull/162.

## How?
Programmatically click on the first editable element.

## Testing Instructions
- Open an editor with empty content and verify that the editor automatically becomes focused
- Open an editor with non-empty content and verify that it doesn't

## Screenshots or screencast 

https://github.com/user-attachments/assets/ed29735f-10e3-4528-a33d-f375be5afba8

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
@@ -16,6 +16,9 @@ public struct EditorConfiguration: Sendable {
     public let shouldUsePlugins: Bool
     /// Toggles visibility of the title field
     public let shouldHideTitle: Bool
+    /// If enabled, the editor automatically becomes focused when initialized
+    /// with an empty content.
+    public let shouldAutoFocus: Bool
     /// Root URL for the site
     public let siteURL: String
     /// Root URL for the site API
@@ -42,6 +45,7 @@ public struct EditorConfiguration: Sendable {
         shouldUseThemeStyles: Bool,
         shouldUsePlugins: Bool,
         shouldHideTitle: Bool,
+        shouldAutoFocus: Bool,
         siteURL: String,
         siteApiRoot: String,
         siteApiNamespace: [String],
@@ -58,6 +62,7 @@ public struct EditorConfiguration: Sendable {
         self.shouldUseThemeStyles = shouldUseThemeStyles
         self.shouldUsePlugins = shouldUsePlugins
         self.shouldHideTitle = shouldHideTitle
+        self.shouldAutoFocus = shouldAutoFocus
         self.siteURL = siteURL
         self.siteApiRoot = siteApiRoot
         self.siteApiNamespace = siteApiNamespace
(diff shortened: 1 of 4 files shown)


ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
```diff
@@ -16,6 +16,9 @@ public struct EditorConfiguration {
     public let shouldUsePlugins: Bool
     /// Toggles visibility of the title field
     public let shouldHideTitle: Bool
+    /// If enabled, the editor automatically becomes focused when initialized
+    /// with an empty content.
+    public let shouldAutoFocus: Bool
```
kean: I'm not sure if it needs to be an option.

dcalhoun: No, I don't perceive a relevant use case for disabling it now. In the name of keeping things simple, I suggest we exclude it unless there is a clear need. WDYT?

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -347,6 +320,24 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
...
+
+    private func autoFocusEditor() {
+        evaluate("""
+        (function() {
+            const editable = document.querySelector('[contenteditable="true"]');
+            if (editable) {
+                editable.focus();
+                editable.click();
+            }
+        })();
```
dcalhoun: Rather than inlining JavaScript, does it makes sense to expose this as a bridge method in `use-host-bridge.js`? It might also simplify adding Android support—are you open to implementing Android support yourself?

dcalhoun (approved):
Test plan succeeded for me. I left a few suggestions for your consideration.

ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
kean: Sure, I removed it.

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
kean: I moved it to JS. Do I need to rebuild the app and include the changes in the PR?
I don't have anything setup to do Android development right now.

dcalhoun: No, the current practice is using the [release script](https://github.com/wordpress-mobile/GutenbergKit/blob/9aa87b6ce2eff49a3edf24400446212f7e8ad092/docs/releases.md) to generate a tag for integration. The release script builds the JS for iOS. 

I.e., whenever we integrate these changes via an WordPress-iOS PR, we'll use a tag and the JS changes will be included.

dcalhoun: Added matching Android functionality in 3c5ca0035ca068c245f3afbf480763ed845c2e47. Thanks for the help!

PR #196, opened 2025-10-20
Add basic block inserter features: grid, search, icons, ordering

kean:
## Changes 

- Add `BlockInserterView` and a set or related types to render the blocks in a grid (closes [CMM-857](https://linear.app/a8c/issue/CMM-857/show-blocks-in-a-grid))
- Add context menus for blocks (closes [CMM-863](https://linear.app/a8c/issue/CMM-863/add-context-menu-and-details-view-for-blocks))
- Make the inserter searchable (closes [CMM-858](https://linear.app/a8c/issue/CMM-858/make-block-inserter-searchable))

## Testing Instructions

- Enable "Native Inserter"
- Tap "+"
- **Verify** that the new inserter is shown
- **Verify** that you can long-press on a block to reveal more details
- **Verify** that you can search for a block using it's name, title, keywords or category (search weights defined in that order) 
- **Verify** that Dynamic Type and Dark Mode are supported

**Note**: you can not yet tap to select a block.

## Technical Notes

I added [SVGKit](https://github.com/SVGKit/SVGKit) to render block icons. It seems like the only viable option.

## Limitations

- `BlockInserterViewModel` establishes _some_ mobile-specific ordering for blocks, but I opened a separate ticket ([CMM-875](https://linear.app/a8c/issue/CMM-875/establish-the-mobile-specific-order-for-blocks)) to further enhance it (see screenshots/comments). I'm considering implementing it on the JS level.
- There is some non-localized text in this PR. I opened [CMM-874](https://linear.app/a8c/issue/CMM-874/add-localization-for-category-names-etc).
- Some block icons contain incorrect SVG ([CMM-873](https://linear.app/a8c/issue/CMM-873/svg-for-icons-for-some-blocks-is-invalid)). I'm considering fixing it at the source and/or patching them in GBK.

## Screenshots

Screenshot 2025-10-21 at 12 59 49 PM Screenshot 2025-10-21 at 1 02 41 PM 
Screenshot 2025-10-21 at 1 02 46 PM  Screenshot 2025-10-21 at 1 04 00 PM


ios/Sources/GutenbergKit/Sources/Helpers/SearchEngine.swift
```diff
@@ -0,0 +1,190 @@
+import Foundation
+
+/// A protocol for items that can be searched
+protocol Searchable {
```
kean, line 4: **Disclaimer**: other code in PR is written by hand, but this file is entirely generated. It uses Levenshtein distance for fuzzy matching, and does scoring depending on weight of searchable fields. It seems to work pretty well, and we do need it in the inserter.

src/utils/blocks.js
```diff
@@ -24,3 +25,65 @@ export function unregisterDisallowedBlocks( allowedBlockTypes ) {
...
 }
+
+/**
+ * Extract and serialize a block's icon.
+ *
+ * @param {Object} blockType The block type object.
+ *
+ * @return {string|null} The serialized icon string or null.
+ */
+function getBlockIcon( blockType ) {
```
kean, line 36: OK, I generated this too. I might need some help verifying if it makes sense (it works).

src/utils/bridge.js
```diff
@@ -93,15 +93,7 @@ export function onBlocksChanged( isEmpty = false ) {
...
-	const blocks = getBlockTypes().map( ( blockType ) => {
-		return {
-			name: blockType.name,
-			title: blockType.title,
-			description: blockType.description,
-			category: blockType.category,
-			keywords: blockType.keywords || [],
-		};
-	} );
+	const blocks = getSerializedBlocks();
```
dcalhoun: I haven't reviewed all the code yet, but wanted to note we need to rebase to include #198 and adapt this implementation so that the remote editor works. 

Good news is that the native inserter does work with the remote editor in my quick merge conflict resolution on my local machine.

ios/Sources/GutenbergKit/Sources/Helpers/SearchEngine.swift
dcalhoun: Searching worked relatively well from my testing. I noted one thing while searching:  we are not honoring the [`inserter`](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#inserter) supports attribute for blocks. Blocks setting this to `false` should not appear in the block inserter or its search results.

src/utils/blocks.js
dcalhoun: Yes, it makes sense. It seems necessary given the desired outcome of providing SVG strings to the native host app. I cannot think of a better alternative or specific problems with the approach at this time.

src/utils/blocks.js
```diff
@@ -24,3 +25,63 @@ export function unregisterDisallowedBlocks( allowedBlockTypes ) {
...
+export function getSerializedBlocks() {
+	return getBlockTypes().map( ( blockType ) => {
+		return {
+			name: blockType.name,
+			title: blockType.title,
+			description: blockType.description,
+			category: blockType.category,
+			keywords: blockType.keywords || [],
+			icon: getBlockIcon( blockType ),
+		};
```
dcalhoun, line 85: As noted in my comment regarding search, we likely need to include the [`inserter`](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#inserter) supports attribute so that we can honor that setting.

dcalhoun (approved):
Approving as this works well from my testing and other PRs build atop this. My inline comments note one issue we should address regarding inserter block inclusion.

src/utils/blocks.js
kean: I added it as a [bug](https://linear.app/a8c/issue/CMM-888/add-support-for-the-inserter-attribute) and will address in the upcoming PRs. Thanks!

dcalhoun:
Noting that this may be a breaking change for `WordPress-iOS`. When we integrate the latest GBK changes, we'll need to address this:

```
/Library/Developer/Xcode/DerivedData/WordPress-bwtwzztzxastvrfeguwfttopcahk/SourcePackages/checkouts/SVGKit/Source/include/Comment.h:17:12 'Comment' has different definitions in different modules; first difference is definition in module 'SVGKit' found super class with type 'CharacterData'
```

Issue #106, opened 2025-02-26
Tracking: Upstream Gutenberg issues


dcalhoun:
The issue serves as a single location to track bugs, limitations, or improvement opportunities originating from/relating to the upstream WordPress or Gutenberg projects. We will likely create individual issues in the upstream repository and list them here. 

## Bugs and Limitations

* We need a [endpoint exposing editor assets](https://github.com/wordpress-mobile/block-editor-assets-endpoint) in core to support a remote/site-specific editor for all WordPress sites without requiring a plugin (see pbArwn-6Fz-p2). 
* We rely upon [several private APIs](https://github.com/search?q=repo%3Awordpress-mobile%2FGutenbergKit+unlock%28&type=code) or workaround non-existent APIs; we need to identify ways to avoid doing this altogether (see p1726742054662019/1726235525.551819-slack-C04PWEZSYFL):
  * [**WORKAROUND**](https://github.com/wordpress-mobile/GutenbergKit/pull/108) | `useBlockEditorSettings` is unavailable as a public (or private) API (see p1726247399840659/1726235525.551819-slack-C04PWEZSYFL)
  * We rely upon the private [ExperimentalBlockCanvas](https://github.com/wordpress-mobile/GutenbergKit/blob/f26928ac3a15ad619dd4f6ebf02402e385edce6f/src/components/visual-editor/index.jsx#L31).
  * We [manually inject](https://github.com/wordpress-mobile/GutenbergKit/blob/947c5315198eea96e3e8c32fcc6afa59ef242638/src/components/visual-editor/use-editor-styles.js#L65-L68) "common" WordPress [layout styles](https://github.com/WordPress/wordpress-develop/blob/8b5816b914a55aee70058fb484faa9cb439f3da2/src/wp-includes/block-supports/layout.php#L9-L194) that appear absent outside of the WP Admin environment. 
[...]

PR #199, opened 2025-10-22
Native Inserter: Implement block insertion

kean:
## What?
- Close [CMM-867: Add support for inserting selected blocks at the selection point](https://linear.app/a8c/issue/CMM-867/add-support-for-inserting-selected-blocks-at-the-selection-point)
- Close [CMM-866: Add support for dynamic blocks – depending on selection](https://linear.app/a8c/issue/CMM-866/add-support-for-dynamic-blocks-depending-on-selection)
- Fix [CMM-888: Add support for the `inserter` attribute](https://linear.app/a8c/issue/CMM-888/add-support-for-the-inserter-attribute)
- Fix [CMM-905: Remove "Pattern Placeholder"](https://linear.app/a8c/issue/CMM-905/remove-pattern-placeholder)
- Fix [CMM-904: Add support for deprecated blocks](https://linear.app/a8c/issue/CMM-904/add-support-for-deprecated-blocks)

## How?

(see comments)

## Testing Instructions
- **Verify** that the selected block is inserted after the currently selected block. If the selected block is an empty paragraph, replace it with a newly created block.
- **Verify** that when you open and close the inserter, it restores the focus

**Note**: there may be other scenarios I'm not thinking of, including some of the composite blocks like "Gallery" or "Columns". For "Columns", I already opened [CMM-880](https://linear.app/a8c/issue/CMM-880/browse-all-does-not-work-when-adding-blocks-to-composite-blocks-like).

## Screenshots or screencast 

https://github.com/user-attachments/assets/7b3c0b64-dd23-4adc-bc99-889ad24b780b

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
@@ -174,6 +174,12 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         try await webView.evaluateJavaScript("editor.getContent();") as! String
     }
 
+    public struct EditorTitleAndContent: Decodable {
+        public let title: String
+        public let content: String
+        public let changed: Bool
+    }
+
     /// Returns the current editor title and content.
     public func getTitleAndContent() async throws -> EditorTitleAndContent {
         let result = try await webView.evaluateJavaScript("editor.getTitleAndContent();")
@@ -250,12 +256,11 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
 
         let host = UIHostingController(rootView: NavigationStack {
             BlockInserterView(
-                blocks: data.blocks,
-                destinationBlockName: data.destinationBlockName,
+                sections: data.sections,
                 mediaPicker: mediaPicker,
                 presentationContext: context,
                 onBlockSelected: { [weak self] block in
-                    self?.insertBlockFromInserter(block.name)
+                    self?.insertBlockFromInserter(block.id)
                 },
                 onMediaSelected: {
                     print("insert media:", $0)
@@ -268,8 +273,8 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         present(host, animated: true)
     }
 
-    private func insertBlockFromInserter(_ blockName: String) {
-        evaluate("window.blockInserter.insertBlock('\(blockName)')")
+    private func insertBlockFromInserter(_ blockID: String) {
+        evaluate("window.blockInserter.insertBlock('\(blockID)')")
     }
 
     private func openMediaLibrary(_ config: OpenMediaLibraryAction) {
--- a/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
@@ -245,16 +245,17 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
 
     // MARK: - Internal (Block Inserter)
 
-    private func showBlockInserter(blocks: [EditorBlock]) {
+    private func showBlockInserter(data: EditorJSMessage.ShowBlockInserterBody) {
         let context = MediaPickerPresentationContext()
 
         let host = UIHostingController(rootView: NavigationStack {
             BlockInserterView(
-                blocks: blocks,
+                blocks: data.blocks,
+                destinationBlockName: data.destinationBlockName,
                 mediaPicker: mediaPicker,
                 presentationContext: context,
-                onBlockSelected: {
-                    print("insert blocks:", $0)
+                onBlockSelected: { [weak self] block in
+                    self?.insertBlockFromInserter(block.name)
                 },
                 onMediaSelected: {
                     print("insert media:", $0)
@@ -265,6 +266,10 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         context.viewController = host
 
         present(host, animated: true)
+    }
+
+    private func insertBlockFromInserter(_ blockName: String) {
+        evaluate("window.blockInserter.insertBlock('\(blockName)')")
     }
 
     private func openMediaLibrary(_ config: OpenMediaLibraryAction) {
@@ -309,7 +314,7 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
                 delegate?.editor(self, didLogException: editorException)
             case .showBlockInserter:
                 let body = try message.decode(EditorJSMessage.ShowBlockInserterBody.self)
-                showBlockInserter(blocks: body.blocks)
+                showBlockInserter(data: body)
             case .openMediaLibrary:
                 let config = try message.decode(OpenMediaLibraryAction.self)
                 openMediaLibrary(config)
--- a/src/components/editor-toolbar/index.jsx
+++ b/src/components/editor-toolbar/index.jsx
@@ -81,13 +81,11 @@ const EditorToolbar = ( { className } ) => {
 	const classes = clsx( 'gutenberg-kit-editor-toolbar', className );
 
 	const addBlockButton = enableNativeBlockInserter ? (
-		<ToolbarButton
+		<Button
 			title={ __( 'Add block' ) }
 			icon={ plus }
-			onClick={ () => {
-				if ( isInserterOpened ) {
-					setIsInserterOpened( false );
-				}
+			onTouchEnd={ ( e ) => {
+				e.preventDefault();
 				showBlockInserter();
 			} }
 			className="gutenberg-kit-add-block-button"
--- a/src/components/editor-toolbar/index.jsx
+++ b/src/components/editor-toolbar/index.jsx
@@ -17,7 +17,7 @@ import {
 	ToolbarButton,
 } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
-import { close, cog, plus } from '@wordpress/icons';
+import { close, cog } from '@wordpress/icons';
 import clsx from 'clsx';
 import { store as editorStore } from '@wordpress/editor';
 
@@ -27,8 +27,8 @@ import { store as editorStore } from '@wordpress/editor';
 import './style.scss';
 import { useModalize } from './use-modalize';
 import { useModalDialogState } from '../editor/use-modal-dialog-state';
-import { showBlockInserter, getGBKit } from '../../utils/bridge';
-import BlockInserterBridge from '../block-inserter-bridge';
+import { getGBKit } from '../../utils/bridge';
+import NativeBlockInserterButton from '../native-block-inserter-button';
 
 /**
  * Renders the editor toolbar containing block-related actions.
@@ -82,15 +82,7 @@ const EditorToolbar = ( { className } ) => {
 	const classes = clsx( 'gutenberg-kit-editor-toolbar', className );
 
 	const addBlockButton = enableNativeBlockInserter ? (
-		<Button
-			title={ __( 'Add block' ) }
-			icon={ plus }
-			onTouchEnd={ ( e ) => {
-				e.preventDefault();
-				showBlockInserter();
-			} }
-			className="gutenberg-kit-add-block-button"
-		/>
+		<NativeBlockInserterButton />
 	) : (
 		<Inserter
 			popoverProps={ {
@@ -104,7 +96,6 @@ const EditorToolbar = ( { className } ) => {
 
 	return (
 		<>
-			{ enableNativeBlockInserter && <BlockInserterBridge /> }
 			<Toolbar
 				className={ classes }
 				label="Editor toolbar"
--- a/src/components/editor-toolbar/index.jsx
+++ b/src/components/editor-toolbar/index.jsx
@@ -28,6 +28,7 @@ import './style.scss';
 import { useModalize } from './use-modalize';
 import { useModalDialogState } from '../editor/use-modal-dialog-state';
 import { showBlockInserter, getGBKit } from '../../utils/bridge';
+import BlockInserterBridge from '../block-inserter-bridge';
 
 /**
  * Renders the editor toolbar containing block-related actions.
@@ -103,6 +104,7 @@ const EditorToolbar = ( { className } ) => {
 
 	return (
 		<>
+			{ enableNativeBlockInserter && <BlockInserterBridge /> }
 			<Toolbar
 				className={ classes }
 				label="Editor toolbar"
--- a/src/components/native-block-inserter-button/index.jsx
+++ b/src/components/native-block-inserter-button/index.jsx
@@ -33,7 +33,7 @@ import useBlockTypesState from '@wordpress/block-editor/build-module/components/
  */
 import { store as blockEditorStore } from '@wordpress/block-editor';
 import { debug } from '../../utils/logger';
-import { serializeBlocksForNative } from '../../utils/blocks';
+import { preprocessBlockTypesForNativeInserter } from '../../utils/blocks';
 import { showBlockInserter } from '../../utils/bridge';
 
 /**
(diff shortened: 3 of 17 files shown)


ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -240,11 +240,17 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
     // MARK: - Internal (Block Inserter)
 
     private func showBlockInserter(blocks: [EditorBlock]) {
-        present(UIHostingController(rootView: NavigationView {
-            BlockInserterView(blocks: blocks) {
-                print("did select:", $0)
+        // Capture the selected block ID before showing the inserter
+        // (the editor will lose focus when the modal appears)
+        Task { @MainActor in
+            let selectedBlockID = await getSelectedBlockID()
```
kean: The alternative was to memorize the `selectedBlockID` inside the editor, but I thought it would be more flexible to provide the native side with a capability to get the ID of the currently selected block and a way to insert block at any position.

src/components/editor-toolbar/index.jsx
```diff
@@ -81,13 +82,11 @@ const EditorToolbar = ( { className } ) => {
...
 	const addBlockButton = enableNativeBlockInserter ? (
-		<ToolbarButton
+		<Button
 			title={ __( 'Add block' ) }
 			icon={ plus }
-			onClick={ () => {
-				if ( isInserterOpened ) {
-					setIsInserterOpened( false );
-				}
+			onTouchEnd={ ( e ) => {
```
kean: `Button` with `onTouchEnd` is what works on mobile without losing focus.

src/components/editor-toolbar/index.jsx
```diff
@@ -81,13 +81,11 @@ const EditorToolbar = ( { className } ) => {
...
```
dcalhoun: The usage of `onTouchEnd` does introduce one downside (there may be others I yet to notice): you cannot cancel the interaction by dragging your finger away from the button. 

My perception is that is simply the nature of `onTouchEnd`, its callback is triggered regardless of the position of the user's finger when it is triggered. Contrastingly, the `onClick` callback is only triggered when a full "click" occurs, which requires the end of the interaction to be atop the target button element. 

We may be able to add logic to track and compare the touch start and end positions, or the element beneath the latter. Or Claude mentioned we could potentially use `onClick` and pair it with `onMouseDown` invoking `preventDefault()` to prevent focus changes.

While the `onTouchEnd` approach may work for a GutenbergKit-specific button, I'm concerned it may not be an avenue we could use for other buttons rendered within Gutenberg components—e.g., dropdown menu toggles and items.

src/components/editor/use-host-bridge.js
```diff
@@ -118,6 +121,44 @@ export function useHostBridge( post, editorRef ) {
...
+			selectedBlockClientId = null
+		) => {
+			const targetBlockClientId =
+				selectedBlockClientId || getSelectedBlockClientId();
+			const block = createBlock( blockName );
+
+			if ( targetBlockClientId ) {
+				const selectedBlock = getBlock( targetBlockClientId );
+
+				// If the selected block is an empty paragraph, replace it
```
dcalhoun: It's functionality like this that make the native inserter concerning. How many are there that we may need to reimplement and maintain? 

I noticed one missing that relates to the current block (similar to the noted composite blocks). For example, when inside a List (technically focused on a List Item), the inserter should display List Item as a top item. Additionally, selecting other block types that cannot be placed in a List block results in a new block inserted after the currently selected List. 

https://github.com/user-attachments/assets/9cac7432-296d-4d65-95a1-dd81389d01d0

dcalhoun (commented):
Thank you for putting this together. It tested relatively well for me. I left a few comments for us to consider. Curious of your thoughts.

src/components/editor/use-host-bridge.js
kean: > How many are there that we may need to reimplement and maintain?

This is probably the main risk. The [inserter](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-editor/src/components/inserter), if you remove the media and tests, is ~2500 lines of JS. Most of it is related to UI and search – not business logic. I'll spend some time reading through it tomorrow to see what else we might be missing. I'm concerned there are some unknowns, but, in principle, it seems to be just isolated enough from the rest of the editor, and we could always expose the APIs that we need.

I'm happy to consider alternatives, but so far, it still seems like the one thing worth doing to push the envelope on mobile. If we are not providing first-class integration for media, we are not doing enough, and this seems like the only technical solution.

> For example, when inside a List (technically focused on a List Item), the inserter should display List Item as a top item.

I'm tracking it in the existing ticket [CMM-866: Add support for dynamic blocks – depending on selection](https://linear.app/a8c/issue/CMM-866/add-support-for-dynamic-blocks-depending-on-selection).

> Additionally, selecting other block types that cannot be placed in a List block results in a new block inserted after the currently selected List.

The web does the same. I've also tested it with `core/table` previously and thought it was a bug, but it was the standard Gutenberg behavior.

src/components/editor-toolbar/index.jsx
kean: > you cannot cancel the interaction by dragging your finger away from the button.

I noticed that too yesterday. I'll try different option. ideally, it should behave like native buttons and cancel action.

> or Claude mentioned we could potentially use onClick and pair it with onMouseDown invoking preventDefault() to prevent focus changes.

I tried that yesterday, but was not satisfied with the result for some reason – I'll give it another go.

src/components/editor/use-host-bridge.js
dcalhoun: > > Additionally, selecting other block types that cannot be placed in a List block results in a new block inserted after the currently selected List.
> 
> The web does the same. I've also tested it with `core/table` previously and thought it was a bug, but it was the standard Gutenberg behavior.

Sorry, I meant to convey that standard functionality does _not_ work with the native inserter. The video in my previous comment shows the working functionality with the web inserter.

kean: Oh, I got what you mean now. It does work with `core/table` and the native inserter – the selected incompatible block is appended below the table. It does **not** work with `core/list` – nothing happens.  I'll look into it in the scope of this PR without creating a ticket.

kean:
I'm putting it back into draft. I've research the existing inserters and they all use the reusable `useInsertionPoint` and `useBlockTypesState` [methods](https://github.com/WordPress/gutenberg/blob/efb328b100eed811d0417a662680c0d938c14247/packages/block-editor/src/components/inserter/quick-inserter.js#L26) to determine what blocks to show and how to insert them. We should tap into it. There is no point in merging the intermediate solution.

I will also look into a way to better architect the interaction between web and native. On a high level, I want to create all the requires context by the inserter, including the callbacks, and pass it a a separate JS object to the native side that it could invoke (as opposed to defining these methods on `window.editor`).

cc @dcalhoun – let me know if it sounds like a plan.

dcalhoun:
@kean overall, that makes sense to me. 

> On a high level, I want to create all the requires context by the inserter, including the callbacks, and pass it a a separate JS object to the native side that it could invoke (as opposed to defining these methods on `window.editor`).

Avoiding `window.editor` is attractive. I do not follow what the replacement would resemble. Could you provide pseudo code to help me understand please?

kean:
I made a significant progress by using `useInsertionPoint` and `useBlockTypesState`. With these hooks, the native side doesn't need to know and implement very little.

### Scenarios

Top-Level Insertion
- Insert a block in empty `core/paragraph` – the paragraph gets replaced with the selected block ✅
- Insert a block in an arbitrary “empty” position in the editor – gets inserted at the selected position ✅
- Insert a block when the title is selected - shows an error “Block X can’t be inserted” ⛔️

Composite Blocks
- Insert a supported block (`core/list-item`) in `core/list` – the list item is added ✅
- Insert an unsupported block (`core/image`) in `core/list` – gets inserted after the block ✅
- Insert a supported block (`core/list`) in `core/quote` – gets inserted in the block ✅

Contextual Blocks
- Insert a block that supports a subset of child blocks (`core/list`) and open the inserter – blocks that support this parent (`core/list-item`) are displayed in a dedicated section at the top ✅
- Insert a block that supports a subset of child blocks (`core/list`) in a composite block (`core/quote`) and open the inserter – blocks that support this parent (`core/list-item`) are displayed in a dedicated section at the top ✅

### Known Issues
- [CMM-892 ‒ Add support for blocks with pre-defined parameters](https://linear.app/a8c/issue/CMM-892/add-support-for-blocks-with-pre-defined-parameters)
- [CMM-893 ‒ The editor shown an error when you try to insert a block and the title is selected](https://linear.app/a8c/issue/CMM-893/the-editor-shown-an-error-when-you-try-to-insert-a-block-and-the-title)

### Recording

https://github.com/user-attachments/assets/8ea78011-09ea-498a-8e1e-0e946caaa8dd

### Current Limitations

The main challenges I encountered:

- Both `useInsertionPoint` and `useBlockTypesState` are private functions that are not exposed even via `@wordpress/private-apis`. It looks like we'll need to update the core to mitigate it.
- Both are hooks, so I had to create an empty React component to use them. I'm not sure if there are any performance implications to it. It seems to re-compute often.

I'll might need some help with it. I'm going to wrap up for today.

> Could you provide pseudo code to help me understand please?

I decided to keep to simple. If there is more than one scenario where we need a generalized solution, then maybe it would be a good time to consider it.

src/components/editor-toolbar/index.jsx
kean: I switched to `onClick` as React seems to handle well.

kean:
I made a couple of small changes:

- Fix [CMM-893 ‒ The editor shows an error when you try to insert a block and the title is selected](https://linear.app/a8c/issue/CMM-893/the-editor-shown-an-error-when-you-try-to-insert-a-block-and-the-title) (adding `undefined` as `clientId` when no blocks are selected)
- Switch on `onClick`
- Rename `BlockInserterBridge` to `NativeBlockInserterButton`. It's no longer an empty component. It renders the actual button.

I think it's now in a mergeable state with a caveat that we need to figure out the best way to use private APIs.

I also logged an additional issue: [CMM-903: Performance: Inserter recomputes the list of blocks on every change in the editor](https://linear.app/a8c/issue/CMM-903/performance-inserter-recomputes-the-list-of-blocks-on-every-change-in).

src/components/native-block-inserter-button/index.jsx
```diff
@@ -0,0 +1,130 @@
...
+// - privateApis: These specific hooks are not exported there
+// - Copying hooks: Too complex, uses multiple internal unlock() calls
+// - Using PrivateQuickInserter component: Not suitable for headless bridge
+//
+// This approach is acceptable because:
+// - We need the exact same insertion logic as the WordPress editor
+// - The hooks are stable in practice (used by core components)
+// - We're building a WordPress editor integration, not a general library
+import useInsertionPoint from '@wordpress/block-editor/build-module/components/inserter/hooks/use-insertion-point';
+import useBlockTypesState from '@wordpress/block-editor/build-module/components/inserter/hooks/use-block-types-state';
```
dcalhoun, line 29: These should likely be grouped in the `WordPress dependencies` section above. Don't you think?

src/components/native-block-inserter-button/index.jsx
```diff
@@ -0,0 +1,130 @@
...
+import { Button } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { plus } from '@wordpress/icons';
+
+/**
+ * Internal dependencies
+ */
+// NOTE: These hooks are internal WordPress APIs not available via public exports
+// or privateApis. We import from build-module as the only way to access the
+// block insertion logic without reimplementing it ourselves.
```
dcalhoun, line 12: I suggest we document this usage in #106. Ideally, we propose these become public APIs and update this code.

src/components/native-block-inserter-button/index.jsx
```diff
@@ -0,0 +1,130 @@
...
+	} );
+
+	const [ inserterItems, , , onSelectItem ] = useBlockTypesState(
+		destinationRootClientId,
+		onInsertBlocks,
+		false // isQuick
+	);
+
+	// Serialize blocks for native consumption
+	const blocks = serializeBlocksForNative( inserterItems );
```
dcalhoun: I suggest we memoize this via [`useMemo`](https://react.dev/reference/react/useMemo) to avoid unnecessary renders—unless `inserterItems` is a very deep object that makes equality comparisons difficult. 

If we are only ever using `serializeBlocksForNative` within React components, we might consider making it a React hook and relocating it to `src/components/native-block-inserter-button/use-serialize-blocks-for-native.jsx`. 

```jsx
export function useSerializeBlocksForNative(inserterItems) {
  return useMemo(() => {
    // existing serializeBlocksForNative logic
  }, [inserterItems]);
}
```

src/components/native-block-inserter-button/index.jsx
```diff
@@ -0,0 +1,130 @@
...
+	}, [ blocks, destinationBlockName, inserterItems, onSelectItem ] );
+
+	return (
+		<Button
+			title={ __( 'Add block' ) }
+			icon={ plus }
+			onClick={ ( e ) => {
+				e.preventDefault();
+				showBlockInserter();
+			} }
```
dcalhoun: From my testing, this does not suffice for retaining focus on the editor canvas. It appears we must prevent the default event within `onMouseDown`. 

```diff
diff --git a/src/components/native-block-inserter-button/index.jsx b/src/components/native-block-inserter-button/index.jsx
index f521204..1c64eee 100644
--- a/src/components/native-block-inserter-button/index.jsx
+++ b/src/components/native-block-inserter-button/index.jsx
@@ -120,10 +120,12 @@ export default function NativeBlockInserterButton() {
 		<Button
 			title={ __( 'Add block' ) }
 			icon={ plus }
-			onClick={ ( e ) => {
-				e.preventDefault();
+			onClick={ () => {
 				showBlockInserter();
 			} }
+			onMouseDown={ ( e ) => {
+				e.preventDefault();
+			} }
 			className="gutenberg-kit-add-block-button"
 		/>
 	);

```

dcalhoun (approved):
Works well from my testing. Code looks good too. I left a few suggestions for your consideration.

src/components/native-block-inserter-button/index.jsx
kean: I pushed the suggested change. Now, when you close the inserter without selection, the focus is restored. 
I feel like I'm misunderstand how the even system works. We'll need a way to fix it across all these buttons in the future.

src/components/native-block-inserter-button/index.jsx
kean: I approached it slightly differently in https://github.com/wordpress-mobile/GutenbergKit/pull/203/files#r2470899555 to eliminate re-computing if when it's not needed. Let's merge what we have, and I'll be happy to refactor and optimize it. I'll follow your advice.

src/components/native-block-inserter-button/index.jsx
kean: Updated.

src/components/native-block-inserter-button/index.jsx
dcalhoun: Sounds good.

src/components/native-block-inserter-button/index.jsx
dcalhoun: I believe the issue with `onClick` is that the relevant event (focus change) has already occurred by the time `onClick` is invoked. So, calling `event.preventDefault()` does not have the desired outcome. Using events firing earlier—like `onMouseDown`—provide an opportunity to prevent the focus change. 

Modifying buttons rendered by Gutenberg will be challenging.

PR #202, opened 2025-10-23
Refactor Sample App and Logging

jkmassel:
## What?
While doing performance improvements, I ran into a few issues:

1. It was hard to debug issues early in the lifecycle before I could attach the Web Inspector. 
2. SwiftUI kept complaining about the demo app's view hierarchy being too complicated to evaluate in reasonable time.

This PR solves both to avoid rebasing later.

## Why?
See above

## How?
I split the view hierarchy into a couple of files and removed a bunch of code.

I added a logging bridge message and added it to the editor configuration.

## Testing Instructions
If you build and run the app there should be no difference. If you update the remote configuration builder in the demo app to use `.debug` as the log level you'll see some messages from existing usage.

n/a


ios/Demo-iOS/Sources/Services/ConfigurationStorage.swift
```diff
@@ -1,23 +1,30 @@
 import Foundation
 
 /// Manages persistence of remote editor configurations
-class ConfigurationStorage {
+class ConfigurationStorage: ObservableObject {
```
jkmassel, line 4: This class is now just injected using EnvironmentObject which removes some code.

ios/Demo-iOS/Sources/Views/AppRootView.swift
```diff
@@ -0,0 +1,99 @@
...
+                let client = WordPressAPI(
+                    urlSession: .shared,
+                    apiRootUrl: try ParsedUrl.parse(input: config.siteApiRoot),
+                    authentication: .authorizationHeader(token: config.authHeader)
+                )
+
+                let apiRoot = try await client.apiRoot.get().data
+
+                let canUsePlugins = apiRoot.hasRoute(route: "/wpcom/v2/editor-assets")
+                let canUseEditorStyles = apiRoot.hasRoute(route: "/wp-block-editor/v1/settings")
```
jkmassel, line 73: We can now hit the actual site and figure out if it supports plugins or editor styles, which makes the demo app more useful for testing with real sites.

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -302,6 +302,12 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
             case .onModalDialogClosed:
                 let body = try message.decode(EditorJSMessage.ModalDialogBody.self)
                 delegate?.editor(self, didCloseModalDialog: body.dialogType)
+            case .log:
+                let log = try message.decode(EditorJSMessage.LogMessage.self)
+
+                if log.level >= controller.configuration.logLevel {
+                    print(log.message)
```
kean: (suggestion/future)
Let's send it to the delegate (`GutenbergEditorControllerDelegate`) and let the app use its logging infrastructure so we could log critical errors and performance metics in production so if someones reaches out with a performance issue, we had an idea of what's going on.

ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
```diff
@@ -53,6 +55,7 @@ public struct EditorConfiguration: Sendable {
         locale: String,
         isNativeInserterEnabled: Bool,
         editorAssetsEndpoint: URL? = nil,
+        logLevel: LogLevel = .warn
```
kean: I would disable logging by default so there is no impact on performance and enable it only in the demo and in the app when the app needs it.

ios/Demo-iOS/Sources/Views/AppRootView.swift
```diff
@@ -0,0 +1,99 @@
...
+import GutenbergKit
+import AuthenticationServices
+import WordPressAPI
+
+struct AppRootView: View {
+    @State private var selectedConfiguration: ConfigurationItem?
+    @State private var configurations: [ConfigurationItem] = [.bundledEditor]
+    @State private var siteUrlInput = ""
+    @State private var authenticationManager = AuthenticationManager()
+    @State private var configurationStorage = ConfigurationStorage()
```
kean: A remaining separate `ConfigurationStorage` instance that is not using `@EnvironmentObject`.

ios/Demo-iOS/Sources/Views/AppRootView.swift
```diff
@@ -0,0 +1,99 @@
...
+
+    var body: some View {
+        EditorList(isNativeInserterEnabled: $isNativeInserterEnabled, selectedConfiguration: $selectedConfiguration)
+        .overlay {
+            if let error {
+                Text(error.localizedDescription)
+            }
+        }
+        .fullScreenCover(item: $selectedConfiguration) { config in
+            editorOverlay
```
kean: (nit) "overlay" usually refers to a view [overlay(alignment:content:)](https://developer.apple.com/documentation/swiftui/view/overlay(alignment:content:)) and not a screen

ios/Demo-iOS/Sources/Views/AppRootView.swift
```diff
@@ -0,0 +1,99 @@
...
+                    .setShouldUseThemeStyles(canUseEditorStyles)
+                    .setShouldUsePlugins(canUsePlugins)
+                    .setSiteUrl(config.siteUrl)
+                    .setSiteApiRoot(config.siteApiRoot)
+                    .setAuthHeader(config.authHeader)
+                    .setNativeInserterEnabled(isNativeInserterEnabled)
+                    .setLogLevel(.debug)
+                    .build()
+
+                await MainActor.run {
```
kean: I wonder why Claude keeps inserting these. The `Task` that this code runs is already isolated to the same actor that the containing type – `@MainActor`. There should generally never be a need to use `await MainActor.run` because any tasks invoked from the UI will already be on main.

ios/Demo-iOS/Sources/Views/AppRootView.swift
```diff
@@ -0,0 +1,99 @@
...
+
+    @State private var editorRemoteConfiguration: EditorConfiguration? = nil
+    @State private var error: Error? = nil
+
+    @AppStorage("isNativeInserterEnabled") private var isNativeInserterEnabled = false
+
+    var body: some View {
+        EditorList(isNativeInserterEnabled: $isNativeInserterEnabled, selectedConfiguration: $selectedConfiguration)
+        .overlay {
+            if let error {
```
kean: Let's put this in a sheet so you can dismiss it. I haven't tested this flow, but it looks like it will be shown in the middle of the screen over other content. You'll need to wrap it in an `Identifiable` type.

ios/Demo-iOS/Sources/Views/AppRootView.swift
```diff
@@ -0,0 +1,99 @@
...
+            } else {
+                ProgressView("Preparing Editor")
+            }
+        }
+    }
+
+    private func loadRemoteEditor(for config: RemoteEditorConfiguration) {
+        Task {
+            do {
+                let client = WordPressAPI(
```
kean, line 64: (nit/future) 
in the app, the equivalent code runs inside the editor view itself. I would suggest to ideally move it to `EditorView` and maybe even consider adding `wprs` as a dependency to `GBK` and implementing it, together with the plugin recommendations, in `GBK` itself.

ios/Demo-iOS/Gutenberg.xcodeproj/xcshareddata/xcschemes/Gutenberg.xcscheme
```diff
@@ -58,7 +58,7 @@
          </EnvironmentVariable>
          <EnvironmentVariable
             key = "GUTENBERG_EDITOR_REMOTE_URL"
-            value = "http://localhost:5174/remote.html"
+            value = "http://localhost:5173/remote.html"
```
kean: I ran into the same issue when starting with the remote editor. I think it's port `5174` because you typically run `make dev-server` for the bundled editor first and then also run `make dev-server-remote` for the remote editor that then has to use port `5174` as `5173` is taken. I can't elaborate more.

ios/Demo-iOS/Gutenberg.xcodeproj/project.pbxproj
```diff
@@ -325,7 +333,6 @@
 				ENABLE_PREVIEWS = YES;
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_KEY_CFBundleURLTypes = "$(INFOPLIST_KEY_CFBundleURLTypes)";
-				"INFOPLIST_KEY_CFBundleURLTypes[0]" = "{ CFBundleURLSchemes = (gutenbergkit); }";
```
kean, line 328: Are you sure we not using these schemes – just checking?

kean (commented):
Nice!

> It was hard to debug issues early in the lifecycle before I could attach the Web Inspector.

In addition to logging, which is must have, it would be fantastic to also add an option to show a "Continue" button in `EditorViewController` to allow you to debug the start up sequence using Safari Developer Tools:

- `EditorViewController` creates `WKWebView`
- You connect it to it in Safari and tap "Continue"
- Editor gets loaded
- You can see everything that's being loaded and use the Safari performance tools – profit

ios/Demo-iOS/Gutenberg.xcodeproj/xcshareddata/xcschemes/Gutenberg.xcscheme
dcalhoun: Yes, the reason I set the remote editor to default as `5174` is because I generally run both when using a development server. 

I suggest keeping `5174` for now, but I'm open to improving the DX to mitigate confusion. 

Additional context:

* I generally run the bundled and remote editor servers at the same time, using both `5173` and `5174` ports.
* We could consider changing the `make dev-server` command to start both servers (downside is lack of ability to pass Vite-specific arguments, if needed). 
* The [remote editor redirects](https://github.com/wordpress-mobile/GutenbergKit/blob/d244240d95ca40f8eb81a6f7f1057b009bb578a2/src/utils/remote-editor.jsx#L89-L92) to the bundled editor when it fails to load.
* The redirect doesn't actually work if a development server is in use due to the separate Vite config/server ports for the bundled and remote editors. 
* One can disable the redirect via the [`?dev_mode`](https://github.com/wordpress-mobile/GutenbergKit/blob/d244240d95ca40f8eb81a6f7f1057b009bb578a2/docs/architecture.md#development-mode) parameter, but we could consider disabling the redirect anytime a development server is in use (the redirect is not helpful while developing).

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
dcalhoun: Does it make sense to instead rely upon the [existing conditional logging](https://github.com/wordpress-mobile/GutenbergKit/blob/9ecd1477c2da940effcceefdbbd11519d633013b/src/utils/logger.js#L57-L65) in JavaScript? 

The new native editor configuration could [set the level](https://github.com/wordpress-mobile/GutenbergKit/blob/9ecd1477c2da940effcceefdbbd11519d633013b/src/utils/logger.js#L42-L55) in JavaScript. Doing so would avoid unnecessary (currently conditionally ignored here on line 308) bridge events. It would also reduce the conditionals to a single place: the JavaScript logger implementation. 

I don't have strong feelings on this. Sharing it for consideration.

dcalhoun (commented):
Changes look to good me. Thank you! 

I left a couple of suggestions to consider.

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
jkmassel: I've addressed this in 8bc00400621debcc8aeb6888d8d8d1f831fe98fe – the logging config is injected into the `GBKit` global and applied to the editor at that point.

ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
jkmassel: There's no real concept of "disabling" logging in the library at the moment, and IMHO errors always be emitted. I've set the default to `error` for now.

If it becomes a problem we can change it later but that'd be a bigger refactor.

ios/Demo-iOS/Sources/Views/AppRootView.swift
jkmassel: Put it as an error alert for now

ios/Demo-iOS/Sources/Views/AppRootView.swift
jkmassel: Yeah, I think the way this ends up going is that there's `GBKit` package, then there's a `GBKit+WP` package that interops with `wprs`. 

The GBKit stuff can land in Gutenberg Core and then this repo is just `GBKit+WP`

ios/Demo-iOS/Gutenberg.xcodeproj/xcshareddata/xcschemes/Gutenberg.xcscheme
jkmassel: Reverted it back for now

ios/Demo-iOS/Gutenberg.xcodeproj/project.pbxproj
jkmassel: Not required because we use `ASWebAuthentication`, which temporarily accepts the URL scheme we register with it while authenticating, then stops once authentication is complete.

kean (approved):

PR #203, opened 2025-10-27
Native Inserter: Add initial Photos integration

kean:
## What?
- Closes [CMM-860: Add trailing bar button to add media from Photos](https://linear.app/a8c/issue/CMM-860/add-trailing-bar-button-to-add-media-from-photos)
- Closes [CMM-903: Performance: Inserter recomputes the list of blocks on every change in the editor](https://linear.app/a8c/issue/CMM-903/performance-inserter-recomputes-the-list-of-blocks-on-every-change-in)

## How?

I used [onFilesDrop](https://github.com/WordPress/gutenberg/blob/439823dfd75b299964ffe2af53f8438a440f676c/packages/block-editor/src/components/use-on-block-drop/index.js#L164) from [use-on-block-drop](https://github.com/WordPress/gutenberg/tree/439823dfd75b299964ffe2af53f8438a440f676c/packages/block-editor/src/components/use-on-block-drop) component as a reference. It uses "transforms" to create blocks and then uses the existing insertion logic introduced in https://github.com/wordpress-mobile/GutenbergKit/pull/199 to insert the blocks created by transforms.

This implementation should make sure that we support any media types, any transforms including third-party block ones, and any insertion points including nested blocks.

## Testing Instructions

- **Verify** that if you select one image, a `core/image` block is added and the item is uploaded to the remote
- **Verify** that if you select multiple images, a gallery is created
- **Verify** that if you select a video, a video is uploaded (note: should use videopress on Dotcom sites; untested)
- **Verify** that media blocks are inserted at the insertion points, including insertion points in nested blocks

**Note**: I focused on making the implementation simple and sound and plan to test it and fix any outstanding issues in future iterations.

## Screenshots or screencast 

https://github.com/user-attachments/assets/a62d9deb-9d68-4c90-850a-0d6636798eac

**Note**: the first frame in the random video I uploaded is black, which is why there is no preview

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
@@ -55,6 +55,7 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         for scheme in CachedAssetSchemeHandler.supportedURLSchemes {
             config.setURLSchemeHandler(schemeHandler, forURLScheme: scheme)
         }
+        config.setURLSchemeHandler(MediaFileSchemeHandler(), forURLScheme: MediaFileSchemeHandler.scheme)
 
         self.webView = GBWebView(frame: .zero, configuration: config)
         self.webView.scrollView.keyboardDismissMode = .interactive
@@ -263,8 +264,8 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
                 onBlockSelected: { [weak self] block in
                     self?.insertBlockFromInserter(block.id)
                 },
-                onMediaSelected: {
-                    print("insert media:", $0)
+                onMediaSelected: { [weak self] selection in
+                    self?.insertMediaFromInserter(selection)
                 }
             )
         })
@@ -276,6 +277,17 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
 
     private func insertBlockFromInserter(_ blockID: String) {
         evaluate("window.blockInserter.insertBlock('\(blockID)')")
+    }
+
+    private func insertMediaFromInserter(_ selection: [MediaInfo]) {
+        guard !selection.isEmpty else { return }
+
+        guard let data = try? JSONEncoder().encode(selection),
+              let string = String(data: data, encoding: .utf8) else {
+            debugPrint("Failed to serialize media array to JSON")
+            return
+        }
+        evaluate("window.blockInserter.insertMedia(\(string))")
     }
 
     private func openMediaLibrary(_ config: OpenMediaLibraryAction) {
--- a/ios/Sources/GutenbergKit/Sources/Media/MediaFileManager.swift
+++ b/ios/Sources/GutenbergKit/Sources/Media/MediaFileManager.swift
@@ -1,0 +1,83 @@
+import Foundation
+import PhotosUI
+import SwiftUI
+import UniformTypeIdentifiers
+
+/// Manages media files for the editor, handling imports, storage, and cleanup.
+///
+/// Files are stored in the Library/GutenbergKit/Uploads directory and served via
+/// a custom `gbk-media-file://` URL scheme. Old files are automatically cleaned up.
+actor MediaFileManager {
+    /// Shared instance for app-wide media management
+    static let shared = MediaFileManager()
+
+    private let fileManager = FileManager.default
+    private let rootURL: URL
+    private let uploadsDirectory: URL
+
+    init(rootURL: URL = URL.libraryDirectory.appendingPathComponent("GutenbergKit")) {
+        self.rootURL = rootURL
+        self.uploadsDirectory = self.rootURL.appendingPathComponent("Uploads")
+        Task {
+            await cleanupOldFiles()
+        }
+    }
+
+    /// Imports a photo picker item and saves it to the uploads directory.
+    ///
+    /// - Returns: MediaInfo with a `gbk-media-file://` URL and detected media type
+    func `import`(_ item: PhotosPickerItem) async throws -> MediaInfo {
+        guard let data = try await item.loadTransferable(type: Data.self) else {
+            throw URLError(.unknown)
+        }
+        let contentType = item.supportedContentTypes.first
+        let fileExtension = contentType?.preferredFilenameExtension ?? "jpeg"
+
+        let fileURL = try await writeData(data, withExtension: fileExtension)
+        return MediaInfo(url: fileURL.absoluteString, type: contentType?.preferredMIMEType)
+    }
+
+    /// Saves media data to the uploads directory and returns a URL with a
+    /// custom scheme.
+    func writeData(_ data: Data, withExtension ext: String) async throws -> URL {
+        let fileName = "\(UUID().uuidString).\(ext)"
+        let destinationURL = uploadsDirectory.appendingPathComponent(fileName)
+
+        try fileManager.createDirectory(at: uploadsDirectory, withIntermediateDirectories: true)
+        try data.write(to: destinationURL)
+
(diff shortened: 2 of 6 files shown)


ios/Sources/GutenbergKit/Sources/Media/MediaFileManager.swift
```diff
@@ -0,0 +1,160 @@
...
+                urlSchemeTask.didReceive(data)
+                urlSchemeTask.didFinish()
+            } catch {
+                urlSchemeTask.didFailWithError(error)
+            }
+        }
+    }
+
+    func webView(_ webView: WKWebView, stop urlSchemeTask: WKURLSchemeTask) {
+        // Nothing to do here for simple file serving
```
jkmassel: You probably want to store the task as a member of this class and cancel it from this method?

ios/Sources/GutenbergKit/Sources/Media/MediaFileManager.swift
```diff
@@ -0,0 +1,160 @@
...
+        if let rootURL {
+            self.rootURL = rootURL
+        } else {
+            self.rootURL = URL.libraryDirectory.appendingPathComponent("GutenbergKit")
+        }
+        self.uploadsDirectory = self.rootURL.appendingPathComponent("Uploads")
+
+        // Create directories synchronously during init
+        try? fileManager.createDirectory(at: self.rootURL, withIntermediateDirectories: true)
+        try? fileManager.createDirectory(at: self.uploadsDirectory, withIntermediateDirectories: true)
```
jkmassel: If these silently fail `saveMediaData` won't work right.

Probably best to just run these on every call to `saveMediaData` if needed – makes the error handling way more straightforward.

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -55,6 +55,7 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         for scheme in CachedAssetSchemeHandler.supportedURLSchemes {
             config.setURLSchemeHandler(schemeHandler, forURLScheme: scheme)
         }
+        config.setURLSchemeHandler(MediaFileSchemeHandler(), forURLScheme: MediaFileSchemeHandler.scheme)
```
kean, line 58: This is the same approach we use for the assets. It's currently the only way to let `WKWebView` access the files on disk, but I hope we can simplify it in the future by extracting the bundled editor on disk and pointing `WKWebView` to that directly and not the readonly bundle.

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -278,6 +279,17 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         evaluate("window.blockInserter.insertBlock('\(blockID)')")
     }
 
+    private func insertMediaFromInserter(_ selection: [MediaInfo]) {
+        guard !selection.isEmpty else { return }
+
+        guard let data = try? JSONEncoder().encode(selection),
+              let string = String(data: data, encoding: .utf8) else {
+            debugPrint("Failed to serialize media array to JSON")
```
kean, line 287: This should never happen.

src/components/native-block-inserter-button/index.jsx
```diff
@@ -80,46 +82,122 @@ export default function NativeBlockInserterButton() {
...
+	};
 
 	return (
 		<Button
 			title={ __( 'Add block' ) }
 			icon={ plus }
 			onClick={ ( e ) => {
 				e.preventDefault();
+
+				const blocks = serializeBlocksForNative( inserterItems );
```
kean: This is the fix for [CMM-903: Performance: Inserter recomputes the list of blocks on every change in the editor](https://linear.app/a8c/issue/CMM-903/performance-inserter-recomputes-the-list-of-blocks-on-every-change-in). Instead of `useEffect`, it now serializes blocks only when it's needed.

Ideally, I would also ensure `useInsertionPoint` isn't called so frequently, but it could be done later.

src/components/native-block-inserter-button/index.jsx
```diff
@@ -80,46 +82,122 @@ export default function NativeBlockInserterButton() {
...
+			const transformation = findTransform(
+				getBlockTransforms( 'from' ),
+				( transform ) =>
+					transform.type === 'files' &&
+					transform.isMatch &&
+					transform.isMatch( validFiles )
+			);
+
+			if ( ! transformation ) {
+				debug( 'insertMedia: No matching transform found', {
```
kean, line 166: I focused only on the happy path and opened [CMM-910: Add error handling for media insertion](https://linear.app/a8c/issue/CMM-910/add-error-handling-for-media-insertion)

ios/Sources/GutenbergKit/Sources/Media/MediaFileManager.swift
kean: Good idea; updated.

ios/Sources/GutenbergKit/Sources/Media/MediaFileManager.swift
kean: It's a nearly-instant operation. It will never be cancelled in flight in practice.

src/components/native-block-inserter-button/index.jsx
```diff
@@ -80,46 +82,123 @@ export default function NativeBlockInserterButton() {
...
+
+			// Find matching transform using WordPress's transform system
+			// This is the same logic as onFilesDrop in use-on-block-drop
+			const transformation = findTransform(
+				getBlockTransforms( 'from' ),
+				( transform ) =>
+					transform.type === 'files' &&
+					transform.isMatch &&
+					transform.isMatch( validFiles )
+			);
```
dcalhoun, line 163: Does this intentionally exclude the `canInsertBlockType( transform.blockName, targetRootClientId )` [logic](https://github.com/WordPress/gutenberg/blob/439823dfd75b299964ffe2af53f8438a440f676c/packages/block-editor/src/components/use-on-block-drop/index.js#L180)?

src/components/native-block-inserter-button/index.jsx
dcalhoun: Sounds good. We can revisit other [memoization ideas](https://github.com/wordpress-mobile/GutenbergKit/pull/199#discussion_r2474322561) later. Maybe we can further reduce `useInsertionPoint` invocations.

dcalhoun (approved):
Selecting media feels very satisfying. Nicely done.

src/components/native-block-inserter-button/index.jsx
kean: I made it consistent with the drag-and-drop and tested the basic scenarios, including nested blocks like `core/quote`. It worked as expected.

</hr>

The linter won't let me call a method with more than one argument as a single-line.

```
transform.type === 'files' &&
canInsertBlockType( transform.blockName, destinationRootClientId ) &&
transform.isMatch( validFiles )
```

It changes it to the following, which is arguably less readable:

```
transform.type === 'files' &&
canInsertBlockType(
    transform.blockName,
    destinationRootClientId
) &&
transform.isMatch( validFiles )
```

It's it a bit too much? wdyt? Can we change it?

dcalhoun: No, we should not modify the linter and code formatter configuration. Doing so would cause problems when we go to place GBK in the Gutenberg repo. 

If we want to improve readability, we'd likely need to hoist the predicate function to reduce nesting. 

```js
function predicate(transform) {
  // [...]
}
const transformation = findTransform(
  getBlockTransforms( 'from' ),
  predicate
}
```

PR #209, opened 2025-10-31
feat: Bundled editor supports plugins

dcalhoun:
Related:
- https://github.com/Automattic/jetpack/pull/45714
- https://github.com/Automattic/jetpack/pull/45715
- https://github.com/wordpress-mobile/WordPress-iOS/pull/24965
- https://github.com/wordpress-mobile/WordPress-Android/pull/22331

## What?

Load plugin assets within the default, bundled editor. 

## Why?

This results in several improvements: 

- Reduce the number of assets retrieved from the site, improving performance through reduced network activity. 
- Enable important patches applied to external modules. 
  - Fix CMM-880 - fix "Browse All" within quick inserter 
  - Fix CMM-895 - fix broken text entry after swiping outside of block canvas 
  - Fix CMM-902 - enable populating the _Patterns_ tab within the block inserter 
- Enable exploratory private API usage to allow planning public APIs. 
- Enable private module access necessary for the native iOS inserter feature. See: p1761568270564209-slack-C08UN8YN6KW.
- Fix CMM-583 - relying upon local `@wordpress` modules avoids WordPress core block limitations
- Fix CMM-897 - no longer loading preferences file from remote site 
- Prepare for project simplification by negating the need for a separate "remote" editor variant (`src/remote.html`); **related code will be removed at a later time** in a separate PR. **Update:** see https://github.com/wordpress-mobile/GutenbergKit/pull/210. 

## How?

* Bundle all necessary `@wordpress` module and expose them via `window.wp` globals, much like the WP Admin environment. 
* Leverage https://github.com/Automattic/jetpack/pull/45715 to fetch only plugin assets. 

## Testing Instructions

> [!tip]
> Use the prototype builds for testing:
> - https://github.com/wordpress-mobile/WordPress-iOS/pull/24965#issuecomment-3474200229
> - https://github.com/wordpress-mobile/WordPress-Android/pull/22331#issuecomment-3474301188

> [!important]
> To test the end result of these changes, https://github.com/Automattic/jetpack/pull/45715 must be applied to your site. Otherwise, plugins will not load.  

- With https://github.com/Automattic/jetpack/pull/45715 applied, verify the editor loads with plugins. 
- Without https://github.com/Automattic/jetpack/pull/45715 applies, verify the editor loads (plugins will not be available). 
- Smoke test various editor functionality—text editing, media uploads, etc. 
- Verify the bugs referenced in the "Why?" section are, in fact, fixed. 

N/A, no navigation changes. 

## Screenshots or screencast 

https://github.com/user-attachments/assets/d61a379d-8b2b-48f5-819c-0020cf787c75


ios/Sources/GutenbergKit/Sources/Cache/EditorAssetsLibrary.swift
```diff
@@ -22,10 +22,14 @@ public actor EditorAssetsLibrary {
     /// reused on future calls.
     func loadManifestContent() async throws -> Data {
         let endpoint: URL
+        // The GutenbergKit bundle includes the required `@wordpress` modules
+        let excludeParam = URLQueryItem(name: "exclude", value: "core,gutenberg")
```
dcalhoun, line 26: This leverages the query parameter introduced in https://github.com/Automattic/jetpack/pull/45715. We exclude "core" and Gutenberg so that only third-party plugin assets are returned.

src/components/editor-load-error/index.jsx
```diff
@@ -12,51 +11,46 @@ import './style.scss';
 /**
  * Displays an error notice when the editor fails to load.
  *
+ * Simple error message avoids `@wordpress` components to avoid
+ * complex dependency management during editor initialization.
```
dcalhoun, line 15: This simple error message is used within the editor entry file if an error occurs during the sequence to load and initialize foundational `@wordpress` modules. Libraries like `react` or `@wordpress/components` may not be available depending on the error, so we use string templates instead. 

Long term, we should broadcast the editor load failure to the native host app and allow it to control how an error is presented to the user.

src/utils/bundled-editor.jsx
```diff
@@ -30,33 +24,57 @@ export function initializeBundledEditor() {
 	// - https://github.com/vitejs/vite/issues/13952
 	// - https://github.com/vitejs/vite/issues/5189#issuecomment-2175410148
 	return awaitGBKitGlobal()
-		.then( initializeApiAndLocale )
-		.then( importEditor )
+		.then( configureLocale )
+		.then( loadRemainingGlobals )
+		.then( initializeApiFetchWrapper )
 		.then( initializeEditor )
+		.then( loadPluginsIfEnabled )
```
dcalhoun, line 32: This sequence is fairly similar to the one it's replacing and the one found in the to-be-deprecated [`src/utils/remote-editor.jsx`](https://github.com/wordpress-mobile/GutenbergKit/blob/3b1af54e56ef0278e993703097a607433840cb4c/src/utils/remote-editor.jsx#L30-L38). The important piece to understand is that a specific loading sequence is required to ensure dependencies are in place before the logic attempts to reference them.

1. Load i18n modules (within `index.html`). 
2. Configure i18n. 
3. Load remaining `@wordpress` modules (depend upon i18n configuration). 
4. Load plugins (depend upon `@wordpress` modules).

src/utils/editor-loader.js
```diff
@@ -107,9 +107,13 @@ async function loadAssets(
...
-	const excludedScriptIDs = new RegExp(
-		disallowedPackages.map( ( script ) => `wp-${ script }-js` ).join( '|' )
-	);
+	const excludedScriptIDs = disallowedPackages.length
+		? new RegExp(
+				disallowedPackages
+					.map( ( script ) => `wp-${ script }-js` )
+					.join( '|' )
+		  )
+		: null;
```
dcalhoun, line 116: This fixes a hidden bug where an empty `disallowedPackages` value resulted in _disallowing all packages_ due to the empty regex created. Falling back to `null` avoids this.

src/utils/wordpress-globals.js
dcalhoun, line 1: We expose all the relevant `@wordpress` modules as `window.wp` globals to mimic the WP Admin environment. This allows plugins to rely upon them as they do in the WP Admin environment.

src/utils/wordpress-i18n.js
dcalhoun, line 1: As noted elsewhere, we must load and configure i18n before the rest of the `@wordpress` modules load. This is due to top-level localization strings found in the `@wordpress` modules. See https://github.com/wordpress-mobile/gutenberg-mobile/issues/4329.

vite.config.js
dcalhoun, line 1: Much of this is copied from the [`vite.config.remote.js`](https://github.com/wordpress-mobile/GutenbergKit/blob/3b1af54e56ef0278e993703097a607433840cb4c/vite.config.remote.js) file, with a few adjustments noted inline.

vite.config.js
```diff
@@ -21,3 +27,152 @@ export default defineConfig( {
...
+ * @return {Object} The transformed code and map.
+ */
+function wordPressExternals() {
+	return {
+		name: 'wordpress-externals-plugin',
+		transform( code, id ) {
+			// Skip transforming node_modules files
+			if ( id.includes( 'node_modules' ) ) {
+				return null;
+			}
```
dcalhoun, line 49: The previous remote editor variant prevented transforming `node_modules` by externalizing them via the `external` Rollup configuration. Because now bundle `@wordpress` modules, we no longer mark them as external. To avoid transforming `node_modules`, we add new guard logic here.

package.json
```diff
@@ -36,9 +37,15 @@
...
 		"@wordpress/icons": "^10.16.0",
 		"@wordpress/preferences": "^4.16.0",
+		"@wordpress/preferences-persistence": "^2.33.0",
 		"@wordpress/private-apis": "^1.10",
 		"@wordpress/rich-text": "^7.16.0",
+		"@wordpress/style-engine": "^2.34.0",
 		"clsx": "^2.1.1",
+		"lodash": "^4.17.21",
+		"moment": "^2.30.1",
+		"react": "^18.3.1",
```
kean, line 48: Just checking – are these dependencies updates part of the change?
Is `lodash` something that plugins and react might need?

src/components/editor-load-error/index.jsx
kean: > we should broadcast the editor load failure to the native host app and allow it to control how an error is presented to the user

That sounds great. And thanks for addressing it. It already helped me see an error due to the new dependencies.

kean (approved):
Tested with my WoA site – worked great, blocks from Jetpack worked.

package.json
dcalhoun: These dependencies were added as GutenbergKit now explicitly [exposes them as globals](https://github.com/wordpress-mobile/GutenbergKit/blob/50f81257f9e0c4b535b494c2f492153b3b89d4af/src/utils/wordpress-globals.js#L58-L62) for plugins. 

Yes, plugins may leverage the `lodash` global. In my testing, the Jetpack plugin does.

PR #212, opened 2025-11-04
Native Inserter: Add inline compact PhotosPicker

kean:
## What?
- Close https://linear.app/a8c/issue/CMM-859/add-inline-photoskit-row-of-photos
- Configure modal presentation detents for the inserter
- Fix an issue with incorrect invocation of asynchronous `insertMedia` method. It was throwing an exception that was previously suppressed by the inserter covering the editor. I replaced it with a new-ish [callAsyncJavaScript](https://developer.apple.com/documentation/webkit/wkwebview/callasyncjavascript(_:arguments:in:in:completionhandler:)) API.
- Update CI to use Xcode 26 and iOS 26

## How?
- n/a

## Testing Instructions

- **Verify** that you can select and upload media item using the new inline photos picker

Known Issue: [CMM-928: Editor restores focus after returning from the background if the inserter if open](https://linear.app/a8c/issue/CMM-928/editor-restores-focus-after-returning-from-the-background-if-the) (found during testing)

## Screenshots or screencast

Screenshot 2025-11-04 at 11 19 10 AM Screenshot 2025-11-04 at 11 19 52 AM
Screenshot 2025-11-04 at 11 19 13 AM  Screenshot 2025-11-04 at 11 19 49 AM

## Recording

https://github.com/user-attachments/assets/2e241c46-d23b-420a-871f-383b58c85654


ios/Sources/GutenbergKit/Sources/Modifiers/CardModifier.swift
```diff
@@ -1,13 +1,11 @@
 import SwiftUI
 
 struct CardModifier: ViewModifier {
+    // - warning: It was previously using .overlay, but it turned out to be
+    // incompatible with inline PhotosPicker – breaks gestures on a device.
     func body(content: Content) -> some View {
         content
             .background(Color(.systemBackground))
-            .overlay(
```
kean, line 7: I've been hitting the weirdest issue with the inline `PhotosPicker`. Not only it a layout issue on iOS 26 I shared earlier and reported to Apple – still no fix/answer. Apparently, any kind of `.overlay` or `.border` makes in non-tappable on device (but not simulator). It took me a bit to figure it out.

ios/Sources/GutenbergKit/Sources/Views/BlockInserter/BlockInserterView.swift
```diff
@@ -13,8 +13,9 @@ struct BlockInserterView: View {
     @StateObject private var iconCache = BlockIconCache()
 
     @State private var selectedMediaItems: [PhotosPickerItem] = []
+    @State private var inlineSelectedMediaItems: [PhotosPickerItem] = []
 
-    private let maxSelectionCount = 10
```
kean, line 17: It was unused. We could limit the selection size, but I'm not a fan of adding artificial limits unless it leads to crashes.

dcalhoun (approved):
Works well for me! 

> I replaced it [JS insert media logic invocation] with a new-ish [callAsyncJavaScript](https://developer.apple.com/documentation/webkit/wkwebview/callasyncjavascript(_:arguments:in:in:completionhandler:)) API.

Nifty. Is this something we should consider using more generally or for specific contexts? 

The fact the following interactions works so well blew my mind. 🤯 

https://github.com/user-attachments/assets/467dab09-c385-437f-a443-5eaff5b48889

kean:
> Is this something we should consider using more generally or for specific contexts?

I'm not sure if it can or should be used for synchronous JS, and I didn't want to change the existing code, but it worked really well for me here. I like that you can pass arguments as objects (Objective-C) without worrying about how to escape or encode them.


> The fact the following interactions works so well blew my mind. 🤯

I had no idea this was possible 🫨

PR #213, opened 2025-11-04
feat: Enable image resizing on mobile devices

dcalhoun:
## What?

Allow resizing images on mobile. This reverts explicit disabling of the feature in Gutenberg based on reports in
https://github.com/WordPress/gutenberg/issues/2675.

The resizing feature appears to work well enough today. We should contribute this change upstream.

## Why?

Close CMM-196. Avoid unnecessarily limiting the UX for mobile devices. 

## How?

Add a package patch removing the `isLargeViewport` requirement for resizing images. 

## Testing Instructions

1. Add an Image block. 
2. Attach media. 
3. Resize the image with drag handles; change the dimension fields within the block inspector.
4. **Verify** the rendered block is accurate, that the block attributes are saved as expected. 

N/A, no navigation changes. 

## Screenshots or screencast 
Before | After
-- | --
before-image | after-image
before-controls | after-controls


kean (approved):
Fantastic, and no performance/usability issues.

PR #214, opened 2025-11-05
feat: Block toolbar indicates scrollable area

dcalhoun:
## What?

Indicate the block toolbar is scrollable.

## Why?

Close CMM-924. Mitigate confusion regarding the number of options within the
block toolbar.

## How?

Add gradients to either side of the horizontal scroll view to communicate the
additional options available. This is only so effective, but may help some. 

## Testing Instructions

1. Select a block.
1. Scroll the block toolbar either direction.
1. **Verify** the gradient appears when additional options are available for a
   given side.

N/A, no navigation changes.

## Screenshots or screencast 
https://github.com/user-attachments/assets/68e5d6e7-4054-41ec-8200-fe1adc985165

### DIFF
--- a/src/components/editor-toolbar/index.jsx
+++ b/src/components/editor-toolbar/index.jsx
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { useState } from '@wordpress/element';
+import { useState, useRef } from '@wordpress/element';
 import {
 	BlockInspector,
 	BlockToolbar,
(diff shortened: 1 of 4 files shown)


kean (approved):
It looks cool, although I'm not sure how effective it is. It depends on what buttons are displayed in a bar. 

Screenshot 2025-11-05 at 3 14 20 PM

I would probably get rid of the separators in the toolbar. I'm not sure they add any value but take space.

dcalhoun:
> It looks cool, although I'm not sure how effective it is. It depends on what buttons are displayed in a bar.

Agreed, it is subtle and effectiveness quite limited; it's dependent upon screen width and where buttons fall. Dynamic spacing of buttons into "pages" like Apple Notes might be ideal long term. 

> I would probably get rid of the separators in the toolbar. I'm not sure they add any value but take space.

The separators are placed for logic groups—text formatting, alignment, etc. I feel there is _some_ value in that.

kean:
> I feel there is some value in that.

I agree, but it feels a bit too much, especially since there are many groups with just one item in them.

Screenshot 2025-11-05 at 3 30 00 PM

PR #216, opened 2025-11-05
Native Inserter: Add quick camera access

kean:
## What?
Closes [CMM-861: Add trailing bar button item to add media from Camera](https://linear.app/a8c/issue/CMM-861/add-trailing-bar-button-item-to-add-media-from-camera).

## Why?
The "Camera" is not as common as "Photos", but it the second most popular option, so it makes sense to provide quick access to it.

## How?
Similar to "Photos" picker.

## Testing Instructions
(See recording)

It requires a device for testing, so I tested it and recorded it, so you don't have to.

## Screenshots or screencast

https://github.com/user-attachments/assets/212a9bb5-2d7a-4e71-ac1a-1ce8c7566c39

### DIFF
--- a/ios/Demo-iOS/Gutenberg.xcodeproj/project.pbxproj
+++ b/ios/Demo-iOS/Gutenberg.xcodeproj/project.pbxproj
@@ -333,6 +333,7 @@
 				ENABLE_PREVIEWS = YES;
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_KEY_CFBundleURLTypes = "$(INFOPLIST_KEY_CFBundleURLTypes)";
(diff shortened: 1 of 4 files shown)


ios/Demo-iOS/Gutenberg.xcodeproj/project.pbxproj
```diff
@@ -333,6 +333,7 @@
 				ENABLE_PREVIEWS = YES;
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_KEY_CFBundleURLTypes = "$(INFOPLIST_KEY_CFBundleURLTypes)";
+				INFOPLIST_KEY_NSCameraUsageDescription = "The app needs access to your camera to capture photos and videos for your content.";
```
kean, line 336: The app already provides these, so no change will be need during integration.

dcalhoun (approved):
Works well!

PR #217, opened 2025-11-05
fix: Display drag-and-drop indicator

dcalhoun:
## What?

Display the block insertion point indicator when dragging blocks.

## Why?

Close CMM-731. Communicate the destination position for a dragged block.

## How?

The popover used for this feature was hidden previously because it
disrupts tap/click events in the block inserter, requiring two taps to
insert a block.

These changes continue hiding the "add block" button popover displayed
alongside empty text blocks, but allows the insertion point indicator to
now be displayed. We disable the insertion point display from the block
inserter only by detecting touch devices (as best possible) and prevent
the relevant `onMouseEnter` event callback triggering the `onHover`
callback.

### 1. Drag insertion point indicator is displayed
1. Insert several blocks, including a non-text block[^1].
1. Attempt to drag and drop the non-text block.
1. **Verify** the insertion point indicator (wide blue line) is displayed at the target location
   between the blocks.

### 2. Inserting blocks from the block inserter requires a single tap
1. Open the editor. 
2. Focus the placeholder block to insert a Paragraph block. 
3. Tap the `+` button open the block inserter. 
4. Tap a block. 
5. **Verify** the block is inserted after a single tap. a

[^1]: Dragging text blocks is difficult on web due to the `contenteditable`
    generally stealing focus and activate cursor/text actions.

N/A, no navigation changes.

## Screenshots or screencast 

https://github.com/user-attachments/assets/60063cce-6986-44dc-954b-c08fd19be26d

### DIFF
--- a/patches/@wordpress+block-editor+14.17.0.patch
+++ b/patches/@wordpress+block-editor+14.17.0.patch
@@ -1,3 +1,27 @@
+diff --git a/node_modules/@wordpress/block-editor/build-module/components/inserter-list-item/index.js b/node_modules/@wordpress/block-editor/build-module/components/inserter-list-item/index.js
+index 89a3bbc..8564798 100644
+--- a/node_modules/@wordpress/block-editor/build-module/components/inserter-list-item/index.js
++++ b/node_modules/@wordpress/block-editor/build-module/components/inserter-list-item/index.js
+@@ -28,6 +28,10 @@ function InserterListItem({
+   ...props
+ }) {
+   const isDraggingRef = useRef(false);
++  const isTouchDeviceRef = useRef(
++    window.matchMedia( '(hover: none)' ).matches ||
++    window.matchMedia( '(pointer: coarse)' ).matches
++  );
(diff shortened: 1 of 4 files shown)


kean (approved):
I'm surprised how drag-n-drop works in a web view, even on a device. I just want to share a couple of small observations and potential things to improve:

- "Gallery" seems to indicate a drop target for itself
- As you drag your finger the insertion point gets continuously animated (appear/disappear) even if there is no change
- The insertion points disappears over some blocks. For example, if you hold your finger over `core/block`, the insertion point is always shown as the closest to your finger. If you do it over `core/list`, it disappears
- I tried using drag-n-drop for text the same way as you can do in the Notes app, and it didn't work. 

https://github.com/user-attachments/assets/bae1b959-5f6a-4194-8be0-759b042202b9

dcalhoun:
Captured the reported improvement opportunities in CMM-936-linear-issue. Most, if not call, occur for cursor devices in the web editor, but worthwhile improvements nonetheless. Thank you for sharing!

PR #218, opened 2025-11-06
Native Inserter: Apply theme styles for pattern previews

kean:
## What?

Closes [CMM-922: Add support for using the current site theme to preview patterns](https://linear.app/a8c/issue/CMM-922/add-support-for-using-the-current-site-theme-to-preview-patterns).

### 1. Use theme styles from `EditorConfiguration.settings`

**Grid**: Before / After

Screenshot 2025-11-06 at 9 38 40 AM Screenshot 2025-11-06 at 9 31 35 AM

**List**: Before / After

Screenshot 2025-11-06 at 9 38 45 AM Screenshot 2025-11-06 at 9 19 59 AM

**Web**

The app does a better job computing the preview frame.

Screenshot 2025-11-06 at 9 20 01 AM

### 2. Use aspect ratio of the image in a full list

Before / After

Screenshot 2025-11-06 at 9 38 50 AM Screenshot 2025-11-06 at 9 29 30 AM

### 3. Add "toggle mode" button to the full list

Screenshot 2025-11-06 at 9 43 49 AM

### 4. Update "Patterns" button in the inserter to use a new image

This is the closest button I could find in [SF Symbols](https://developer.apple.com/sf-symbols/). We might want to ideally use Gutenberg icons, but we don't have access to them in GBK and have no asset catalog for now. I suggest do it later.

Screenshot 2025-11-06 at 9 44 28 AM

## Testing

I hardcoded editor settings in the demo for testing purposes. I tried to use `WordPressAPI` to fetch them, but it always fails with 401.


ios/Sources/GutenbergKit/Sources/Views/HTMLPreview/HTMLPreviewManager.swift
```diff
@@ -253,19 +270,27 @@ private func makeHTML(content: String, viewportWidth: Int, css: String) -> Strin
...
                 background: white;
-                display: flex;
-                align-items: center;
                 min-height: 100%;
             }
             .is-root-container {
                 width: 100%;
             }
+            /* Fix for WordPress image blocks with rounded corners in WebKit */
+            .wp-block-image {
```
kean, line 276: I don't know how else to fix it but without it the images were rendered like this:

Screenshot 2025-11-06 at 9 48 17 AM

ios/Sources/GutenbergKit/Sources/Views/HTMLPreview/HTMLPreviewView+Preview.swift
```diff
@@ -1,254 +0,0 @@
-#if DEBUG
-import SwiftUI
```
kean, line 2: It no longer had  support for theme styles, so I decided to nuke it. You can re-create it with a single prompt if you ever need it.

ios/Sources/GutenbergKit/Sources/Views/HTMLPreview/HTMLPreviewManager.swift
```diff
@@ -14,9 +14,8 @@ import SwiftUI
 /// - Disk cache: Stores full-size rendered images at pattern viewport width
 /// - Concurrent rendering with request queuing
 /// - Background execution of disk I/O operations
-public actor HTMLPreviewManager {
-    public static let shared = HTMLPreviewManager()
```
dcalhoun, line 18: Do we no longer utilize a singleton to support theme styles?

ios/Sources/GutenbergKit/Sources/Views/HTMLPreview/HTMLPreviewManager.swift
dcalhoun: For the editor, it appears the relevant styles are [injected by the `BlockCanvas`](https://github.com/WordPress/gutenberg/blob/c21a48d7826ff73f66ad73b1bdef13635063b557/packages/block-editor/src/components/block-canvas/index.js#L61-L65). We pass along the [styles as a prop](https://github.com/wordpress-mobile/GutenbergKit/blob/e892f6e003b04923d85d53ad43cbbfc0d8b4fcdf/src/components/visual-editor/index.jsx#L137). Since we are rendering the preview HTML in our pseudo-editor HTML environment, we lack those styles. 

I imagine more than the broken image styles will surface. We may need to do deeper testing, and possibly identify a more scalable solution. 

From testing briefly, I note the hard-code image styles are applied to images that are not intended to have rounded corners and there are missing spacing gaps between elements. 

<details><summary>Styling issues</summary>

Image
Image
Image

</details>

dcalhoun (approved):
This works well from my testing. Noting improvement opportunities worth following up on later.

PR #220, opened 2025-11-06
Native Inserter: Poor man internationalization

kean:
## What?
Fixes [CMM-874: Add localization for category names, action in the inserter, etc](https://linear.app/a8c/issue/CMM-874/add-localization-for-category-names-action-in-the-inserter-etc).

## How?
Since there aren't a lot of strings we need to localize, and the current translation process is only supported in the app, I suggest we just delegate the localization to the app.


dcalhoun (approved):
Demo app ran as expected. I tested a few localized strings in WP-iOS via the https://github.com/wordpress-mobile/WordPress-iOS/pull/24708 prototype build.

PR #221, opened 2025-11-07
Native Inserter: Add support for inserting existing site media

kean:
## What?
When user selects existing site media in the new inserter, it will now be inserted without re-uploading it.

## Why?
Matches how GBM and web works.

## How?
I'm not completely happy with the implementation, but at least I found [getBlockAndPreviewFromMedia](https://github.com/WordPress/gutenberg/blob/c21a48d7826ff73f66ad73b1bdef13635063b557/packages/block-editor/src/components/inserter/media-tab/media-tab.js#L80) in the media-tab.js. Unfortunately, it only deals with individual images and doesn't support galleries.

Alternative considered: remove the custom code that creates `core/gallery` and instead use `getBlockAndPreviewFromMedia` and insert selected media as individual `core/image` blocks. This way, the user can decide later whether to put them in a gallery. I don't know what the ideal UX is, but the current one poses some technical challenges. It might also not be what everyone wants.

## Testing Instructions
- Use the changes from https://github.com/wordpress-mobile/WordPress-iOS/pull/24708 and point to the local dev server

### DIFF
--- a/src/components/native-block-inserter-button/index.jsx
+++ b/src/components/native-block-inserter-button/index.jsx
@@ -5,7 +5,13 @@ import { Button } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { plus } from '@wordpress/icons';
 import { useSelect, useDispatch } from '@wordpress/data';
-import { findTransform, getBlockTransforms, parse } from '@wordpress/blocks';
+import {
+	findTransform,
+	getBlockTransforms,
+	parse,
+	createBlock,
+} from '@wordpress/blocks';
+import { getBlockAndPreviewFromMedia } from '@wordpress/block-editor/build-module/components/inserter/media-tab/utils';
 import { useRef } from '@wordpress/element';
 
 // NOTE: These hooks are internal WordPress APIs not available via public exports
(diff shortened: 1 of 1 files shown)


src/components/native-block-inserter-button/index.jsx
```diff
@@ -116,92 +122,181 @@ export default function NativeBlockInserterButton() {
...
-			onInsertBlocks( blocks );
-			return true;
+	/**
+	 * Insert new media files (without IDs) using WordPress file transforms.
+	 * Downloads files and uses transform system for block creation and upload.
+	 *
+	 * @param {Array} mediaArray Array of media objects without IDs
+	 * @return {Promise<boolean>} True if insertion succeeded
+	 */
+	const insertMediaWithoutIds = async ( mediaArray ) => {
```
kean, line 233: This code should be unchanged. I'm not sure why git doesn't recognize these as moved lines. I did run a smoke test.

dcalhoun (approved):
Works as expected. 🚀

dcalhoun:
Rebased the branch atop `trunk` as the changes in https://github.com/wordpress-mobile/GutenbergKit/commit/166d1e0b3514fd9da04b5d1b1b547f4e381f9fae were needed for testing the WordPress-iOS integration.

PR #224, opened 2025-11-11
fix: Mitigate Android `ERR_UPLOAD_FILE_CHANGE` errors

dcalhoun:
## What?

Mitigate `ERR_UPLOAD_FILE_CHANGE` errors occurring when uploading files selected from a cloud-based content provider on Android. 

## Why?

Fix CMM-782. Due to a [Chrome bug](https://issues.chromium.org/issues/40123366), uploading media often failed if the file was provided by a cloud-based content provider (e.g., the Google Drive app). The error is almost always thrown, but generally only causes video (or larger) file uploads to fail. 

This issue is also [noted in the `tus-js-client` project](https://github.com/tus/tus-js-client/issues/255#issuecomment-828386273), which is used by VideoPress. The `ERR_UPLOAD_FILE_CHANGE` error occurs for most all media uploads (images, videos, etc), but seemingly mostly causes VideoPress uploads to fail, likely due to the tus usage. 

The `ERR_UPLOAD_FILE_CHANGE` error is also _not_ thrown when using GutenbergKit/VideoPress from the Chrome browser directly. It only occurs in the Android WebView. 

## How?

Cache the selected file and upload it rather than the reference provided by the cloud-based content provider. This mirrors a [workaround suggested](https://issues.chromium.org/issues/40123366#comment80) on the Chrome issue tracker.  

We clear the cached files when the editor is destroyed to avoid a ballooning cache. 

## Testing Instructions

> [!note] 
> Prior to the changes, the Android demo app did not have logic in place for handling the file picker. To replicate the original failures, you can checkout the first commit of this branch. 

> [!tip]
> The WordPress-Android [prototype build](https://github.com/wordpress-mobile/WordPress-Android/pull/22347#issuecomment-3518623070) can be used for testing. 

1. Insert a Video block. 
2. Tap _Upload_. 
3. Select a video file from a cloud-based content provider (often times the default picker is driven by a cloud-based provider). 
4. **Verify** the upload succeeds. 

N/A no navigation changes. 

## Screenshots or screencast 

https://github.com/user-attachments/assets/b93d139b-fb7a-49bc-8af0-7158f3002616

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/FileCache.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/FileCache.kt
@@ -3,6 +3,7 @@ package org.wordpress.gutenberg
 import android.content.Context
 import android.net.Uri
 import android.provider.OpenableColumns
+import android.util.Log
 import android.webkit.MimeTypeMap
 import java.io.File
 import java.io.FileOutputStream
@@ -13,6 +14,7 @@ import java.io.IOException
  * in WebView when uploading files from cloud storage providers.
  */
 object FileCache {
+    private const val TAG = "FileCache"
     private const val CACHE_DIR_NAME = "gutenberg_file_uploads"
     private const val BUFFER_SIZE = 8192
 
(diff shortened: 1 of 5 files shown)


adalpari (approved):
The code looks solid. Tested it [here](https://github.com/wordpress-mobile/WordPress-Android/pull/22347#pullrequestreview-3453190233), and it's working as expected! :shipit: 

PS: Thanks for the video, it helps to quickly understand testing steps if there's anything the tester is not familiar with.

PR #225, opened 2025-11-11
feat: Discover Android demo app configuration

dcalhoun:
## What?

Mirror the [iOS demo app](https://github.com/wordpress-mobile/GutenbergKit/blob/ef47b0924274364611efe3a5d6444ecda92f4905/ios/Demo-iOS/Sources/Views/AppRootView.swift#L61-L91) by configuring `plugins` and `themeStyles` options through site API discovery. 

## Why?

Allows remote editor configurations to match the capabilities of the site itself. 

## How?

Leverage the `WpApiClient` to retrieve the available routes for a site. Configure the editor for a site when launching the editor. 

> [!note]
> It appears the `WpApiClient` caches the values retrieved, so they become stale after initial retrieval. A future improvement would be updating the `wordpress-rs` project to allow disabling caching for these requests, like [iOS does](https://github.com/wordpress-mobile/GutenbergKit/blob/ef47b0924274364611efe3a5d6444ecda92f4905/ios/Demo-iOS/Sources/Services/AuthenticationManager.swift#L31). 

## Testing Instructions

1. Add a remote editor configuration entry. 
2. **Verify** the resulting GBK configuration matches that of the site.[^1]

[^1]: The Gutenberg plugin defines the `/wp-block-editor/v1/settings` route, enabling `themeStyles`; the Jetpack plugin defines the `/wpcom/v2/editor-assets` route, enabling `plugins`.  

N/A, no user-facing changes. 

## Screenshots or screencast 
N/A, no user-facing changes.

### DIFF
--- a/android/app/src/main/java/com/example/gutenbergkit/MainActivity.kt
+++ b/android/app/src/main/java/com/example/gutenbergkit/MainActivity.kt
@@ -44,14 +44,19 @@ import com.example.gutenbergkit.ui.dialogs.AddConfigurationDialog
 import com.example.gutenbergkit.ui.dialogs.DeleteConfigurationDialog
 import com.example.gutenbergkit.ui.dialogs.DiscoveringSiteDialog
 import com.example.gutenbergkit.ui.theme.AppTheme
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 import org.wordpress.gutenberg.BuildConfig
 import org.wordpress.gutenberg.EditorConfiguration
 
 class MainActivity : ComponentActivity(), AuthenticationManager.AuthenticationCallback {
     private val configurations = mutableStateListOf<ConfigurationItem>()
     private val isDiscoveringSite = mutableStateOf(false)
+    private val isLoadingCapabilities = mutableStateOf(false)
     private lateinit var configurationStorage: ConfigurationStorage
     private lateinit var authenticationManager: AuthenticationManager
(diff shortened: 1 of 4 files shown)


android/app/src/main/java/com/example/gutenbergkit/MainActivity.kt
```diff
@@ -112,14 +118,42 @@ class MainActivity : ComponentActivity(), AuthenticationManager.AuthenticationCa
...
-            .setPlugins(true)
-            .setSiteURL(config.siteUrl)
-            .setSiteApiRoot(config.siteApiRoot)
-            .setNamespaceExcludedPaths(arrayOf())
-            .setAuthHeader(config.authHeader)
-            .build()
+    private fun loadRemoteEditor(config: ConfigurationItem.RemoteEditor) {
+        isLoadingCapabilities.value = true
+
+        CoroutineScope(Dispatchers.Main).launch {
```
adalpari: ⛏️ Just as a good practice, it's preferred to use `lifecycleScope` instead of creating or using generic ones. This helps with memory leaks and manual management since the scope is attached to the activity lifecycle. E.g: if the activity is destroyed, the job is automatically cancelled.

adalpari (approved):
LGTM! I just left a minor cpomment

android/app/src/main/java/com/example/gutenbergkit/MainActivity.kt
dcalhoun: Thank you for this feedback. Very helpful. Addressed in f9c8015a23d678d963f4ef92a5eb5565a365a87b.

PR #226, opened 2025-11-12
task: Improve demo app consistency

dcalhoun:
> [!tip]
> I recommend reviewing each commit individually to reduce complexity. 

## What?

Update the demo app to avoid the "remote editor" term. Improve consistency between the Android and iOS demo apps. 

## Why?

The "remote editor" term is no longer relevant and may cause confusion. 

## How?

Update all strings and class/method names. 

## Testing Instructions

Smoke test the demo app. 

N/A, not relevant for this rename. 

## Screenshots or screencast 
N/A, not relevant for this rename.


adalpari (approved):
Nice improvements!!
LGTM!

PR #227, opened 2025-11-18
fix: Incorrect media attachment position

dcalhoun:
## What?

Ensure the selected media is attached to the correct location in the editor canvas. 

## Why?

Fix CMM-925. Previously, selected media would always be placed in the last open media placeholder slot. This resulted in incorrect media attachments. 

## How?

Replace the singleton media upload callback with callbacks unique to each media placeholder slot. A unique ID is associated with each media placeholder, which is then used as a unique callback once the native host has selected media. 

## Testing Instructions

> [!note]
> Testing likely requires the WordPress host app for accessing the _Media Library_ source. 

**1. Multiple media uploads instances**
1. Open the editor. 
2. Insert more than one _Image_ blocks. 
3. Select the an _Image_ block that is not the last one. 
4. Attach media to the block. 
5. **Verify** the media is attached to the correct block. 

**2. Reordered media upload instances**
1. Open the editor. 
2. Insert an _Image_ block. 
3. Insert an _Video_ block. 
4. Press the "move block upward" button to move the _Video_ block to the first position. 
6. Attach media to the block. 
7. **Verify** the media is attached to the correct block. 

N/A, no navigation changes. 

## Screenshots or screencast 
N/A, no visual changes.

### DIFF
--- a/src/components/editor/use-dev-mode-notice.js
+++ b/src/components/editor/use-dev-mode-notice.js
@@ -15,8 +15,7 @@ import { isDevMode } from '../../utils/dev-mode';
 let noticeShown = false;
 
 export function useDevModeNotice() {
-	const { createWarningNotice, removeAllNotices } =
-		useDispatch( noticesStore );
+	const { createWarningNotice } = useDispatch( noticesStore );
 
 	useEffect( () => {
 		const hasDevMode = isDevMode();
@@ -34,5 +33,5 @@ export function useDevModeNotice() {
 				}
 			);
 		}
-	}, [ createWarningNotice, removeAllNotices ] );
+	}, [ createWarningNotice ] );
 }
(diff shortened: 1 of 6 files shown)


src/components/editor/use-dev-mode-notice.js
```diff
@@ -15,8 +15,7 @@ import { isDevMode } from '../../utils/dev-mode';
 let noticeShown = false;
 
 export function useDevModeNotice() {
-	const { createWarningNotice, removeAllNotices } =
```
dcalhoun, line 18: Merely removing an unused callback.

kean (approved):
It worked well, but I noticed the lack of upload progress (unrelated to the change in this PR).

dcalhoun:
> [...] I noticed the lack of upload progress (unrelated to the change in this PR).

@kean specifically the Jetpack Image Compare block or all media blocks? If the former, I believe that is the implementation of that specific block; there is no indicator when using it in a web browser either.

kean:
It only happens with the Image Compare block that was mentioned in the ticket.

PR #228, opened 2025-11-18
Add native navigation bar overlay so modals better fit in

kean:
## What?
Fake the bottom sheet presentation style on iOS by using the existing modal show/hide events.

## Why?
It makes it visually clear to the user that the "Block Inspector" is a modal and make it blend in better with the native navigation controls.

## How?
Adds `is-ios` class for iOS-specific styles.

## Screenshots or screencast 

Before / After

Screenshot 2025-11-19 at 6 06 56 AM Screenshot 2025-11-18 at 11 06 32 AM

## Alternatives Considered

Show "Block Inspector" covering only half of the screen and keep the rest of the editor interactive, which is challenging to achieve without introducing a lot of small little bugs. This PR is a stop-gap solution for the initial release, or maybe is a long-term solution since there are also advantages of showing this screen fullscreen.

### DIFF
--- a/src/index.scss
+++ b/src/index.scss
@@ -237,3 +237,16 @@ $baseline-interactive-font-size: 17px;
 		min-width: 274px !important;
 	}
 }
+
+// iOS-specific styling for block settings menu
+body.is-ios.gutenberg-kit {
+	.block-settings-menu {
+		background-color: rgba(0, 0, 0, 0.3) !important;
+	}
+
+	.block-settings-menu .components-popover__content {
+		border-top-left-radius: 26px !important;
+		border-top-right-radius: 26px !important;
+		background-color: #ffffff !important;
+	}
+}
--- a/src/index.scss
+++ b/src/index.scss
@@ -239,14 +239,12 @@ $baseline-interactive-font-size: 17px;
 }
 
 // iOS-specific styling for block settings menu
-body.is-ios.gutenberg-kit {
-	.block-settings-menu {
-		background-color: rgba(0, 0, 0, 0.3) !important;
-	}
+.gutenberg-kit.is-ios .block-settings-menu {
+	background-color: rgba(0, 0, 0, 0.3);
 
-	.block-settings-menu .components-popover__content {
-		border-top-left-radius: 26px !important;
-		border-top-right-radius: 26px !important;
-		background-color: #ffffff !important;
+	& .components-popover__content {
+		background-color: #ffffff;
+		border-top-left-radius: 26px;
+		border-top-right-radius: 26px;
 	}
 }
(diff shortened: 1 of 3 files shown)


kean:
(not in the scope of the PR) We should probably also hide `block-editor-block-card` and just show a block name as a "title" to reduce the amount of of vertical space used. You don't need to see all this information every time. The sheet header also shouldn't scroll together with the content. The "close" button should stay fixed.

src/index.scss
```diff
@@ -237,3 +237,16 @@ $baseline-interactive-font-size: 17px;
...
+	.block-settings-menu {
+		background-color: rgba(0, 0, 0, 0.3) !important;
+	}
+
+	.block-settings-menu .components-popover__content {
+		border-top-left-radius: 26px !important;
+		border-top-right-radius: 26px !important;
+		background-color: #ffffff !important;
+	}
+}
```
dcalhoun: I feel we should:

- Avoid `!important` every whenever possible, it often creates future specificity conflicts. 
- Use the lowest level of selector specificity necessary to allow reconciling future, unpredictable styling needs. 
- If we are going to rely upon the Sass nesting syntax, leverage it to its full extent to avoid repetition. 
- Alphabetize attributes to simplify searching and modifying code. 

Feel free to share counter thoughts. 

```suggestion
.gutenberg-kit.is-ios .block-settings-menu {
	background-color: rgba(0, 0, 0, 0.3);

	& .components-popover__content {
		background-color: #ffffff;
		border-top-left-radius: 26px;
		border-top-right-radius: 26px;
	}
}
```

dcalhoun (approved):
Functions as expected! 🚀 

> We should probably also hide `block-editor-block-card` and just show a block name as a "title" to reduce the amount of of vertical space used. You don't need to see all this information every time. 

Possibly. While it's easy enough to hide it with CSS selectors, I consider CSS selectors to be private APIs; there is no guarantee they will not change. Ideally, there is configuration props for the block inspector allowing us to hide this information. 

> The sheet header also shouldn't scroll together with the content. The "close" button should stay fixed.

Gutenberg's `Popover` component includes a proper fixed header feature (with close button), but it's not displayed in GutenbergKit's current implementation as we do not use the `expandOnMobile` prop; we instead apply positioning styles manually. 

I previously [explored refactoring](https://github.com/wordpress-mobile/GutenbergKit/compare/trunk...ae5cc4d266c5eea72f5ac4c0bd67bb93421685e7#diff-3910666fecd47590c97354f78c0ab3ce4d5c363617576862c227b725b66798b0R121) GutenbergKit's implementation to avoid the [positioning styles](https://github.com/wordpress-mobile/GutenbergKit/commit/906663929215b1a6044d8db7f238cc2dcb754a3a#diff-6ea08c95a60c10153bc1d5ae1e330e1fa103cc61df85954a91f741359450a5b0L32-L40), but encountered other issues that hampered my progress—scrolling behavior, focus management, a11y, etc. Ideally we further embrace (and contribute towards) Gutenberg's components rather than overriding their styles to suit our needs.

PR #230, opened 2025-11-20
fix: Inline inserter opens native inserter

dcalhoun:
## What?

Ensure inline inserter's _Browse all_ button opens the native block inserter. 

It's worth noting the current outcome is not perfect. Namely, the quick inserter remains open after opening the native inserter. This occurs because the former relies upon losing focus for its dismissal. The native inserter does not change the WebView's focus state; thus, the quick inserter remains open. 

Long term, I think the most ideal case is that we introduce a new filter hook to the Gutenberg project that allows replacing the inserter used throughout the UI. I.e., when tapping the quick inserter's inline button, it opens the native inserter directly rather than displaying the quick inserter. I captured this as a note in https://github.com/wordpress-mobile/GutenbergKit/issues/106. 

## Why?

Fix CMM-973. The bug made inserting blocks into layout blocks difficult. 

## How?

* Control the native inserter visibility via the existing `isInserterOpen` store state, ensuring the state is correct throughout the UI. 
* Memoize various callbacks to avoid unnecessary re-renders—this addresses some of https://github.com/wordpress-mobile/GutenbergKit/pull/199#discussion_r2474322561. 
* Disable existing a11y logic that is unnecessary for the native inserter, as it manages the relevant a11y state itself (e.g., marking content outside the modal dialog inert). 

## Testing Instructions

1. Open the inserter. 
2. Select _Columns_. 
3. Tap the inline inserter button. 
4. Tap _Browse all_. 
5. Select a block to insert. 
6. **Verify** the selected block is inserted. 

Navigate the testing steps. **Verify** content is correct marked inert when the native inserter is opened. 

## Screenshots or screencast 

https://github.com/user-attachments/assets/1958ab88-fd42-4dc7-aab0-cedbf2ecfffb

### DIFF
--- a/src/components/editor-toolbar/index.jsx
+++ b/src/components/editor-toolbar/index.jsx
@@ -98,7 +98,10 @@ const EditorToolbar = ( { className } ) => {
 	);
 
 	const addBlockButton = enableNativeBlockInserter ? (
-		<NativeBlockInserterButton />
+		<NativeBlockInserterButton
+			open={ isInserterOpened }
+			onToggle={ setIsInserterOpened }
+		/>
 	) : (
 		<Inserter
 			popoverProps={ {
--- a/src/components/editor-toolbar/index.jsx
+++ b/src/components/editor-toolbar/index.jsx
@@ -60,10 +60,13 @@ const EditorToolbar = ( { className } ) => {
 	}, [] );
 	const { setIsInserterOpened } = useDispatch( editorStore );
 
-	useModalize( isInserterOpened );
+	useModalize( isInserterOpened && ! enableNativeBlockInserter );
 	useModalize( isBlockInspectorShown );
 
-	useModalDialogState( isInserterOpened, 'block-inserter' );
+	useModalDialogState(
+		isInserterOpened && ! enableNativeBlockInserter,
+		'block-inserter'
+	);
 	useModalDialogState( isBlockInspectorShown, 'block-inspector' );
 
 	function openSettings() {
--- a/src/components/native-block-inserter-button/index.jsx
+++ b/src/components/native-block-inserter-button/index.jsx
@@ -12,7 +12,7 @@ import {
 	createBlock,
 } from '@wordpress/blocks';
 import { getBlockAndPreviewFromMedia } from '@wordpress/block-editor/build-module/components/inserter/media-tab/utils';
-import { useRef } from '@wordpress/element';
+import { useRef, useEffect, useCallback } from '@wordpress/element';
 
 // NOTE: These hooks are internal WordPress APIs not available via public exports
 // or privateApis. We import from build-module as the only way to access the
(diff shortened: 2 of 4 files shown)


src/components/editor-toolbar/index.jsx
```diff
@@ -60,10 +60,13 @@ const EditorToolbar = ( { className } ) => {
...
 
-	useModalize( isInserterOpened );
+	useModalize( isInserterOpened && ! enableNativeBlockInserter );
 	useModalize( isBlockInspectorShown );
 
-	useModalDialogState( isInserterOpened, 'block-inserter' );
+	useModalDialogState(
+		isInserterOpened && ! enableNativeBlockInserter,
+		'block-inserter'
+	);
```
dcalhoun, line 69: These marked content outside of the block inserter modal dialog "inert," and thus inaccessible by screen readers. This is unnecessary for the native inserter, as the native UI handles this itself. The WebView is inaccessible to screen readers when the native inserter is open.

src/components/editor-toolbar/index.jsx
```diff
@@ -95,7 +98,10 @@ const EditorToolbar = ( { className } ) => {
 	);
 
 	const addBlockButton = enableNativeBlockInserter ? (
-		<NativeBlockInserterButton />
+		<NativeBlockInserterButton
+			open={ isInserterOpened }
+			onToggle={ setIsInserterOpened }
+		/>
```
dcalhoun, line 104: This API matches the Gutenberg `Inserter` a few lines beneath this.

src/components/native-block-inserter-button/index.jsx
```diff
@@ -118,276 +124,322 @@ export default function NativeBlockInserterButton() {
...
-					media,
-					mediaType
+			/**
+			 * Insert media from WordPress media library (with existing IDs).
+			 * Creates blocks directly with media attributes, avoiding re-upload.
+			 *
+			 * @param {Array} items Array of media objects with IDs
+			 * @return {boolean} True if insertion succeeded
+			 */
+			const insertMediaWithIds = ( items ) => {
```
dcalhoun, line 203: Renamed to `items` to avoid duplicating the `mediaArray` name in the upper scope now that `insertMediaWithIds` is declared within `insertMedia`. It was relocated to simplify memoization.

src/components/native-block-inserter-button/index.jsx
```diff
@@ -118,276 +124,322 @@ export default function NativeBlockInserterButton() {
...
+
+	// Watch for controlled open state changes
+	useEffect( () => {
+		// Only trigger when transitioning from false to true
+		if ( open && ! prevOpen.current ) {
+			prepareAndShowInserter();
 		}
-	};
+		prevOpen.current = open;
+	}, [ open, prepareAndShowInserter ] );
```
dcalhoun, line 434: This is the crux of the change. Rather than `onClick` directly showing the inserter, the global `isInserterOpen` state now does so. This allows UI outside of this component's `Button` to open the native inserter.

src/components/native-block-inserter-button/index.jsx
```diff
@@ -118,276 +124,322 @@ export default function NativeBlockInserterButton() {
...
+		]
+	);
 
-	const insertPattern = ( patternName ) => {
-		const pattern = patterns?.find( ( p ) => p.name === patternName );
-		if ( ! pattern ) {
-			debug( `Pattern "${ patternName }" not found` );
-			return false;
+	const prepareAndShowInserter = useCallback( () => {
+		const sections = preprocessBlockTypesForNativeInserter(
```
dcalhoun, line 360: I used logging and the React Developer Tools to verify this complex logic is only run when opening the inserter.

kean (approved):
Thank you for address it. The code looks good.

As a potential future improvement, I would replacing the "quick inserter" itself with a native inserter (just show is as a small sheet the way it is shown now).

Screenshot 2025-11-20 at 2 47 12 PM

dcalhoun:
> As a potential future improvement, I would replacing the "quick inserter" itself with a native inserter (just show is as a small sheet the way it is shown now).

Agreed. That is what I meant by the following in the PR description:

> Long term, I think the most ideal case is that we introduce a new filter hook to the Gutenberg project that allows replacing the inserter used throughout the UI. I.e., when tapping the quick inserter's inline button, it opens the native inserter directly rather than displaying the quick inserter. I captured this as a note in https://github.com/wordpress-mobile/GutenbergKit/issues/106.

PR #232, opened 2025-11-20
Add support for Editor Settings

kean:
## What?
The demo app will now honor site's block editor settings (`wp-block-editor/v1/settings`).

## Why?

The idea is to use the newly introduced `EditorService` to manage the initial load and subsequent refresh of all resources required to launch the editor: editor settings and assets. 

In the future PRs, I plan to simplify the code in `EditorAssetsLibrary` and move it to `EditorService`. It will fetch the manifest and the assets before "starting" the editor and will inject the preprocessed manifest using the configuration. This will give the host app control over when and how the assets are loaded and stored. With the new implementation, the requests will run in parallel, cutting the first load time by around 900ms (due to parallelization alone).

I will then generate the Android implementation based on the Swift code.

## How?

The new `EditorService` provides two simply APIs for the host app. 

The `setup` method is what you call to configure the editor. On first load, it will fetch the required resources (blocking operation). On the subsequent loads, it will simply use the previously loaded data (changes infrequently). It's up to the app how to handle errors.

```swift
public func setup(_ configuration: inout EditorConfiguration) async throws
```

The host app will also be responsible for calling `refresh` to update the previously fetched resources (or load the initial ones). It's safe to call `refresh` multiple times as it will attach to the existing task if present.  It's up to the app how to handle errors.

```swift
public func refresh() async throws
```

## Testing Instructions
Add a new "remote" editor and verify that the styles are applied.

## Screenshots or screencast

Before / After
Screenshot 2025-11-20 at 12 41 36 PM Screenshot 2025-11-20 at 12 41 05 PM

### DIFF
--- a/ios/Demo-iOS/Sources/Views/AppRootView.swift
+++ b/ios/Demo-iOS/Sources/Views/AppRootView.swift
@@ -72,7 +72,7 @@ struct AppRootView: View {
                 let canUsePlugins = apiRoot.hasRoute(route: "/wpcom/v2/editor-assets")
                 let canUseEditorStyles = apiRoot.hasRoute(route: "/wp-block-editor/v1/settings")
 
-                let updatedConfiguration = EditorConfigurationBuilder()
+                var updatedConfiguration = EditorConfigurationBuilder()
                     .setShouldUseThemeStyles(canUseEditorStyles)
                     .setShouldUsePlugins(canUsePlugins)
                     .setSiteUrl(config.siteUrl)
@@ -81,6 +81,19 @@ struct AppRootView: View {
                     .setNativeInserterEnabled(isNativeInserterEnabled)
                     .setLogLevel(.debug)
                     .build()
+
+                if let baseURL = URL(string: config.siteApiRoot) {
+                    let service = EditorService(
+                        siteID: config.siteUrl,
+                        baseURL: baseURL,
+                        authHeader: config.authHeader
+                    )
+                    do {
+                        try await service.setup(&updatedConfiguration)
+                    } catch {
+                        print("Failed to setup editor environment, confinuing with the default or cached configuration:", error)
+                    }
+                }
 
                 self.activeEditorConfiguration = updatedConfiguration
             } catch {
--- a/ios/Sources/GutenbergKit/Sources/Helpers/String+SafeDirectoryName.swift
+++ b/ios/Sources/GutenbergKit/Sources/Helpers/String+SafeDirectoryName.swift
@@ -1,0 +1,38 @@
+import Foundation
+
+extension String {
+    /// Converts a string (such as a URL) into a safe directory name by removing illegal filesystem characters
+    ///
+    /// This method filters out characters that are not allowed in directory names across different filesystems,
+    /// including: `/`, `:`, `\`, `?`, `%`, `*`, `|`, `"`, `<`, `>`, newlines, and control characters.
+    ///
+    /// Example:
+    /// ```swift
+    /// let url = "https://example.com/path?query=1"
+    /// let safeName = url.safeFilename
+    /// // Result: "https---example.com-path-query-1"
+    /// ```
+    var safeFilename: String {
+        // Define illegal characters for directory names
+        let illegalChars = CharacterSet(charactersIn: "/:\\?%*|\"<>")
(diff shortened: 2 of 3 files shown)


ios/Sources/GutenbergKit/Sources/Helpers/String+SafeDirectoryName.swift
```diff
@@ -0,0 +1,38 @@
...
+    /// This method filters out characters that are not allowed in directory names across different filesystems,
+    /// including: `/`, `:`, `\`, `?`, `%`, `*`, `|`, `"`, `<`, `>`, newlines, and control characters.
+    ///
+    /// Example:
+    /// ```swift
+    /// let url = "https://example.com/path?query=1"
+    /// let safeName = url.safeFilename
+    /// // Result: "https---example.com-path-query-1"
+    /// ```
+    var safeFilename: String {
```
kean, line 15: We could just replace it with sha256 as the folder name is only needed for technical purposes. wdyt?

kean:
Oh, actually, I'll call `setup` automatically in `EditorViewController`. The host app will just need to know about the `refresh`.

ios/Demo-iOS/Sources/Views/AppRootView.swift
```diff
@@ -82,6 +82,19 @@ struct AppRootView: View {
...
+                if let baseURL = URL(string: config.siteApiRoot) {
+                    let service = EditorService(
+                        siteID: config.siteUrl,
+                        baseURL: baseURL,
+                        authHeader: config.authHeader
+                    )
+                    do {
+                        try await service.setup(&updatedConfiguration)
+                    } catch {
+                        print("Failed to setup editor environment, confinuing with the default or cached configuration:", error)
```
dcalhoun, line 94:
```suggestion
                        print("Failed to set up editor environment, continuing with the default or cached configuration:", error)
```

ios/Sources/GutenbergKit/Sources/Helpers/String+SafeDirectoryName.swift
dcalhoun: Yeah, hashing is likely sufficient for this purpose. 

A readable name might make debugging easier, but that'd likely be infrequent and manageable.

ios/Sources/GutenbergKit/Sources/Service/EditorService.swift
```diff
@@ -0,0 +1,114 @@
+import Foundation
+
+/// Service for fetching the editor settings and other parts of the enrvironment
```
dcalhoun:
```suggestion
/// Service for fetching the editor settings and other parts of the environment
```

dcalhoun (approved):
Very nice! 

I suppose we could consider adding a _Use theme styles_ toggle beneath the _Native Inserter_ toggle, if helpful.

ios/Sources/GutenbergKit/Sources/Service/EditorService.swift
```diff
@@ -0,0 +1,114 @@
...
+    ///   - baseURL: Root URL for the site API
+    ///   - authHeader: Authorization header value
+    ///   - urlSession: URLSession to use for network requests (defaults to .shared)
+    public init(siteID: String, baseURL: URL, authHeader: String, urlSession: URLSession = .shared) {
+        self.siteID = siteID
+        self.baseURL = baseURL
+        self.authHeader = authHeader
+        self.urlSession = urlSession
+
+        self.storeURL = URL.documentsDirectory
```
kean, line 32: This should go into "Application" not "Documents".

ios/Sources/GutenbergKit/Sources/Helpers/String+SafeDirectoryName.swift
kean: I was also leaning towards hashing. It's simpler and more robust.

kean:
> Use theme styles toggle beneath the Native Inserter toggle, if helpful.

Good idea. It's a bit difficult due to how it's currently structured. I'll add a note to add it after other changes are done.
I'm still thinking about adding a "Preferences" panel. You need a discoverable way to toggle it quickly, and we might use it to show Gutenberg version and information about plugins as a starter.

PR #233, opened 2025-11-20
Faster and more robust editor asset management

kean:
## What?

The subsequent editor loads are now **~1.3s** faster. The first launch is **~1s** faster.

There are a few changes compared to the previous implementation:

- Move code that fetches assets (manifest and files) to the new `EditorService`. The assets manifest is now fetched in parallel with the editor settings.
- Only `EditorService` now triggers the actual download of the assets. The assets are put into a `GutenbergKit/{siteURL}/assets` folder.
- Remove `EditorAssetsProvider` and instead inject the plugin styles and scripts in `window.GBKit` when launching the editor (the same way it's done for editor settings). It happens only if `EditorService` verifies that all the required files exist on disk. If not, the plugins are disabled until the next time.
- `EditorService` reads editor settings and assets manifest in one atomic operation. When new manifest is loaded, it replaces the current one only after fetching all assets for it. Until then, the editor will always use the previous version of the manifest.
- Remove `editorSettings` from `EditorConfiguration`. Add them to internal `EditorDependencies` types fetched by `EditorViewController`

**Minor changes:**
- Remove `editorAssetsEndpoint` parameter from `EditorConfiguration`
- Make logging a cross-cutting concern, so it could be configured once and would be easier to use across GBK 

**Planned future changes:**

- Cleanup unused assets on launch to make sure the disk usage does not grow
- Improve how and when refresh and assets download is triggered
- Cover with unit tests and generate the same code in Kotlin

## Screenshots or screencast 

Before / After 

**Note**: the *preparing editor* message disappears quicker because "fetch editor settings" is now a step inside `EditorViewController`.

https://github.com/user-attachments/assets/bdf6cbbd-4561-464a-bfc1-61a7ef02901b


ios/Sources/GutenbergKit/Sources/Service/EditorAssetsManifest.swift
```diff
@@ -0,0 +1,106 @@
+import Foundation
+import CryptoKit
+import SwiftSoup
+
+struct EditorAssetsManifest: Codable {
```
kean, line 5: This is an existing `struct`. I made no changes to it – just renamed the file.

ios/Demo-iOS/Sources/Views/AppRootView.swift
```diff
@@ -109,13 +99,15 @@ struct AppRootView: View {
     }
 
     private func createBundledConfiguration() -> EditorConfiguration {
-        EditorConfigurationBuilder()
+        let config = EditorConfigurationBuilder()
             .setShouldUsePlugins(false)
-            .setSiteUrl("")
+            .setSiteUrl("bundled")
```
dcalhoun: What is the purpose of this change? Is this merely an identifier of why the configuration is empty?

ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
```diff
@@ -26,16 +26,10 @@ public struct EditorConfiguration: Sendable {
...
     /// Authorization header
     public let authHeader: String
-    /// Raw block editor settings from the WordPress REST API
-    public let editorSettings: String
     /// Locale used for translations
     public let locale: String
     /// Enables the native inserter UI in the editor
     public let isNativeInserterEnabled: Bool
-    /// Endpoint for loading editor assets, used when enabling `shouldUsePlugins`
-    public var editorAssetsEndpoint: URL?
```
dcalhoun, line 36: Why are we removing `editorAssetsEndpoint`? I believe it's useful in the future whenever we change the REST API endpoint used, it also allows alternatives for third-parties as needed.

ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
```diff
@@ -26,16 +26,10 @@ public struct EditorConfiguration: Sendable {
...
-    /// Logs emitted at or above this level will be printed to the debug console
-    public let logLevel: LogLevel
```
dcalhoun, line 38: Are host apps still able to configure the log levels emitted from the JavaScript logger utility? I believe that to be beneficial.

ios/Sources/GutenbergKit/Sources/EditorLogging.swift
```diff
@@ -0,0 +1,45 @@
...
+public enum EditorLogger {
+    /// The shared logger instance used throughout GutenbergKit.
+    public nonisolated(unsafe) static var shared: EditorLogging?
+
+    /// The log level. Messages below this level are ignored.
+    public nonisolated(unsafe) static var logLevel: EditorLogLevel = .error
+}
+
+func log(_ level: EditorLogLevel, _ message: @autoclosure () -> String) {
+    guard level.priority >= EditorLogger.logLevel.priority,
```
dcalhoun, line 22: The [previous idea](https://github.com/wordpress-mobile/GutenbergKit/pull/202#discussion_r2460868311) was allowing the JavaScript logger to control this to avoid duplication. I suppose we must add this now so that we can also log directly from the iOS code, correct?

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -157,14 +150,15 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
...
+            editorSettings: \(dependencies?.editorSettings ?? "undefined"),
             locale: '\(configuration.locale)',
             post: {
                 id: \(configuration.postID ?? -1),
                 title: '\(configuration.escapedTitle)',
                 content: '\(configuration.escapedContent)'
             },
-            logLevel: '\(configuration.logLevel)'
+            logLevel: '\(EditorLogger.logLevel.rawValue)',
+            manifest: \(dependencies?.manifest ?? "undefined")
```
dcalhoun: Any performance implications from adding this? 🤔  I suppose it is no larger than `editorSettings`, eh? 

I wonder if it's worth spreading the properties over the top-level object. I could see one day allowing GBK configuration to control `allowedBlockTypes`.

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -247,8 +246,16 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         guard !_isEditorSetup else { return }
         _isEditorSetup = true
 
-        setUpEditor()
-        loadEditor()
+        Task { @MainActor in
+            do {
+                self.dependencies = try await service.dependencies(for: configuration)
+            } catch {
+                print("Failed to setup editor environment, continuing with the default or cached configuration:", error)
```
dcalhoun:
```suggestion
                print("Failed to set up the site's editor configuration, continuing with the default or cached configuration:", error)
```

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -483,14 +490,19 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
...
 
         // Retain for 5 seconds and let it prefetch stuff
         DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(5)) {
             _ = editorViewController
         }
+
+        Task {
+            await EditorService.shared(for: configuration.siteURL)
+                .refresh(configuration: configuration)
+        }
```
dcalhoun: Is this not invoked from opening the view a few lines above?

src/utils/bridge.js
```diff
@@ -336,13 +337,15 @@ export function awaitGBKitGlobal( timeoutMs = 3000 ) {
...
+	const gbkit = getGBKit();
+
+	// iOS provides manifest directly in GBKit configuration
+	if ( window.webkit && gbkit.manifest ) {
+		return gbkit.manifest;
 	}
 
-	const { siteApiRoot, editorAssetsEndpoint, authHeader } = getGBKit();
+	// Android fallback - fetch from API
+	const { siteApiRoot, editorAssetsEndpoint, authHeader } = gbkit;
```
dcalhoun: I don't believe there is a downside to de-structuring all values up front. 

```suggestion
	const { manifest, siteApiRoot, editorAssetsEndpoint, authHeader } = getGBKit();

	// iOS provides manifest directly in GBKit configuration
	if ( window.webkit && manifest ) {
		return manifest;
}

	// Android fallback - fetch from API
```

Alternatively, I don't believe there is a downside to invoking `getGBKit()` multiple times, as it [early returns](https://github.com/wordpress-mobile/GutenbergKit/blob/dac7dc6beda3915fbca83dc09bfc1aee721a440d/src/utils/bridge.js#L195-L197) with the cached value on the `window`. 


```suggestion
	const { manifest } = getGBKit();

	// iOS provides manifest directly in GBKit configuration
	if ( window.webkit && gbkit.manifest ) {
		return gbkit.manifest;
}

	// Android fallback - fetch from API
	const { siteApiRoot, editorAssetsEndpoint, authHeader } = getGBKit();
```

ios/Sources/GutenbergKit/Sources/Service/EditorService.swift
```diff
@@ -1,78 +1,171 @@
...
-    /// Returns `true` is the resources requied for the editor already exist.
+    /// Returns `true` if the resources required for the editor already exist.
     private var isEditorLoaded: Bool {
-        FileManager.default.fileExists(atPath: editorSettingsFileURL.path())
+        FileManager.default.fileExists(atPath: stateFileURL.path)
     }
 
     /// Refresh the editor resources.
-    public func refresh() async throws {
+    /// Will not refresh more often than once every 30 seconds.
```
dcalhoun: Is the specific context where rapid invocations might occur or is merely a safeguard?

ios/Sources/GutenbergKit/Sources/EditorLogging.swift
```diff
@@ -0,0 +1,45 @@
...
+          let logger = EditorLogger.shared else {
+        return
+    }
+    logger.log(level, message())
```
dcalhoun, line 26: The JavaScript loggers [prepends](https://github.com/wordpress-mobile/GutenbergKit/blob/dac7dc6beda3915fbca83dc09bfc1aee721a440d/src/utils/logger.js#L75) `[GBK]` as a consistent identifier to all messages. Does it make sense to do so here or is that counter to iOS/Xcode practices?

dcalhoun (commented):
Thank you for creating this! 

This makes sense to me a high level. It feels like a worthwhile improvement overall, especially given the performance improvements. 

I left a few thoughts to consider. Many are for bettering my own understanding.

ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
kean: I suggest to add it again in the future if it's needed. For now, I kept it simple as `EditorService` knows what endpoints to use, and the code in the app was duplicating the same endpoint.

ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
kean: Yes, I still inject the same log level in JS: `logLevel: '\(EditorLogger.logLevel.rawValue)',`

ios/Sources/GutenbergKit/Sources/EditorLogging.swift
kean: >  also log directly from the iOS code, correct?

Yes, it's for logging directly from Swift. It look like there is no way around duplicating it. I think it's fine.

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
kean: > Any performance implications from adding this? 🤔 I suppose it is no larger than editorSettings, eh?

I wouldn't expect it to be slower than the current implementation, but I already had some second thoughts yesterday. Is it safe to pass these larger strings directly in JS this way? Does it need escaping? Is there a more robust way to do it? Should we pass the whole pre-processed manifest as a separate object? Both settings and manifest seem to have `<scripts> ` and `<styles>`, is there a better way to inject them? What about `allowedBlockTypes`? Should the app combine them (for core and plugins)?

ios/Sources/GutenbergKit/Sources/Service/EditorService.swift
kean: When you open the app, the editor (`EditorService`) now refreshes all these resources as part of `warmup`. This code makes sure that if you open the editor shortly after opening the app, we don't refresh it twice in a row.

ios/Sources/GutenbergKit/Sources/EditorLogging.swift
kean: I'll add it in the logger implementation in the host app. For `OSLog` in the demo, there is no need for a prefix as it has a "subsystem" name that you can display next to the individual log messages and filter messages by.
Screenshot 2025-11-21 at 10 08 54 AM

ios/Demo-iOS/Sources/Views/AppRootView.swift
kean: Oh, I added it as part of another changes I later reverted. I'll revert this as well.

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
kean: Not without a delay, but I refactored it now, so these lines are no longer needed.

kean:
I made a few more changes:

- Added unit tests and fixed a couple of issues (it explains most of the line additions)
- Updated EditorService to store processed manifest on disk (it takes around 100ms to generate it)
- Added support for cleaning up orphaned asset files, so we don’t take too much space on disk

With these changes, this whole step now takes 2ms.

I had to switch to a different task today and didn't have time to review the final PR. I'll do it when I have time. It should be good.

ios/Demo-iOS/Sources/Views/AppRootView.swift
kean: Reverted

src/utils/bridge.js
kean: I reverted this change together with how manifest is passed to the editor (now it again uses separate post message scheme).

kean:
I updated the unit tests and added a few more. It should be good for review and merge now.

ios/Demo-iOS/Sources/Views/AppRootView.swift
```diff
@@ -40,9 +40,13 @@ struct AppRootView: View {
         }
         .onChange(of: self.selectedConfiguration) { oldValue, newValue in
             switch newValue {
-            case .bundledEditor: activeEditorConfiguration = createBundledConfiguration()
-            case .editorConfiguration(let config): self.loadEditorConfiguration(for: config)
-            case .none: self.activeEditorConfiguration = nil
+            case .bundledEditor:
+                let config = createBundledConfiguration()
+                activeEditorConfiguration = config
```
dcalhoun, line 45: Likely worthwhile consistently using `self` in this `switch`.
```suggestion
                self.activeEditorConfiguration = config
```

ios/Demo-iOS/Sources/Views/EditorView.swift
```diff
@@ -202,6 +198,9 @@ private final class EditorViewModel {
 
 #Preview {
     NavigationStack {
-        EditorView(configuration: .default)
+        let config = EditorConfigurationBuilder()
+            .setSiteUrl("preview")
```
dcalhoun: Should we revert this given the `"bundled"` change was reverted?

ios/Sources/GutenbergKit/Sources/Service/EditorService.swift
```diff
@@ -81,34 +198,289 @@ public actor EditorService {
...
+    /// - Returns: The local file URL where the asset is stored
+    private func fetchAsset(from urlString: String) async throws -> URL {
+        guard let url = URL(string: urlString) else {
+            throw URLError(.badURL)
+        }
+
+        let localURL = assetsDirectoryURL.appendingPathComponent(cachedFilename(for: urlString))
+
+        if FileManager.default.fileExists(atPath: localURL.path) {
+            log(.debug, "Found cached asset for: \(url.lastPathComponent) (\(localURL.fileSize.formatted))")
```
dcalhoun: This seems to always log `0 KB` for each file in my testing. Is that expected? 

This also occurs with `Cleaned up 2 orphaned assets (Zero KB)`. 

Image

ios/Demo-iOS/Sources/Services/.swift
dcalhoun, line 1: This file has an empty file name. Is that intentional?

ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
dcalhoun: Actually, I believe we must retain this to support WPCOM sites. They require setting a namespace endpoint—e.g., `https://public-api.wordpress.com/wpcom/v2/sites/<site_id>/editor-assets`. 

From my testing, plugin loading currently fails for WPCOM Simple sites.

dcalhoun (commented):
Overall, this tested well for me in the demo app. 

As stated in a new inline comment, I believe we must retain the `editorAssetsEndpoint` configuration. I also noted a few oddities likely worth addressing. 

This may be understood, but we need to open a sibling WordPress-iOS PR integrating these changes.

src/utils/bridge.js
dcalhoun: Would you please explain the rationale for reverting for my own education? 🙇🏻‍♂️

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
dcalhoun: I agree, my original performance concerns are likely invalid; I figured I'd share my thought nonetheless. 

I do not have a better idea for injecting them at this time. 

Regarding escaping, maybe there is benefit. That said, if the itself relies upon the same scripts we are retrieving, would site already be compromised? Is the mobile editor exposing anything that site does not already? 

For `allowedBlockTypes`, I envision this eventually becoming an editor configuration option rather than a value returned from the editor assets endpoint. I consider its current location a stopgap solution while various blocks do not function in the mobile editor.

ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
kean: Good catch. So far, I've only tested it with an Atomic site as the demo doesn't seem to support Simple sites:

Screenshot 2025-12-02 at 9 46 59 AM

I'll probably revert it for now, but there has to be a better way to do it. I'm not completely happy with how the configuration is spread out between GBK and the app. I'm leaning towards centralizing this knowledge in GBK so it would work in both the app and the demo. The [configuration code](https://github.com/wordpress-mobile/WordPress-iOS/blob/4bc758160f4f26bdc9164caad3e262faf9d59c23/WordPress/Classes/Utility/Editor/EditorConfiguration%2BBlog.swift#L10) is also a bit hard to follow at the moment, but it seems like it's better be left for future PRs.

dcalhoun: Correct, the demo app does not support WPCOM sites. I tested via integrating with the WordPress-iOS app. 

I open to refactoring the API for configuring and starting the editor. 

As much as possible, we should avoid WPCOM/Jetpack references in the GutenbergKit code base. There are few that exist today, but they need to be removed in the future.

ios/Demo-iOS/Sources/Services/.swift
kean: Removed

ios/Sources/GutenbergKit/Sources/Service/EditorService.swift
kean: Fixed (removed these for now)

ios/Demo-iOS/Sources/Views/EditorView.swift
kean: Reverted

jkmassel (approved):
After some discussion with @dcalhoun and @kean, I'm approving this to merge and we can deal with breakage later. This and #232 represent some pretty serious refactoring, so we should get them landed then incrementally adjust after.

PR #241, opened 2025-11-27
XCFramework experiment

mokagio:
Demo for building the package as an XCFramework, with CI bare bone integration tests.

image

---


.buildkite/commands/verify-xcframework-test-fails-with-missing-resources.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -eu
+
+set -o pipefail
+
+echo "--- :pencil2: Breaking bundle resource path to verify test catches failures"
+# Replace "Gutenberg" with "folder_that_does_not_exist" in the HTMLPreviewManager
+sed -i.bak 's/forResource: "Gutenberg", withExtension: nil/forResource: "folder_that_does_not_exist", withExtension: nil/g' ios/Sources/GutenbergKit/Sources/Views/HTMLPreview/HTMLPreviewManager.swift
```
mokagio, line 8: Since we can't inject values that would result in a crash _and_ [we can't test for crashes anyways when targeting iOS](https://github.com/swiftlang/swift-evolution/blob/d2111b0f8e725ffa68d124e9570cccb20bde194c/proposals/testing/0008-exit-tests.md#:~:text=If%20a%20platform%20does%20not%20support%20exit%20tests%20(generally%20because%20it%20does%20not%20support%20spawning%20or%20awaiting%20child%20processes)%2C%20then%20we%20define%20SWT_NO_EXIT_TESTS%20when%20we%20build%20it.) we hack the code to force a crash and test that the tests fail as expected.

.buildkite/commands/verify-xcframework-test-fails-with-missing-resources.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -eu
+
+set -o pipefail
+
+echo "--- :pencil2: Breaking bundle resource path to verify test catches failures"
+# Replace "Gutenberg" with "folder_that_does_not_exist" in the HTMLPreviewManager
```
mokagio:
```suggestion
# Replace "Gutenberg" with "folder_that_does_not_exist" in the HTMLPreviewManager to trigger `assert`.
# (`assert` works because we build the XCFramework with -Onone in these tests)
```

mokagio:
Closing in favor of #318

PR #245, opened 2025-12-03
Add editorSettingsEndpoint to EditorConfiguration

kean:
## What?
- Fix an issue with editor settings not loading for Simple sites. It follows the same approach as used for `editorAssetsEndpoint`. I intentionally didn't do any refactoring.
- Update `EditorService` to add query parameters to both the custom and the default endpoints for asset manifest (to keep the behavior consistent with the production version of this API):

```
URLQueryItem(name: "exclude", value: "core,gutenberg")
```

## Testing Instructions
- (Optional) Use draft integration PR https://github.com/wordpress-mobile/WordPress-iOS/pull/25044
- Open the site and the editor
- **Verify** that the editor settings are loaded, e.g. styles are applied
- **Verify** that the request URL to fetch settings is constructed correctly, e.g. https://public-api.wordpress.com/wp-block-editor/v1/sites/228225032/settings

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorConfiguration.swift
@@ -32,6 +32,8 @@ public struct EditorConfiguration: Sendable {
     public let locale: String
     /// Enables the native inserter UI in the editor
     public let isNativeInserterEnabled: Bool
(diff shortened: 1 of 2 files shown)


dcalhoun (approved):
Tested well for me. 🚀

PR #246, opened 2025-12-03
Disable cookies in EditorService

kean:
## What?

Fix an issue with authentification cookies taking priority over application passwords that are required for the editor  (similar to https://github.com/wordpress-mobile/GutenbergKit/commit/30ebac210924ecc8e9dee3980c101ef24b1befa6).

## Testing Instructions

I tested it by creating a temporary integration branch and verifying that my JN site starts returning 200.

Screenshot 2025-12-03 at 10 33 44 AM


dcalhoun (approved):
I was unable to reproduce the original issue, but this makes sense to me.

PR #248, opened 2025-12-04
Fix compilation on macos

jkmassel:
## What?
Implements minimum viable macOS compilation. This doesn't mean we actually support macOS, it just means the code compiles there.

## Why?
Enables running `swift test`. Not having it while trying to do TDD for the logic parts of this library is a headache and doesn't work well with `claude`.

## How?
A `CrossPlatform.swift` file that papers over the platform-specific differences.

## Testing Instructions
Ensure CI passes.


ios/Sources/GutenbergKit/Sources/Views/HTMLPreview/HTMLWebViewRenderer.swift
```diff
@@ -35,7 +35,7 @@ final class HTMLWebViewRenderer {
 
         // Take screenshot of entire content with the maximum target size we'll ever need (roughly)
         // Calculate snapshot width such that neither dimension exceeds the maximum size
-        let maxDimension = min(UIScreen.main.bounds.width, UIScreen.main.bounds.height, 640) - 48
```
jkmassel, line 38: We probably shouldn't do this anyway given that we need to adopt `UIScene` soon?

`640` seems like a pretty reasonable size  – if anyone feels strongly we could inject this value into the function from the view layer?

kean: It is going to be a pretty significant waste of resources as 640 is nearly double of 375 on iPhones. I'd suggest to keep it for now and add make it conditional.

jkmassel: Done in 832b133ce3d19c1fc57d373ffb86aef8b0b73480

kean (approved):

PR #252, opened 2025-12-08
Update to Swift 6

jkmassel:
## What?
What it says in the title

## Why?
Concurrency safety, and #250

## How?
Just a few little changes.

## Testing Instructions
If this compiles, it should be fine.

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/Views/BlockInserter/BlockInserterViewModel.swift
+++ b/ios/Sources/GutenbergKit/Sources/Views/BlockInserter/BlockInserterViewModel.swift
@@ -61,17 +61,14 @@ class BlockInserterViewModel: ObservableObject {
         let task = Task<[MediaInfo], Never> { @MainActor in
             var results: [MediaInfo] = []
             var anyError: Error?
-            await withTaskGroup(of: Void.self) { group in
+
+            do {
                 for item in items {
-                    group.addTask {
-                        do {
-                            let item = try await self.fileManager.import(item)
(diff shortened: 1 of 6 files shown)


ios/Sources/GutenbergKit/Sources/Views/BlockInserter/BlockInserterViewModel.swift
```diff
@@ -61,17 +61,14 @@ class BlockInserterViewModel: ObservableObject {
         let task = Task<[MediaInfo], Never> { @MainActor in
             var results: [MediaInfo] = []
             var anyError: Error?
-            await withTaskGroup(of: Void.self) { group in
+
+            do {
                 for item in items {
-                    group.addTask {
-                        do {
-                            let item = try await self.fileManager.import(item)
```
jkmassel, line 68: Multiple operations like this are competing for many of the same resources (network and Disk IO) so I'm not sure parallelizing them here is much of a win.

If you don't love this I can try something new, this was just a quick fix.

kean: I feel like I generated this code I have no memory of it. I do not like unbounded parallelization for sure.

For file operations, it's largely irrelevant. I'm not sure if there may be scenarios where it needs to trigger download of the original data from the cloud. In that case, you probably want to trigger them all at once and Apple will presumably take care of the networking and limiting the requests.

kean (approved):

PR #253, opened 2025-12-09
Add iOS Preload Tests

jkmassel:
Just the tests for #250, split out because there's a lot of code.


ios/Tests/GutenbergKitTests/Resources/editor-asset-manifest-test-case-1.json
dcalhoun, line 1: FYI this file and the editor settings file include personal information—e.g., name and email address.

Makefile
```diff
@@ -143,6 +143,10 @@ test-js-watch: npm-dependencies ## Run JavaScript tests in watch mode
 test-swift-package: build ## Run Swift package tests
 	$(call XCODEBUILD_CMD, test)
 
+.PHONY: test-swift-package-logic
+test-swift-package-logic:
+	$(call XCODEBUILD_LOGIC_CMD, test)
```
dcalhoun: `XCODEBUILD_LOGIC_CMD` appears undefined. 

```
❯ make test-swift-package-logic
make: Nothing to be done for `test-swift-package-logic'.
```

dcalhoun (commented):
Thank you for the thorough test coverage! 

I skimmed the test names, and a few dozen test implementations. The tests look good to me. 

Some of the more granular tests seem less valuable (e.g., Foundation extensions). Hopefully their long-term value exceeds any maintenance cost—presumably we won't change these much and generative AI should make changes easier. I'm not opposed to keeping them, but we could consider reduction or consolidation if deemed worthwhile.

ios/Tests/GutenbergKitTests/Resources/editor-asset-manifest-test-case-1.json
jkmassel: Addressed, thanks!!

Makefile
jkmassel: Addressed in 209696f39a21ee122972d9f585d607cb1949dab9 (may change if #250 is further adjusted)

Thanks!

dcalhoun:
I rebased atop the latest trunk branch with the following command to make review easier:

```bash
git rebase --onto origin/trunk origin/add/preload-list add/preload-list-tests
```

CLAUDE.md
```diff
@@ -98,13 +98,10 @@ npm run clean:l10n    # Clean only translations directory
 
 ```bash
 # Build Swift package
-make build-swift-package
-
-# Build Swift package (force refresh of npm deps/translations if needed)
-make build-swift-package REFRESH_DEPS=1 REFRESH_L10N=1
+swift build
```
dcalhoun, line 101: Is the usage of `swift` over `make` to avoid required simulators or some other purpose?

dcalhoun (approved):

CLAUDE.md
jkmassel: The `make` stuff is used in CI, which builds the JS. Claude kept running it unnecessarily which was slow, and also polluted my git repo with the JS changes

dcalhoun: @jkmassel makes sense. 

I've been addressing similar inefficiencies elsewhere by introducing `REFRESH_[target]` environment variables for the [`npm-dependencies`](https://github.com/wordpress-mobile/GutenbergKit/blob/704fbbd3200f6c91a82f8973276dab1854e5cd1b/Makefile#L30-L39) and [`prep-translations`](https://github.com/wordpress-mobile/GutenbergKit/blob/704fbbd3200f6c91a82f8973276dab1854e5cd1b/Makefile#L43-L59) targets. We might consider adding a similar `REFRESH_JS_BUILD` here. 

My goal has been to structure the `Makefile` to be the de facto interface for the project, where all documentation for users and LLMs point to the same scripts. Where the `Makefile`... 

1. Always ensures the correct dependencies are in place for a given target;
2. Delivers efficient re-runs (for local runs);
3. And provides necessary environment variables for full control when necessary (for CI servers or resets). 

Not stating we must revert these changes, just offering an alternative for consideration.

jkmassel: I'm not 100% sure I understand, but I agree with it in principle so I'd be happy to 👍 a PR that does this?

dcalhoun:
@jkmassel it seems the automated Swift tests fail in trunk after this merge. Strangely, the tests seemingly pass for PR runs.

PR #260, opened 2025-12-16
Add Android Preload List

jkmassel:
## What?

Adds an Android preloading system for the Gutenberg editor that prefetches editor dependencies before display, matching the existing iOS implementation.

## Why?

The Gutenberg editor requires several API calls and asset downloads during initialization, which can cause slow launch times. This PR introduces a preloading architecture that:

- Pre-fetches API responses (post types, themes, settings) so the editor doesn't need to make network calls during startup
- Downloads and caches JavaScript/CSS asset bundles for faster subsequent launches
- Provides progress feedback to the user during preparation
- Establishes feature parity with the existing iOS preloading system

## How?

**New Core Components:**
- `EditorService` - Orchestrates loading all editor dependencies with progress tracking
- `EditorPreloadList` - Pre-fetched API responses injected into the editor JavaScript
- `EditorAssetsLibrary` (in stores/) - Downloads and caches asset bundles
- `EditorHTTPClient` - Authenticated HTTP client for WordPress REST API
- `RESTAPIRepository` - REST API calls with response caching
- `EditorURLCache` - URL-level caching with configurable policies

**New Model Classes:**
- `EditorAssetBundle` - Represents downloaded editor assets
- `EditorCachePolicy` - Configurable caching strategies (Always, MaxAge, Ignore)
- `EditorSettings`, `EditorDependencies` - Editor configuration models
- `GBKitGlobal`, `LocalEditorAssetManifest`, `RemoteEditorAssetManifest` - Asset manifest parsing

**Demo App Updates:**
- Added `SitePreparationActivity` and `SitePreparationViewModel` showing the preloading flow with progress UI

**Other Changes:**
- Minor iOS test fixes and `EditorSettings` improvements for consistency
- This PR removes the existing warmup code – [a more recent version of `androidx.webkit` exposes `prerenderUrlAsync` ](https://developer.android.com/reference/androidx/webkit/WebViewCompat#prerenderUrlAsync(android.webkit.WebView,java.lang.String,android.os.CancellationSignal,java.util.concurrent.Executor,androidx.webkit.PrerenderOperationCallback)) which we should adopt for this purpose.

## Testing Instructions

1. Build and run the Android demo app
2. Select a site configuration
3. Observe the "Site Preparation" screen showing loading progress
4. Tap "Prepare from scratch" to test full loading without cache
5. Tap "Prepare" to test cached loading
6. Verify the editor loads correctly with preloaded data
7. Run `make test-android` to verify all new unit tests pass

See the tests in #250 for examples of tests sites you can use.


android/app/src/main/java/com/example/gutenbergkit/SitePreparationActivity.kt
```diff
@@ -0,0 +1,553 @@
...
+        enableEdgeToEdge()
+
+        val configurationItem = parseConfigurationItem(
+            intent.getStringExtra(EXTRA_CONFIGURATION_ITEM)
+        )
+
+        val viewModel = ViewModelProvider(
+            this,
+            SitePreparationViewModelFactory(application, configurationItem)
+        )[SitePreparationViewModel::class.java]
```
adalpari, line 131: ⛏️Aren't we using injection in the example project? ViewModels can be simply injected.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorDependenciesSerializer.kt
```diff
@@ -0,0 +1,109 @@
...
+        return file.absolutePath
+    }
+
+    /**
+     * Reads [EditorDependencies] from a file and deletes the file.
+     *
+     * @param filePath The absolute path to the serialized dependencies file.
+     * @return The deserialized dependencies, or `null` if the file doesn't exist or is invalid.
+     */
+    fun readFromDisk(filePath: String?): EditorDependencies? {
```
adalpari: ⛏️ kotlin nitpick: Do we need to accept null for `filePath`? I would avoid nullability when possible.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorSettings.kt
```diff
@@ -0,0 +1,97 @@
...
+         * @param data The raw JSON string from the block editor settings endpoint.
+         */
+        fun fromData(data: String): EditorSettings {
+            val jsonValue = try {
+                json.parseToJsonElement(data)
+            } catch (e: Exception) {
+                null
+            }
+
+            val stringValue = jsonValue?.toString() ?: ""
```
adalpari:
```suggestion
            val stringValue = jsonValue?.toString().orEmpty()
```

android/Gutenberg/src/main/java/org/wordpress/gutenberg/stores/EditorAssetsLibrary.kt
```diff
@@ -0,0 +1,399 @@
...
+            } catch (e: Exception) {
+                Log.w(TAG, "Failed to background cache: $url", e)
+            }
+        }
+    }
+
+    /**
+     * Shuts down background operations.
+     */
+    fun shutdown() {
```
adalpari: I don't see this function being called. So, creating a `CoroutineScope` as a variable could cause leaks because of no lifecycle awareness..
❓ What about using the ´ApplicationScope´, ´ViewModelScope´, or similar?

android/Gutenberg/src/main/java/org/wordpress/gutenberg/stores/EditorURLCache.kt
```diff
@@ -0,0 +1,180 @@
...
+            if (!cachePolicy.allowsResponseWith(storageDate, currentDate)) {
+                return null
+            }
+
+            EditorURLResponse(
+                data = entry.data,
+                responseHeaders = EditorHTTPHeaders(entry.headers)
+            )
+        } catch (e: Exception) {
+            null
```
adalpari: Do we want to track the exception here?

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -701,14 +759,15 @@ class GutenbergView : WebView {
 
     override fun onDetachedFromWindow() {
         super.onDetachedFromWindow()
+        coroutineScope.cancel()
```
adalpari: ⚠️  Good approach since we are cancelling the scope when detached from Windows. But if the view is reattached again because of other scenarios (I'm not aware about all the scenarios), the `coroutineScope` is dead forever.
There are a few lifecycle aware CoroutineScopes we can use on Android. This class is a tricky one, but maybe you could try using:
```
findViewTreeLifecycleOwner()?.lifecycleScope.run {
    ...
}
```

android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorDependenciesSerializer.kt
jkmassel: Addressed in 8060669f6b495b2cbdd6093b42a55a4b6ca12f2d thanks!

android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorSettings.kt
jkmassel: Nice thanks!!

android/Gutenberg/src/main/java/org/wordpress/gutenberg/stores/EditorAssetsLibrary.kt
jkmassel: I've adjusted the library code to accept an injected lifecycle so that the host app can decide how this should work in 8060669f6b495b2cbdd6093b42a55a4b6ca12f2d. 

Thanks!!

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
jkmassel: Addressed as part of the scope injection in 8060669f6b495b2cbdd6093b42a55a4b6ca12f2d

android/Gutenberg/src/main/java/org/wordpress/gutenberg/stores/EditorURLCache.kt
jkmassel: In this case probably not, because we'll treat a cache failure as a simple cache miss. I've modified this to use `runCatching` in c64592b03127aadac6ae5c119733c156112759ec, which should hopefully make the intent more clear!

android/app/src/main/java/com/example/gutenbergkit/SitePreparationActivity.kt
jkmassel: I don't think we are at the moment – I'd be happy to adopt it, but I think it's outside the scope of this PR :)

dcalhoun:
@claude

android/app/src/main/java/com/example/gutenbergkit/SitePreparationActivity.kt
```diff
@@ -0,0 +1,553 @@
...
+        modifier = modifier.fillMaxSize(),
+        contentPadding = PaddingValues(16.dp),
+        verticalArrangement = Arrangement.spacedBy(16.dp)
+    ) {
+        // Dependencies Status
+        item {
+            DependenciesStatusCard(
+                hasDependencies = uiState.editorDependencies != null
+            )
+        }
```
dcalhoun, line 250: This doesn't appear to correctly denote the presence of caches created from loading the editor (i.e., without using the dedicated "prepare" buttons). It also does not recognize old caches—navigating here, preparing a cache, navigating back to the root view, and re-navigating here results in an inaccurate "editor dependencies missing." 

Additionally, in the latter scenario, the loading view incorrect displays "Loading Editor" rather than "Starting Editor." The progress bar displayed also never updates after initial render in this context. 

<details><summary>Unrecognized preload caches recording</summary>
<p>


https://github.com/user-attachments/assets/eb8784ec-06c2-4c5a-849c-7c745fec37e1



</p>
</details>

android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorConfiguration.kt
```diff
@@ -23,23 +26,40 @@ open class EditorConfiguration constructor(
...
+        fun builder(siteURL: String, siteApiRoot: String, postType: String = "post"): Builder = Builder(siteURL, siteApiRoot, postType = postType)
+
+        @JvmStatic
+        fun bundled(): EditorConfiguration = Builder("https://example.com", "https://example.com/wp-json/", "post")
+            .setEnableOfflineMode(true)
+            .build()
     }
 
-    class Builder {
+    class Builder(private var siteURL: String, private var siteApiRoot: String, private var postType: String) {
```
dcalhoun, line 56: I perceive this makes these values required by the editor configuration. The optional values were useful for using the block editor for non-site-specific contexts in the future—e.g., editing comment text in the Jetpack app, editing entries in the Day One app. However, I perceive this can still be accomplished with something like the following. 👍🏻 

```kotlin
val config = EditorConfiguration.bundled()
      .toBuilder()
      .setContent("my content here")
      .build()
```

android/Gutenberg/src/main/java/org/wordpress/gutenberg/RESTAPIRepository.kt
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * Fetches post data for the given post ID.
+     *
+     * @param id The post ID to fetch.
+     * @return The response containing the post data.
+     */
+    suspend fun fetchPost(id: Int): EditorURLResponse {
+        val url = buildPostUrl(id)
+        val response = httpClient.perform("GET", url)
```
dcalhoun: Why might this and other similar lines use a string `"GET"` rather than `EditorHttpMethod.GET`? This doesn't appear to use the internal helper—should it?

android/Gutenberg/src/main/java/org/wordpress/gutenberg/views/EditorProgressView.kt
```diff
@@ -0,0 +1,91 @@
...
+
+        // Create progress bar
+        progressBar = ProgressBar(context, null, android.R.attr.progressBarStyleHorizontal).apply {
+            layoutParams = LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.WRAP_CONTENT).apply {
+                marginStart = dpToPx(16)
+                marginEnd = dpToPx(16)
+            }
+            max = 100
+            progress = 0
+        }
```
dcalhoun, line 56: There is an unexpected artifact on the far right of the progress bar (a detached, filled dot at the 100% mark). It's present when the bar first renders, and while it progressively fills. Not sure why this occurs. 

<details><summary>Progress bar artifact screenshot</summary>

Image

</details>

dcalhoun (commented):
This look really sound—nicely done! 

I spent most of my time reviewing the logic itself, and little time on the automated tests. 

I tested the functionality in the demo app. Testing went well. The performance improvement is notable on sites like `slow.wpmt.co`. For other sites, there a notable delay remains during editor start (after preloading) that presumably stems from time spent loading editor/plugin assets. My understanding is that we might further optimize those in the future with things like referenced `prerenderUrlAsync`. 

I feel Claude's review is accurate, and the inline comments are likely worth considering. Outside of that, I noted a few oddities in my own inline comments to consider. None of these appear as blockers, some only impact the demo app. 

@jkmassel let me know what you think about Claude's and my review and if you'd like for me to approve as-is. 

---

An aside: with the editor settings (background color) now applied to the Android demo app, I noted the editor height is unexpectedly cut off. I addressed this in #275.

adalpari (approved):
Apart from the opened comments (@dcalhoun and Claude) the PR looks good to me. I don't think there are any blocker. So, I am unblocking this PR as it's big (and complex) enough to consider it done. 
I would suggest to achieve the suggestions (if you consider it necessary) in a new iteration. To avoid iterating over this big PR.

Great changes! 👏 

PS: I see there's a conflict that must be resolved though.

nbradbury:
@jkmassel I noticed a couple of things which may or may not be a problem. First, I tried to add a Jurassic Ninja site that doesn't have the Gutenberg plugin, and I got this error when starting the editor:

no-route

I'm not sure if this error would occur in a production app, but if so we could stand to add some padding to it.

Next, I tried to add a wp.com simple site that **does** have Gutenberg, and I see the "Discovering site" message but it never gets added.

https://github.com/user-attachments/assets/e613cce8-08fb-46e5-b333-bf34938848dd

Also, I realize this is just a demo app, but it would still be nice to have feedback (like a "Cache cleared" toast) when tapping "Clear preload cache." When I clicked it I assumed it didn't do anything, but looking at the code I see it does.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorConfiguration.kt
jkmassel: Yep that's the idea behind `.bundled` – it's for using the editor without an associated website.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/views/EditorProgressView.kt
jkmassel: I saw this too and assumed it was something Android-specific. @adalpari or @nbradbury  can you weigh in on this?

android/Gutenberg/src/main/java/org/wordpress/gutenberg/RESTAPIRepository.kt
jkmassel: Good call – addressed in 80aa38bf9ff1f68d3bc9f8d5b7df9b79918990ad

jkmassel:
@nbradbury's issue is fixed for iOS in #288.

This PR is already pretty huge, so I'll open a follow-up to address it there.

dcalhoun:
> @nbradbury's issue is fixed for iOS in #288.
> 
> This PR is already pretty huge, so I'll open a follow-up to address it there.

To clarify, I believe the Demo app does not support OAuth tokens at all, so WPCOM Simple sites will not work there. They must be tested in the WordPress/Jetpack app.

For the Gutenberg `rest_no_route` issue, this seems like CMM-1163-linear-issue, which is a regression we need to address for both iOS and Android.

PR #261, opened 2025-12-18
Don’t send cookies with requests

jkmassel:
## What?
Applies the fix in #246 to `EditorHttpClient` and adds tests to it.

## Why?
We don't want a regression

## Testing Instructions
The CI tests covers this.

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorHTTPClient.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorHTTPClient.swift
@@ -11,7 +11,12 @@ public protocol EditorHTTPClientProtocol: Sendable {
 ///
 /// Implement this protocol to inspect or log all network requests.
 public protocol EditorHTTPClientDelegate {
-    func didPerformRequest(_ request: URLRequest, response: URLResponse, data: Data)
+    func didPerformRequest(_ request: URLRequest, response: URLResponse, data: EditorResponseData)
+}
+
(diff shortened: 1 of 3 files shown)


ios/Sources/GutenbergKit/Sources/EditorHTTPClient.swift
```diff
@@ -35,17 +40,17 @@ public actor EditorHTTPClient: EditorHTTPClientProtocol {
...
-    private let requestTimeout: TimeInterval
+    private let requestTimeout: TimeInterval?
 
     public init(
-        urlSession: URLSession,
+        urlSession: URLSessionProtocol,
         authHeader: String,
         delegate: EditorHTTPClientDelegate? = nil,
-        requestTimeout: TimeInterval = 60 // `URLRequest` default
+        requestTimeout: TimeInterval? = nil
```
dcalhoun, line 53: What is the rationale for changing this default?

dcalhoun (approved):

ios/Sources/GutenbergKit/Sources/EditorHTTPClient.swift
jkmassel: @kean had noted that we were re-stating the default, which isn't especially useful. Making it optional defaults to what the framework does, but in a way that if it ever changes we just match it.

PR #275, opened 2026-01-06
fix: Android demo app fills viewport

dcalhoun:
## What?

Ensure the editor HTML expands to fill the viewport height.

## Why?

The editor background is clipped if the root element does not expand to fill the entire height. 

## How?

Set explicit `layoutParams` with MATCH_PARENT on `GutenbergView` when used inside Compose's `AndroidView`. This fixes CSS viewport units (`vh`, `vw`, `dvh`) calculating as zero.

When a WebView is embedded in `AndroidView` without explicit `layoutParams`, the WebView's initial size is calculated as zero during the first composition pass. This causes CSS viewport units to resolve to 0px, even though JavaScript APIs like `window.innerHeight` return correct values.

Setting `MATCH_PARENT` `layoutParams` ensures the WebView knows its intended size constraints before loading content, allowing CSS viewport calculations to work correctly.

This is a known issue with WebView in Jetpack Compose:
- https://issuetracker.google.com/issues/238245846
- https://github.com/google/accompanist/issues/1224

## Testing Instructions

This is particularly notable if a theme's styles are applied to the editor via #260. 

<details><summary>Example background clipping</summary>
<p>

gbk-android-body-background

</p>
</details> 

To verify the fix manually, rely upon the Chrome DevTools to inspect the element height: 

1. Open the demo app editor on Android. 
2. Inspect the WebView via the Chrome DevTools. 
3. Verify the `html` element does not have a height of `0`, but fills the entire viewport. 

N/A, no navigation changes. 

## Screenshots or screencast 
Before | After 
-- | --
before | after

### DIFF
--- a/android/app/src/main/java/com/example/gutenbergkit/EditorActivity.kt
+++ b/android/app/src/main/java/com/example/gutenbergkit/EditorActivity.kt
@@ -2,6 +2,7 @@ package com.example.gutenbergkit
 
 import android.content.Intent
 import android.os.Bundle
+import android.view.ViewGroup
 import android.webkit.WebView
 import android.content.pm.ApplicationInfo
 import androidx.activity.ComponentActivity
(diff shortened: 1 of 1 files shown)


nbradbury (approved):
Looks good! :shipit:

PR #292, opened 2026-01-23
fix: Plugin loading respects `plugins` configuration

dcalhoun:
## What?

Update the `EditorService` to respect the `plugins` configuration option. The `EditorService` on both iOS and Android was unconditionally calling `prepareAssetBundle()` regardless of the `plugins` configuration setting. This caused plugin assets to always be loaded even when plugins were disabled.

## Why?

Fix CMM-1173. 

The previous behavior of ignoring the `plugins` configuration effectively disabled the configuration option's ability to control the user experience. This is problematic if a host app needs to selectively disable plugin support. 

## How?

Added `prepareAssetBundleIfEnabled()` method that checks the configuration before attempting to load plugin assets:
- iOS: Checks `configuration.shouldUsePlugins`
- Android: Checks `configuration.plugins`

When plugins are disabled, the method returns an empty asset bundle immediately instead of attempting to download or read from cache.

## Testing Instructions

1. Enable the `plugins` configuration. 
2. Load the editor a compatible site so that an asset bundle is created. 
3. Verify plugin blocks are loaded. 
4. Close the editor. 
5. Disable the `plugins` configuration. 
6. Open the editor for the site. 
7. Verify plugin blocks are not loaded. 

N/A, no UX changes. 

## Screenshots or screencast 
N/A, no UX changes.

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/services/EditorService.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/services/EditorService.kt
@@ -201,7 +201,7 @@ class EditorService(
 
         return coroutineScope {
             val settingsDeferred = async { prepareEditorSettings() }
-            val assetBundleDeferred = async { prepareAssetBundle() }
+            val assetBundleDeferred = async { prepareAssetBundleIfEnabled() }
             val preloadListDeferred = async { preparePreloadList() }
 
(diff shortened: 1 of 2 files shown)


dcalhoun:
I can no longer reproduce this. Closing until I am able to reproduce the problem. 🙃

PR #293, opened 2026-01-26
fix: Skip post preloading for non-positive post IDs

dcalhoun:
## What?
Skips preloading post data when the post ID is non-positive (e.g., -1 or 0), which are sentinel values used by host apps for unsaved drafts.

## Why?
Fix CMM-1175.

WordPress-iOS uses `-1` as a sentinel value for new/unsaved draft post IDs. When the user refreshes My Site (pull-to-refresh), the editor's preload cache is invalidated. On the next editor open for a new post, GutenbergKit attempted to fetch `/posts/-1` from the API, which returned a 404 error and caused the editor to fail loading.

The preloading logic only checked for `null`/`nil` post IDs, not for non-positive values like `-1`.

## How?
Added an additional guard in `preparePreloadList()` on both iOS and Android to check that `postID > 0` before attempting to fetch post data:

**iOS** (`EditorService.swift:194`):
```swift
// Before:
if let postID = self.configuration.postID {

// After:
if let postID = self.configuration.postID, postID > 0 {
```

**Android** (`EditorService.kt:292`):
```kotlin
// Before:
if (postId != null) {

// After:
if (postId != null && postId > 0) {
```

Also changed the iOS `EditorService.init` to accept `(any EditorHTTPClientProtocol)?` instead of `EditorHTTPClient?` to enable dependency injection for testing.

## Testing Instructions
1. Open WordPress iOS app
2. Navigate to My Site
3. Pull-to-refresh on the My Site screen (this invalidates the editor cache)
4. Tap to create a new post
5. Verify the editor loads successfully without showing an error

### Unit Tests
New tests added for both platforms:
- `prepare does not fetch post when postID is negative` - verifies no request to `/posts/-1`
- `prepare does not fetch post when postID is zero` - verifies no request to `/posts/0`  
- `prepare fetches post when postID is positive` - verifies normal behavior for valid post IDs

Run tests:
```bash
# iOS
swift test --filter "EditorServiceTests"

# Android
./gradlew :Gutenberg:testDebugUnitTest --tests "org.wordpress.gutenberg.services.EditorServiceTest"
```

## Screenshots or screencast
N/A - Bug fix with no visual changes

### DIFF
--- a/ios/Demo-iOS/Gutenberg.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ios/Demo-iOS/Gutenberg.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,24 +1,6 @@
 {
-  "originHash" : "2aee256c0b38622aa9bb903b83d446aa51554bf7f6d1f43aca6d447f82d2a906",
+  "originHash" : "11fbb2cf922b873080bfec9421893617428fff2aba5645ca7121f53f5debada1",
   "pins" : [
-    {
-      "identity" : "svgview",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/exyte/SVGView.git",
-      "state" : {
-        "revision" : "6465962facdd25cb96eaebc35603afa2f15d2c0d",
-        "version" : "1.0.6"
-      }
-    },
-    {
-      "identity" : "swiftsoup",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/scinfu/SwiftSoup.git",
-      "state" : {
-        "revision" : "aa85ee96017a730031bafe411cde24a08a17a9c9",
-        "version" : "2.8.8"
-      }
-    },
     {
       "identity" : "wordpress-rs",
       "kind" : "remoteSourceControl",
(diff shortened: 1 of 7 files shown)


ios/Demo-iOS/Gutenberg.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
dcalhoun, line 1: These changes are unrelated. They occurred when opening the iOS demo app in Xcode.

ios/Tests/GutenbergKitTests/Stores/EditorAssetLibraryTests.swift
```diff
@@ -841,30 +841,44 @@ final class ProgressTracker: @unchecked Sendable {
...
     private let lock = NSLock()
 
+    /// URLs requested via `perform(_:)`. Use this to verify which endpoints were called.
+    private var _requestedURLs: [URL] = []
+    var requestedURLs: [URL] {
+        lock.withLock { _requestedURLs }
+    }
+
+    /// Handler for generating response data based on request URL. Defaults to returning empty data.
+    var urlResponseHandler: ((URL) -> Data) = { _ in Data() }
```
dcalhoun, line 858: Replaced the single-use `getResponse` mock with composable URL-based mock responses. Requires numerous test changes but allows for more nuanced mocks. E.g., `EditorService` triggers multiple disparate requests.

jkmassel (approved):
I double-checked, and `wp_posts` is declared `bigint(20) unsigned` so I don't think we'll ever need to worry about negative post IDs.

PR #299, opened 2026-01-27
Support editing posts of custom post types

crazytonyli:
There are two issues in loading custom posts from the editor
1. A request to `wp/v2/post/<id>` is sent out when preparing editor. This URL is hard-coded.
2. A JS exception is thrown when editing a custom post. The JS code contains a hard-coded list of "post entities", which appears to be the root cause.

This PR adds `restBase` and `restNamespace` to `EditorConfiguration`. The values will be used to fix the hard-coded `wp/v2/post` request. They are passed to the JS library and are added to the "post entities" list.

The second commit of this PR updates the Demo project to browse custom posts and allows opening them in the GBK editor. You can checkout this commit to reproduce the issue locally.


ios/Demo-iOS/Sources/Views/PostsListView.swift
```diff
@@ -0,0 +1,145 @@
...
+            postTypeDetails: postTypeDetails,
+            editorConfiguration: editorConfiguration,
+            editorDependencies: editorDependencies
+        )
+    }
+
+    var body: some View {
+        Group {
+            if viewModel.isLoading && viewModel.posts.isEmpty {
+                ProgressView("Loading Posts...")
```
dcalhoun: May not be worth the effort given this is a demo app, but we might changes this and other static "Posts" references to be `"\(viewModel.postTypeDetails.name) entries"` or simply `"entries"`.

ios/Demo-iOS/Sources/Views/PostsListView.swift
```diff
@@ -0,0 +1,145 @@
...
+                params: PostListParams(perPage: 20, status: [.custom("any")])
+            )
+
+            var loadedPosts: [AnyPostWithEditContext] = []
+            for try await page in sequence {
+                loadedPosts.append(contentsOf: page)
+            }
+            self.posts = loadedPosts
+        } catch {
+            self.error = error
```
dcalhoun, line 138: For a WP.com WoW/Atomic site, I was unable to load posts/pages and encountered the following error. 

<details><summary>Post loading error screenshot</summary>

Image

</details>

src/utils/bridge.js
```diff
@@ -301,6 +303,8 @@ export async function getPost() {
 		return {
 			id: post?.id ?? -1,
 			type: post?.type || 'post',
+			restBase: post?.restBase || 'posts',
+			restNamespace: post?.restNamespace || 'wp/v2',
```
dcalhoun, line 307: We need to update the JavaScript tests to address the current failures. Something like the following. 

<details><summary>Example diff</summary>

```diff
diff --git a/src/utils/bridge.test.js b/src/utils/bridge.test.js
index 6928581..5c2175b 100644
--- a/src/utils/bridge.test.js
+++ b/src/utils/bridge.test.js
@@ -236,6 +236,8 @@ describe( 'getPost', () => {
 			expect( result ).toEqual( {
 				id: 123,
 				type: 'page',
+				restBase: 'posts',
+				restNamespace: 'wp/v2',
 				status: 'draft',
 				title: { raw: 'Host Title' },
 				content: { raw: 'Host Content' },
@@ -299,6 +301,8 @@ describe( 'getPost', () => {
 			expect( result ).toEqual( {
 				id: 789,
 				type: 'post',
+				restBase: 'posts',
+				restNamespace: 'wp/v2',
 				status: 'draft',
 				title: { raw: 'GBKit Title' },
 				content: { raw: 'GBKit Content' },
@@ -322,6 +326,8 @@ describe( 'getPost', () => {
 			expect( result ).toEqual( {
 				id: 101,
 				type: 'page',
+				restBase: 'posts',
+				restNamespace: 'wp/v2',
 				status: 'publish',
 				title: { raw: 'Fallback Title' },
 				content: { raw: 'Fallback Content' },
@@ -338,6 +344,8 @@ describe( 'getPost', () => {
 			expect( result ).toEqual( {
 				id: -1,
 				type: 'post',
+				restBase: 'posts',
+				restNamespace: 'wp/v2',
 				status: 'draft',
 				title: { raw: '' },
 				content: { raw: '' },
@@ -364,6 +372,8 @@ describe( 'getPost', () => {
 			expect( result ).toEqual( {
 				id: -1,
 				type: 'post',
+				restBase: 'posts',
+				restNamespace: 'wp/v2',
 				status: 'draft',
 				title: { raw: 'Title' },
 				content: { raw: 'Content' },

```

</details>

ios/Sources/GutenbergKit/Sources/Model/EditorConfiguration.swift
```diff
@@ -177,7 +177,7 @@ public struct EditorConfigurationBuilder {
     title: String = "",
     content: String = "",
     postID: Int? = nil,
-    postType: String,
+    postType: PostTypeDetails,
```
dcalhoun, line 180: There are Swift test failures that seem to relate to this change and the one in `EditorPreloadList`. We need to update the automated tests. 

<details><summary>Example diff</summary>

```diff
diff --git a/ios/Tests/GutenbergKitTests/Model/GBKitGlobalTests.swift b/ios/Tests/GutenbergKitTests/Model/GBKitGlobalTests.swift
index e1fcd1a..e32a2d3 100644
--- a/ios/Tests/GutenbergKitTests/Model/GBKitGlobalTests.swift
+++ b/ios/Tests/GutenbergKitTests/Model/GBKitGlobalTests.swift
@@ -20,7 +20,7 @@ struct GBKitGlobalTests: MakesTestFixtures {
 
   private func makePreloadList() -> EditorPreloadList {
     EditorPreloadList(
-      postType: "post",
+      postType: .post,
       postTypeData: EditorURLResponse(data: Data(), responseHeaders: [:]),
       postTypesData: EditorURLResponse(data: Data(), responseHeaders: [:]),
       activeThemeData: EditorURLResponse(data: Data(), responseHeaders: [:]),
@@ -97,8 +97,8 @@ struct GBKitGlobalTests: MakesTestFixtures {
 
   @Test("maps postType to post.type")
   func mapsPostType() throws {
-    let postConfig = makeConfiguration(postType: "post")
-    let pageConfig = makeConfiguration(postType: "page")
+    let postConfig = makeConfiguration(postType: .post)
+    let pageConfig = makeConfiguration(postType: .page)
 
     let postGlobal = try GBKitGlobal(configuration: postConfig, dependencies: makeDependencies())
     let pageGlobal = try GBKitGlobal(configuration: pageConfig, dependencies: makeDependencies())
diff --git a/ios/Tests/GutenbergKitTests/Stores/EditorAssetLibraryTests.swift b/ios/Tests/GutenbergKitTests/Stores/EditorAssetLibraryTests.swift
index 0e354a4..eec81ee 100644
--- a/ios/Tests/GutenbergKitTests/Stores/EditorAssetLibraryTests.swift
+++ b/ios/Tests/GutenbergKitTests/Stores/EditorAssetLibraryTests.swift
@@ -9,7 +9,7 @@ struct EditorAssetLibraryTests {
 
     static var testConfiguration: EditorConfiguration {
         EditorConfigurationBuilder(
-            postType: "post",
+            postType: .post,
             siteURL: URL(string: "https://example.com")!,
             siteApiRoot: URL(string: "https://example.com/wp-json")!,
         )
@@ -20,7 +20,7 @@ struct EditorAssetLibraryTests {
 
     static var minimalConfiguration: EditorConfiguration {
         EditorConfigurationBuilder(
-            postType: "post",
+            postType: .post,
             siteURL: URL(string: "https://example.com")!,
             siteApiRoot: URL(string: "https://example.com/wp-json")!
         )

```

</details>

ios/Demo-iOS/Sources/Views/SitePreparationView.swift
```diff
@@ -51,10 +51,32 @@ struct SitePreparationView: View {
...
+                        Text("Post Type")
+                        Spacer()
+                        ProgressView()
+                    }
+                } else {
+                    Picker("Post Type", selection: $viewModel.selectedPostTypeDetails) {
+                        ForEach(viewModel.postTypes, id: \.self) { postType in
+                            Text(postType.name).tag(postType)
+                        }
+                    }
```
dcalhoun, line 67: This appears to be a pre-existing issue, but the Prepare Editor button does not take this selection into consideration. So, it always preloads the post type rather than the selected type. 

We might should fix this so that we can further validate the correct data is preloaded.

dcalhoun (commented):
This tested well for me. I noted the correct post type is provided to the WebView configuration. I was unable to verify the correct post type is loaded due to the issue with the Preload Editor button noted in an inline comment. 

We might should address the inline comments before merging.

crazytonyli:
@dcalhoun Thanks for the review. I have fixed all the issues.

ios/Demo-iOS/Sources/Views/PostsListView.swift
crazytonyli: For simplicity, I changed the wording to "entries"

ios/Demo-iOS/Sources/Views/PostsListView.swift
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    private func openPost(_ post: AnyPostWithEditContext) {
+        let configuration = viewModel.editorConfiguration.toBuilder()
+            .setPostType(viewModel.postTypeDetails)
+            .setPostID(Int(post.id))
+            .setTitle(post.title?.raw ?? "")
+            .setContent(post.content.raw ?? "")
+            .build()
```
dcalhoun, line 76: This doesn't take into consideration the toggles controlling the native inserter and network logging. The result is the editor opening with the default `false` value for both. 

This isn't a regression with the GBK library, but it's an oddity for the demo app. So, not a blocker, but we ideally address this oversight.

ios/Demo-iOS/Sources/Views/SitePreparationView.swift
dcalhoun, line 244: Similar to to the "Prepare Editor" [comment](https://github.com/wordpress-mobile/GutenbergKit/pull/299/changes#r2738365619), this is not preloading the correct post type. It seems we need to provide the correct post type.

dcalhoun (approved):
The latest iteration tested well for me. 

Approving this to keep things moving forward, but we should address the inline comments before merging to avoid confusing outcomes in the demo app. 

Also, will you please open an integration PR for WordPress-iOS? 

I recommend using referencing a GBK commit hash for the integration to start. Once the WordPress-iOS PR is approved, we can publish a GBK pre-release and use that.

src/utils/api-fetch.js
```diff
@@ -107,15 +107,19 @@ function tokenAuthMiddleware( options, next ) {
...
+		return next( options );
+	}
 
-	if ( isDisabled ) {
+	const disabledPath = `/${ restNamespace }/${ restBase }/${ id }`;
+
+	if (
+		options.path === disabledPath ||
+		options.path?.startsWith( `${ disabledPath }?` )
+	) {
```
crazytonyli, line 122: @dcalhoun This is a new change in the library after your approval.

I found a bug where the post editor content (after opening an existing custom post) shows on screen initially, but disappears in about a second afterwards. This issue only occurs on bundle editor, and does not occurs when using `make dev-server`.

crazytonyli:
@dcalhoun I have integrated this PR in https://github.com/wordpress-mobile/WordPress-iOS/pull/25208

ios/Demo-iOS/Sources/Views/SitePreparationView.swift
```diff
@@ -166,8 +212,12 @@ class SitePreparationViewModel {
             preconditionFailure("Unable to prepare editor without editor configuration – the UI should prevent this")
         }
 
+        let config = configuration
+            .toBuilder()
+            .setPostType(self.selectedPostTypeDetails)
+            .build()
         let cacheInterval: TimeInterval = 86_400  // Cache for one day
-        self.prepareEditor(with: EditorService(configuration: configuration, cachePolicy: .maxAge(cacheInterval)))
+        self.prepareEditor(with: EditorService(configuration: config, cachePolicy: .maxAge(cacheInterval)))
```
dcalhoun: Are these changes now unnecessary due to https://github.com/wordpress-mobile/GutenbergKit/pull/299/changes/fe7dcd46a6a4a3f28e9c9ca3e43ac9189d54bd35?

src/utils/api-fetch.js
dcalhoun: Good catch! I imagine the reason it only showcased itself intermittently is due to a race condition. 

This change makes sense given we intentionally disable API requests for entity content since we seed the title and content from the native host app.

ios/Demo-iOS/Sources/Views/SitePreparationView.swift
```diff
@@ -111,11 +135,19 @@ struct SitePreparationView: View {
...
+        didSet { updateEditorConfiguration() }
+    }
 
-    var enableNetworkLogging: Bool = false
+    var postTypes: [PostTypeDetails] = []
 
-    var postType: String = "post"
+    var selectedPostTypeDetails: PostTypeDetails = .post {
+        didSet { updateEditorConfiguration() }
+    }
```
dcalhoun: Even with the `didSet`, it appears the initial opening of an existing entry uses the incorrect values. 

Steps to reproduce:

1. Open a site with a CPT. 
2. Note the native inserter is enabled. 
3. Navigate to the CPT list. 
4. Open an existing entry. 
5. Note the block inserter is the web version, not the native version. 

Existing the editor and toggling the inserter fixes the issue. Also, creating a new entry never exhibits this issue.

dcalhoun:
I'll also a note couple of oddities, but I believe they are not strictly related to these changes. 

When switching post types...

- The preloading status note never resets. 
- The editor incorrectly hides the progress bar when opening. 

Ideally, both of these reset when switching post types as we need to fetch the newly selected post type.

crazytonyli:
@dcalhoun Thanks for finding all those issues. The demo app should be in a better shape now. 🤞

dcalhoun (approved):
Latest changes worked well for me. 🚀 

The only remaining demo app oddity I encountered is that the Local Caches section intermittently gets stuck in a loading state. I'm unsure of the reproduction steps. However, I can reproduce this on the trunk branch, so I believe it is independent of the proposed changes.

crazytonyli:
⬆️ Rebased without code changes, except removing the `make build` commits.

PR #301, opened 2026-01-27
fix: Fetch uncached assets from remote server

dcalhoun:
## What?

Prevents crashes when the editor requests assets that aren't in the local cache and adds a fallback to fetch them from the original server.

## Why?

Inserting certain plugin blocks (e.g., the Writing Prompt block) causes a crash because:

1. The block's CSS references relative assets like `url(/icon.svg)`
2. All manifest URLs are rewritten to use the `gbk-cache-https` scheme
3. The browser resolves the SVG URL against the cached CSS URL
4. The scheme handler receives a request for an SVG that was never downloaded (only `.js`/`.css` files are cached)
5. The `assetDataPath(for:)` method's precondition fails when the path escapes the bundle root, causing a crash

Fix CMM-1176. Fix JETPACK-IOS-1JFR. 

## How?

Two changes:

1. **Added `isValidAssetPath(for:)`** - A safe method to check if a URL resolves to a valid path within the bundle root, without triggering the precondition crash

2. **Added remote fetching fallback** - When an asset isn't in the local cache (invalid path or missing file), the provider now fetches it from the original HTTPS URL and serves it to the WebView

## Testing Instructions

1. Open the editor for a site with block assets support. 
3. Insert the "Writing Prompt" block (from Jetpack)
4. Verify the block loads without crashing[^1]
5. Verify the block's icon appears correctly (fetched from remote)

[^1]: I was unable to reproduce the crash locally, only in the TestFlight build. 

N/A - No UI changes

## Screenshots or screencast

N/A

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
@@ -107,7 +107,7 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
     private let editorService: EditorService
     private let mediaPicker: MediaPickerController?
     private let controller: GutenbergEditorController
-    private let bundleProvider = EditorAssetBundleProvider()
+    private let bundleProvider: EditorAssetBundleProvider
 
     // MARK: - Private Properties (UI)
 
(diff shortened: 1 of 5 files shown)


dcalhoun:
@jkmassel I reviewed and merged #302. Will you please review this again at your convenience? 🙇🏻‍♂️

jkmassel (approved):

PR #302, opened 2026-01-27
Suggestions

jkmassel:
I had a bunch of suggestions for #301, so consider this a very large "suggest" review block.

Namely:

1. Fetching anything from the remote should be done by an `EditorHttpClient` because it'll automatically deal with stuff like "make sure there's an auth header included for private sites".
2. Now that we're doing IO, the `EditorAssetBundleProvider` needs synchronization, so it's all `@MainActor` now. This means we don't need to do log coalescing (or locking) because it's all on the main thread. This also removes concurrency warnings.

I tested this against `jetpack.wpmt.co` using the demo app and it still works ok.

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/Services/EditorAssetBundleProvider.swift
+++ b/ios/Sources/GutenbergKit/Sources/Services/EditorAssetBundleProvider.swift
@@ -146,7 +146,7 @@ extension EditorAssetBundleProvider: WKURLSchemeHandler {
     /// Fetches an asset from its original remote URL and serves it to the WebView.
     ///
     /// This is used when an asset isn't in the local bundle (e.g., images referenced
-    /// in CSS files that weren't downloaded because only JS/CSS files are cached).s
+    /// in CSS files that weren't downloaded because only JS/CSS files are cached).
     private func fetchFromRemote(for urlSchemeTask: any WKURLSchemeTask) {
         guard let originalRequest = self.originalRequest(for: urlSchemeTask.request) else {
             Logger.assetLibrary.info("     Failed to construct original URL")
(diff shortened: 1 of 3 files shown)


ios/Sources/GutenbergKit/Sources/Services/EditorAssetBundleProvider.swift
```diff
@@ -152,60 +146,57 @@ extension EditorAssetBundleProvider: WKURLSchemeHandler {
...
-    /// in CSS files that weren't downloaded because only JS/CSS files are cached).
-    private func fetchFromRemote(
-        url: URL,
-        urlSchemeTask: any WKURLSchemeTask,
-        loggerMessages: inout [String]
-    ) {
-        guard let originalURL = self.originalURL(for: url) else {
-            loggerMessages.append("     Failed to construct original URL")
-            let response = HTTPURLResponse(url: url, statusCode: 400, httpVersion: nil, headerFields: nil)!
+    /// in CSS files that weren't downloaded because only JS/CSS files are cached).s
```
dcalhoun:
```suggestion
    /// in CSS files that weren't downloaded because only JS/CSS files are cached).
```

ios/Sources/GutenbergKit/Sources/Services/EditorAssetBundleProvider.swift
```diff
@@ -152,60 +146,57 @@ extension EditorAssetBundleProvider: WKURLSchemeHandler {
...
-                headerFields: httpResponse.allHeaderFields as? [String: String]
-            )!
-
-            urlSchemeTask.didReceive(schemeResponse)
-            urlSchemeTask.didReceive(data)
-            urlSchemeTask.didFinish()
         }
-        task.resume()
+
+        self.runningTasks.append(taskHandle)
```
dcalhoun, line 180: I perceive these are never removed/emptied. Is there a memory leak concern? Should we concern ourselves with cleaning these up at some point or is the risk negligible?

dcalhoun (approved):
Works well in my testing. Thank you for sharing these suggestions. Very helpful for me. I learned valuable iOS practices.

PR #305, opened 2026-01-28
fix: Editor preloading survives blocked asset downloads

dcalhoun:
## What?

Makes the editor resilient to individual asset download failures during preloading, preventing the editor from failing to load when content blockers (like AdGuard) block analytics scripts.

## Why?

Fix CMM-1179.

When AdGuard blocks `stats.wp.com/w.js`, the `EditorAssetLibrary.buildBundle()` method was failing entirely because it used `withThrowingTaskGroup` - any single asset failure would cause the entire bundle download to fail, resulting in an "Editor Error" screen with "bad URL" message.

### Production code (`EditorAssetLibrary.swift`)
- Changed `buildBundle()` from `withThrowingTaskGroup` to `withTaskGroup`
- Individual asset download failures are now caught, logged as warnings, and the download continues
- Updated `fetchAsset()` to return `Void` instead of `URL` (return value was unused)

### Test code (`EditorAssetLibraryTests.swift`)
- Simplified `EditorAssetLibraryMockHTTPClient` by:
  - Making `urlResponseHandler` throwable: `((URL) throws -> Data)`
  - Removing `shouldThrowError` and `downloadResponse` properties
  - Using the same handler for both `perform()` and `download()` methods
- Added test for graceful handling of blocked assets

## Testing Instructions

1. Use a tool to block `stats.wp.com` in some manner—e.g., AdGuard, Proxyman. 
2. Open a post in an app using GutenbergKit
3. Verify the editor loads successfully (previously showed "Editor Error")
4. Check logs for warning about failed asset download

### Automated Testing
- Run `make test-swift-package` - all tests pass including new test `buildBundle continues when individual asset downloads fail`

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/Stores/EditorAssetLibrary.swift
+++ b/ios/Sources/GutenbergKit/Sources/Stores/EditorAssetLibrary.swift
@@ -130,16 +130,22 @@ public actor EditorAssetLibrary {
         let editorRepresentation = try manifest.buildEditorRepresentation(for: self.configuration)
         try bundle.writeManifest(editorRepresentation: editorRepresentation)
 
-        try await withThrowingTaskGroup { group in
+        await withTaskGroup { group in
             let links = (manifest.scripts + manifest.styles).filter { self.isSupportedAsset($0) }
-            
+
             for asset in links {
                 group.addTask {
-                    (asset, try await self.fetchAsset(url: asset, into: bundle))
+                    do {
+                        try await self.fetchAsset(url: asset, into: bundle)
+                    } catch {
+                        // Log and continue - individual asset failures shouldn't block the editor
+                        // This handles cases like content blockers blocking analytics scripts
+                        log(.warn, "Failed to download asset \(asset.lastPathComponent): \(error.localizedDescription)")
+                    }
                 }
             }
-            
-            for try await _ in group {
+
+            for await _ in group {
                 complete += 1
                 await progress?(EditorProgress(completed: complete, total: links.count))
             }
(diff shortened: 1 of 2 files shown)


ios/Sources/GutenbergKit/Sources/Stores/EditorAssetLibrary.swift
```diff
@@ -150,20 +156,18 @@ public actor EditorAssetLibrary {
...
         let destinationPath = bundle.bundleRoot.appending(path: url.path(percentEncoded: false))
         let destinationParent = destinationPath.deletingLastPathComponent()
-        
+
         // Ensure the destination directory exists
         try FileManager.default.createDirectory(at: destinationParent, withIntermediateDirectories: true)
 
         try FileManager.default.copyItem(at: tempUrl, to: destinationPath)
-
-        return destinationPath
```
dcalhoun, line 166: This was unused. Removed to simplify the new `withTaskGroup` syntax.

ios/Tests/GutenbergKitTests/Stores/EditorAssetLibraryTests.swift
```diff
@@ -844,8 +901,6 @@ final class EditorAssetLibraryMockHTTPClient: EditorHTTPClientProtocol, @uncheck
     var getCallCount = 0
     var downloadCallCount = 0
     var downloadedURLs: [URL] = []
-    var downloadResponse: URL?
-    var shouldThrowError: Error?
```
dcalhoun, line 848: Removed in favor of composing the `urlResponseHandler` for unique scenarios—e.g., URL-specific data and errors.

ios/Sources/GutenbergKit/Sources/Stores/EditorAssetLibrary.swift
```diff
@@ -150,20 +156,18 @@ public actor EditorAssetLibrary {
     
     /// Downloads a single asset and copies it into the temporary bundle directory.
     ///
-    private func fetchAsset(url: URL, into bundle: EditorAssetBundle) async throws -> URL {
+    private func fetchAsset(url: URL, into bundle: EditorAssetBundle) async throws {
```
jkmassel:
```suggestion
    @discardableResult
    private func fetchAsset(url: URL, into bundle: EditorAssetBundle) async throws -> URL {
```

At least for testing, it's probably useful to know exactly where the file ended up. If we want to make it not-a-warning to ignore it, we can use `@discardableResult`

jkmassel (approved):
Discussed in Slack – this seems ok – when combined with #301 it'll make it so that we fetch any failed assets at runtime.

ios/Sources/GutenbergKit/Sources/Stores/EditorAssetLibrary.swift
dcalhoun: Makes sense. Addressed in 728a18880c8e585bbff86f7e81afdf15be72433a.

PR #181, opened 2025-09-19
feat: AJAX uses token authentication

dcalhoun:
Related: 

- 192611-ghe-Automattic/wpcom
- https://github.com/Automattic/jetpack/pull/45220
- https://github.com/Automattic/jetpack/pull/45254
- https://github.com/Automattic/jetpack/pull/47827

## What?

Authenticate AJAX requests with application passwords sent via an authorization header. Vendor [`wp-util.js`](https://github.com/WordPress/wordpress-develop/blob/117af7e9a37c02ee17ac8f143cb46b6b0f4cde15/src/js/_enqueues/wp/util.js) and configure the WordPress AJAX and media globals.

## Why?

The GutenbergKit editor does not have authorization cookies, so we must rely upon a different authorization mechanism. Additionally, GutenbergKit excludes core WordPress assets from the editor assets endpoint, so `wp-util.js` (which provides `wp.ajax` and `wp.template`) must be vendored and loaded directly.

Ref CMM-713. Close CMM-768. 

## How?

- Set the `Authorization` header via `jQuery.ajaxPrefilter` and by overloading the `window.wp.ajax` utilities. This general-purpose AJAX auth is always initialized.
- Vendor and load `wp-util.js` after jQuery and lodash are on `window`, since its IIFE captures jQuery via closure at execution time.
- Filter the `lodash-js-after` inline script from editor assets. WordPress's `_.noConflict()` call wipes `window._` because GutenbergKit doesn't load Underscore.js.
- Alias `wp.media.ajax` and `wp.media.post` to the authenticated `wp.ajax` methods, since WordPress core's [`media-models.js`](https://github.com/WordPress/wordpress-develop/blob/117af7e9a37c02ee17ac8f143cb46b6b0f4cde15/src/js/_enqueues/wp/media/models.js) is not loaded.
- Conditionally initialize the VideoPress AJAX bridge when `videopress/video` is not in `allowed_block_types`, so `core/video` blocks extended to rely upon VideoPress upload services continue to work.
- Derive the Android `WebViewAssetLoader` domain from `siteURL` so that the editor document shares the site's origin, making REST API and `admin-ajax.php` requests same-origin and eliminating CORS restrictions without server-side headers.
- Fix the iOS demo app to use the `home` URL from the REST API response instead of `url`, which returns `http://` for WordPress.com sites.

### 1. Verify AJAX requests use token authentication

1. `make build`
1. Run the demo app in Xcode/Android Studio (use production build, not dev server). 
[...]

PR #316, opened 2026-02-10
Integrate WPAndroid

jkmassel:
## What?

Integrate GutenbergKit with WordPress.com (WP Android) by moving editor loading UI into `GutenbergView`, adding WP.com REST API namespace support, adding native CORS proxy for Android, improving HTTP logging on both platforms, and fixing falsy post ID handling.

## Why?

This branch prepares GutenbergKit for use in WP Android, which requires several changes:

1. **Internal loading UI**: Host apps (like WP Android) shouldn't need to implement their own loading/progress/error UI — `GutenbergView` should manage this internally, matching how iOS already works.
2. **WP.com namespace support**: WP.com's REST API uses a `sites/{id}/` namespace segment (e.g., `/wp/v2/sites/123/posts`) that `RESTAPIRepository` didn't support.
3. **CORS proxy**: The Android WebView serves pages from `appassets.androidplatform.net`, making every `fetch()` to the WordPress REST API cross-origin. A native request proxy bypasses CORS without requiring server-side changes.
4. **Post ID = 0 handling**: Both platforms and the JS bridge treated `0` as a valid post ID, which caused the editor to attempt fetching a non-existent post.
5. **Verbose HTTP logging**: Debugging integration issues requires detailed request/response logging on both iOS and Android `EditorHTTPClient`.

### Android: Move loading UI into `GutenbergView` (refactor)
- Changed `GutenbergView` from extending `WebView` to extending `FrameLayout`, containing an internal `WebView` plus overlay views (`EditorProgressView`, `ProgressBar`, `EditorErrorView`).
- Removed the `EditorLoadingListener` interface and all consumer-side loading UI from the demo app's `EditorActivity`.
- Loading states (`PROGRESS` → `SPINNER` → `READY` / `ERROR`) are managed internally with animated transitions.

### Android: CORS proxy in `shouldInterceptRequest`
- Added `proxyRequest()` to `GutenbergView` that intercepts non-asset, non-cached requests and performs them via `HttpURLConnection`, bypassing CORS.
- Handles OPTIONS preflight requests with a synthetic 204 response.
- Forwards cookies from `CookieManager` and adds `Access-Control-Allow-*` headers to responses.

### WP.com namespace support (iOS + Android)
- Added `buildNamespacedURL()` to `RESTAPIRepository` (Android) and equivalent logic on iOS that inserts the namespace (e.g., `sites/123/`) after the API version segment in REST paths.
- Added tests for namespaced URL building on both platforms.

### Post ID = 0 → nil
- `EditorConfiguration` on both platforms now coerces `postId = 0` to `nil`.
- JS bridge uses `||` instead of `??` for `post.id` so falsy values (0, null, undefined) fall back to `-1`.
- Added unit tests on both platforms.

### Verbose HTTP logging (iOS + Android)
- `EditorHTTPClient` on both platforms now logs full request/response details (method, URL, headers, status code, body on error, WP error codes) at debug/error levels.

### Other
- Updated AGP to 8.10.1 to match WP Android.
- Skip theme style loading in `EditorService` when `themeStyles` is disabled.
- Deleted unused `activity_editor.xml` layout file.

## Testing Instructions

1. Open the Android demo app and verify the editor loads with built-in progress bar → spinner → editor transitions.
2. Verify no loading UI code is needed in the host activity.
3. Test with a WP.com site using `siteApiNamespace: ["sites/{id}/"]` and confirm REST API calls include the namespace.
4. Create a new post (post ID = 0) and confirm the editor renders without trying to fetch post 0.
5. Check logcat/console for detailed HTTP request/response logs during editor initialization.

Loading states use standard Android `ProgressBar` and `TextView` components with system accessibility defaults. No custom accessibility handling was added — verify screen readers announce the progress bar and error state text.

## Screenshots or screencast

N/A – no visual design changes beyond moving existing loading UI into the library.

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/EditorHTTPClient.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/EditorHTTPClient.kt
@@ -25,12 +25,27 @@ interface EditorHTTPClientProtocol {
 }
 
 /**
+ * The response data from an HTTP request, either in-memory bytes or a downloaded file.
+ */
+sealed class EditorResponseData {
+    data class Bytes(val data: ByteArray) : EditorResponseData() {
+        override fun equals(other: Any?): Boolean {
+            if (this === other) return true
+            if (other !is Bytes) return false
+            return data.contentEquals(other.data)
+        }
+        override fun hashCode(): Int = data.contentHashCode()
+    }
+    data class File(val file: java.io.File) : EditorResponseData()
+}
+
+/**
  * A delegate for observing HTTP requests made by the editor.
  *
  * Implement this interface to inspect or log all network requests.
  */
 interface EditorHTTPClientDelegate {
-    fun didPerformRequest(url: String, method: EditorHttpMethod, response: Response, data: ByteArray)
+    fun didPerformRequest(url: String, method: EditorHttpMethod, response: Response, data: EditorResponseData)
 }
 
 /**
@@ -141,18 +156,32 @@ class EditorHTTPClient(
 
     override suspend fun download(url: String, destination: File): EditorHTTPClientDownloadResponse =
         withContext(Dispatchers.IO) {
+            Log.d(TAG, "DOWNLOAD $url")
+            Log.d(TAG, "  Destination: ${destination.absolutePath}")
+
             val request = Request.Builder()
                 .url(url)
                 .addHeader("Authorization", authHeader)
                 .get()
                 .build()
 
-            val response = client.newCall(request).execute()
+            Log.d(TAG, "  Request headers: ${redactHeaders(request.headers)}")
+
+            val response: Response
+            try {
+                response = client.newCall(request).execute()
+            } catch (e: IOException) {
+                Log.e(TAG, "DOWNLOAD $url – network error: ${e.message}", e)
+                throw e
+            }
+
             val statusCode = response.code
             val headers = extractHeaders(response)
+            Log.d(TAG, "DOWNLOAD $url – $statusCode")
+            Log.d(TAG, "  Response headers: ${redactHeaders(response.headers)}")
 
             if (statusCode !in 200..299) {
-                Log.e(TAG, "HTTP error downloading $url: $statusCode")
+                Log.e(TAG, "DOWNLOAD $url – HTTP error: $statusCode")
                 throw EditorHTTPClientError.DownloadFailed(statusCode)
             }
 
@@ -163,8 +192,14 @@ class EditorHTTPClient(
                         input.copyTo(output)
                     }
                 }
-                Log.d(TAG, "Downloaded file: file=${destination.absolutePath}, size=${destination.length()} bytes, url=$url")
-            } ?: throw EditorHTTPClientError.DownloadFailed(statusCode)
+                Log.d(TAG, "DOWNLOAD $url – complete (${destination.length()} bytes)")
+                Log.d(TAG, "  Saved to: ${destination.absolutePath}")
+            } ?: run {
+                Log.e(TAG, "DOWNLOAD $url – empty response body")
+                throw EditorHTTPClientError.DownloadFailed(statusCode)
+            }
+
+            delegate?.didPerformRequest(url, EditorHttpMethod.GET, response, EditorResponseData.File(destination))
 
             EditorHTTPClientDownloadResponse(
                 file = destination,
@@ -175,6 +210,8 @@ class EditorHTTPClient(
 
     override suspend fun perform(method: EditorHttpMethod, url: String): EditorHTTPClientResponse =
         withContext(Dispatchers.IO) {
+            Log.d(TAG, "$method $url")
+
             // OkHttp requires a body for POST, PUT, PATCH methods
             // GET, HEAD, OPTIONS, DELETE don't require a body
             val requiresBody = method in listOf(
@@ -190,7 +227,15 @@ class EditorHTTPClient(
                 .method(method.toString(), requestBody)
                 .build()
 
-            val response = client.newCall(request).execute()
+            Log.d(TAG, "  Request headers: ${redactHeaders(request.headers)}")
+
+            val response: Response
+            try {
+                response = client.newCall(request).execute()
+            } catch (e: IOException) {
+                Log.e(TAG, "$method $url – network error: ${e.message}", e)
+                throw e
+            }
 
             // Note: This loads the entire response into memory. This is acceptable because
             // this method is only used for WordPress REST API responses (editor settings, post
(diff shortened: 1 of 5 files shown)


android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -190,96 +227,245 @@ class GutenbergView : WebView {
...
+                    // Cache miss – fall through to proxy below
+                }
+
+                // Proxy all other requests natively so they bypass CORS
+                // (the editor page is served from appassets.androidplatform.net,
+                // making every API call cross-origin). This doesn't use `EditorHTTPClient` because
+                // it's not actually a native call – the editor is building the request so we don't
+                // want to modify it in any way.
+                Log.d(TAG, "shouldInterceptRequest: proxying request: ${request.url}")
+                return proxyRequest(request)
```
dcalhoun: I encounter plugin load failures from CORS errors when launching the editor for a WoW/Atomic site or `jetpack.wpmt.co`. 

<details><summary>Plugin load failure</summary>

Image

</details>

<details><summary>CORS errors</summary>

Image

</details>

Disabling this proxy by restate the previous implementation resolves the issue. 

<details><summary>Example revert diff</summary>

```diff
diff --git a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
index 32c10632..78ec151d 100644
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
@@ -414,6 +414,8 @@ class GutenbergView : FrameLayout {
                     // Cache miss – fall through to proxy below
                 }
 
+                return super.shouldInterceptRequest(view, request)
+
                 // Proxy all other requests natively so they bypass CORS
                 // (the editor page is served from appassets.androidplatform.net,
                 // making every API call cross-origin). This doesn't use `EditorHTTPClient` because

```

</details>

There appears to be something off with the new proxy logic.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -341,39 +529,149 @@ class GutenbergView : WebView {
...
+     * Proxies an HTTP request natively to bypass CORS restrictions.
+     *
+     * The editor page is served from `https://appassets.androidplatform.net`, so every
+     * `fetch()` call the JavaScript makes to the WordPress REST API is cross-origin.
+     * Rather than requiring the server to whitelist the synthetic WebView origin, this
+     * method intercepts the request in [shouldInterceptRequest] and performs it with
+     * OkHttp, which is not subject to CORS. OkHttp's connection pool handles cleanup
+     * automatically when the response stream is closed by the WebView.
+     */
+    private fun proxyRequest(request: WebResourceRequest): WebResourceResponse? {
```
dcalhoun: Presumably, this would negate the need for server logic like https://github.com/Automattic/jetpack/pull/45292. Correct?

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
dcalhoun: This same issue appears to cause _all requests_ to fail with CORS errors. E.g., uploading an image fails.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -341,39 +529,149 @@ class GutenbergView : WebView {
...
+            return createCorsPreflightResponse(requestedHeaders)
+        }
+
+        try {
+            val method = request.method ?: "GET"
+            val isTrustedHost = trustedHosts.contains(request.url.host)
+
+            // OkHttp requires a body for POST/PUT/PATCH even if empty
+            val requiresBody = method in listOf("POST", "PUT", "PATCH")
+            val body = if (requiresBody) "".toRequestBody(null) else null
```
dcalhoun: Presumably, this (erroneously?) empty body is the reason image uploads fail. 

From researching, it seems you cannot retrieve/pass along body values from `WebResourceRequest`. Is proxying mutating requests actually possible in this manner or do we need to abandon this newly added proxy?

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -341,39 +529,149 @@ class GutenbergView : WebView {
...
+                .map { it.trim() }
+                .find { it.startsWith("charset=", ignoreCase = true) }
+                ?.substringAfter("=")
+                ?.trim()
+                ?: "UTF-8"
+
+            val responseHeaders = mutableMapOf<String, String>()
+            response.headers.forEach { (key, value) ->
+                responseHeaders[key] = value
+            }
```
dcalhoun: Claude noted this may erroneously discard duplicate headers (e.g., multiple `Set-Cookie`). Should we join duplicates?

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -82,15 +89,40 @@ const val ASSET_URL = "https://appassets.androidplatform.net/assets/index.html"
...
+    private val trustedHosts: Set<String> by lazy {
+        buildSet {
+            Uri.parse(configuration.siteApiRoot).host?.let { add(it) }
+            Uri.parse(configuration.siteURL).host?.let { add(it) }
+            add("public-api.wordpress.com")
+            // WordPress.com CDN hosts (Photon image proxy)
+            add("i0.wp.com")
+            add("i1.wp.com")
+            add("i2.wp.com")
+            add("i3.wp.com")
```
dcalhoun: Is it feasible for us to hoist these WP.com references to host app (Wordpress and/or Demo) to keep GutenbergKit free of these references?

android/Gutenberg/src/main/java/org/wordpress/gutenberg/views/EditorErrorView.kt
```diff
@@ -0,0 +1,84 @@
...
+        // Create title
+        titleText = TextView(context).apply {
+            layoutParams = LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.WRAP_CONTENT).apply {
+                topMargin = dpToPx(16)
+                marginStart = dpToPx(16)
+                marginEnd = dpToPx(16)
+            }
+            gravity = Gravity.CENTER
+            TextViewCompat.setTextAppearance(this, android.R.style.TextAppearance_Material_Subhead)
+            text = "Failed to load editor"
```
dcalhoun, line 53: For a future PR, we should likely localize strings like this in a manner similar to iOS. 

https://github.com/wordpress-mobile/GutenbergKit/blob/5b66f8ec33edc49004dbd27b59dfd53279acdf38/ios/Sources/GutenbergKit/Sources/EditorLocalization.swift#L33-L56

dcalhoun (commented):
Thank you for implementing this! 🙇🏻‍♂️ 

The overarching approach looks sound to me; it's aligned with much of the current iOS implementation. 

While testing, I encountered a few critical issues that I noted in inline comments. We should discuss and address those.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
jkmassel: I'd prefer to avoid requiring Jetpack for this feature to work, but WDYT?

dcalhoun: Yes, agreed. If we can avoid the need for Jetpack plugin code for `admin-ajax` support that'd be great. 

However, the CORS fix is only one facet of that support. For `admin-ajax` support, we still must enable authenticating those requests with app passwords with the [`application_password_is_api_request`](https://developer.wordpress.org/reference/hooks/application_password_is_api_request/) hook we use in https://github.com/Automattic/jetpack/pull/45220. 

I.e., if we solve the CORS problem in the client, Jetpack is still required to enable application password authentication of `admin-ajax` (VideoPress only). If other plugins want that support, they need to opt in for themselves via the hook.

dcalhoun: Noting that I determined that we could negate the need for https://github.com/Automattic/jetpack/pull/45292#issuecomment-4146205203 by using the [site URL](https://github.com/wordpress-mobile/GutenbergKit/pull/181/changes/78bc47ece365ed012dad2119c630ebe63778654c) for loading the local editor assets in the WebView.

PR #318, opened 2026-02-11
Implement GutenbergKitResources binary dependency build

mokagio:
> [!NOTE]  
> Closed in favor of #376

## What?

Adds infrastructure to build and distribute `GutenbergKitResources` as a pre-built XCFramework, decoupling the JS build from Swift consumers.

Based on the research work done on #315 .

## Why?

Currently, every Swift consumer must run the full JS build to get Gutenberg editor assets.
With a binary XCFramework for resources, consumers can pull a pre-built artifact from CDN instead, significantly reducing build times and simplifying integration.

## How?

- New `GutenbergKitResources` SPM target wrapping bundled editor assets (HTML, CSS, JS)
- `Package.swift` uses `Context.environment` to switch between local source and CDN binary target
- `build_xcframework.sh` assembles a proper `.framework` from SPM archive DerivedData (binary, swiftmodule, modulemap, headers, resource bundles) since SPM archives don't produce installable frameworks
- Fastlane `release` lane handles versioning, GitHub release, and S3 upload
- CI pipeline updated with XCFramework build step and artifact publishing
- `GutenbergKit` target depends on `GutenbergKitResources` instead of bundling assets directly
- All `package` access modifiers migrated to `internal` (required by `packageAccess: false`)
- Tracked JS build artifacts removed from Git

## Testing Instructions

1. `make build` — builds web assets and copies to all platform directories
2. `make build-swift-package` — builds the Swift package with local resources
3. `make build-resources-xcframework` — produces `GutenbergKitResources-<sha>.xcframework.zip`
4. Verify the XCFramework contains both `ios-arm64` and `ios-arm64_x86_64-simulator` slices with the binary, Modules, Headers, and resource bundle

---

_Generated with the help of Claude Opus 4.6_


fastlane/Fastfile
```diff
@@ -0,0 +1,173 @@
...
+PROJECT_ROOT = File.expand_path('..', __dir__)
+
+LANE_VALUE_VERSION = 'GBK_VERSION'
+LANE_VALUE_GITHUB_TOKEN = 'GBK_GITHUB_TOKEN'
+
+GITHUB_REPO = 'wordpress-mobile/GutenbergKit'
+GIT_REMOTE_NAME = 'origin'
+GIT_BASE_BRANCH = 'trunk'
+
+# Code Signing
```
mokagio:
```suggestion
```

ios/Sources/GutenbergKit/Sources/Model/EditorCachePolicy.swift
```diff
@@ -94,7 +94,7 @@ public enum EditorCachePolicy: Sendable {
     ///   (i.e., the cached response hasn't expired yet).
     /// - ``always``: Always returns `true` - cached responses are always used.
     ///
-    package func allowsResponseWith(date: Date, currentDate: Date = .now) -> Bool {
+    func allowsResponseWith(date: Date, currentDate: Date = .now) -> Bool {
```
mokagio, line 97: SwiftPM required `packageAccess: false` on the GutenbergKit target for binary target compatibility.

With that flag off, the compiler no longer passes `-package-name`, so the `package` access modifier silently degrades to `fileprivate` breaking visibility across files within the module.

ios/Sources/GutenbergKit/Sources/Views/HTMLPreview/HTMLPreviewManager.swift
```diff
@@ -52,7 +53,7 @@ public final class HTMLPreviewManager: ObservableObject {
     // MARK: - Initialization
 
     public init(themeStyles: String? = nil) {
-        let gutenbergCSS = Self.loadGutenbergCSS() ?? ""
+        let gutenbergCSS = GutenbergKitResources.loadGutenbergCSS() ?? ""
         assert(!gutenbergCSS.isEmpty, "Failed to load Gutenberg CSS from bundle. Previews will not render correctly.")
```
mokagio, line 57: There might be more robust ways to handle a path error internally, but I consider them out of scope for the moment.

.gitignore
```diff
@@ -189,9 +189,15 @@ local.properties
 /android/Gutenberg/src/main/assets/assets
 /android/Gutenberg/src/main/assets/index.html
 
-# Disabled removing these files until this is published like Android in CI.
-# /ios/Sources/GutenbergKit/Gutenberg/assets
-# /ios/Sources/GutenbergKit/Gutenberg/index.html
+/ios/Sources/GutenbergKit/Gutenberg/assets
+/ios/Sources/GutenbergKit/Gutenberg/index.html
```
mokagio: These should not actually be generated anymore. We can keep them here for a while to avoid issues while local copies update, or get rid of them already to avoid carrying dead weight. What do you think?

```suggestion
```

build_xcframework.sh
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env bash
```
mokagio, line 1: It took me a while to get this to run in #315 and it's based on a setup that had it in the root and I wasn't game to move it just yet.

mokagio: https://linear.app/a8c/issue/AINFRA-1983/followup-gutenbergkitresources-setup-move-script-away-from-root tracks moving it in a better location

mokagio:
@dcalhoun I can't tell if the E2E tests are due to my changes or not. I noticed you recently worked on an E2E fix, which I rebased this on top of. Maybe with your extra context you can let me know? Thanks!

dcalhoun:
> @dcalhoun I can't tell if the E2E tests are due to my changes or not. I noticed you recently worked on an E2E fix, which I rebased this on top of. Maybe with your extra context you can let me know? Thanks!

Thanks for the continued efforts on this, @mokagio! 🙇🏻‍♂️ 

It seems the CI failures are likely related to the changes in this pull request. When running the iOS UI tests locally in Xcode, I see errors related to the `GutenbergKitResources` module. The line of the error changes based on whether one is running the tests against a production build (`make build`) or development server (`make dev-server` with a `GUTENBERG_EDITOR_URL` environment variable [set in Xcode](https://github.com/wordpress-mobile/GutenbergKit/blob/8a5ceb3848d6c29313a3c3fb36c1afc53949a8e0/docs/code/getting-started.md#ios)). See the details below. 

<details><summary>Dev server error</summary>
<p>

```
GutenbergKitResources/resource_bundle_accessor.swift:44: Fatal error: unable to find bundle named GutenbergKit_GutenbergKitResources
```

I was able to circumvent the error with the following diff. 

```diff
diff --git a/ios/Sources/GutenbergKit/Sources/EditorViewController.swift b/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
index 1993d818..828d2519 100644
--- a/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
@@ -465,7 +465,7 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
                 onSelection: { [weak self] in self?.didSelectBlockInserterItem($0) },
                 onClose: { [weak self] in self?.notifyInserterClosed() }
             )
-            .environmentObject(htmlPreviewManager)
+//            .environmentObject(htmlPreviewManager)
         })
 
         context.viewController = host

```

error-dev-server

</p>
</details> 

<details><summary>Production build error</summary>
<p>

```
GutenbergKitResources/resource_bundle_accessor.swift:44: Fatal error: unable to find bundle named GutenbergKit_GutenbergKitResources
```

I was unable to identify a quick way to circumvent the error. 

error-production-build

</p>
</details>

mokagio:
Thanks for the hints @dcalhoun . I'll see where they lead.

Makefile
```diff
@@ -318,8 +334,19 @@ test-android-e2e-dev: ## Run Android E2E tests against the Vite dev server (must
...
 
+.PHONY: release-on-ci
+release-on-ci: ## Trigger a release build on Buildkite (requires BUILDKITE_API_ACCESS_TOKEN and NEW_VERSION)
+	@[ -n "$(BUILDKITE_API_ACCESS_TOKEN)" ] || (echo "Error: BUILDKITE_API_ACCESS_TOKEN is not set" && exit 1)
+	@[ -n "$(NEW_VERSION)" ] || (echo "Error: NEW_VERSION is not set. Usage: make release-on-ci NEW_VERSION=1.2.3" && exit 1)
+	@echo "Triggering release build for version $(NEW_VERSION)..."
+	@curl -sX POST "https://api.buildkite.com/v2/organizations/automattic/pipelines/gutenbergkit/builds" \
+		-H "Authorization: Bearer $(BUILDKITE_API_ACCESS_TOKEN)" \
+		-H "Content-Type: application/json" \
+		-d '{"commit":"HEAD","branch":"ainfra-1967-implement-gutenbergkitresources-binary-dependency-build","message":"Release $(NEW_VERSION)","env":{"NEW_VERSION":"$(NEW_VERSION)"}}' \
```
dcalhoun, line 345: Noting that we should update the `branch` before merging.

.buildkite/pipeline.yml
```diff
@@ -54,7 +54,13 @@ steps:
       plugins: *plugins
 
     - label: ':swift: Test Swift Package'
-      command: swift test
+      command: make test-swift-package REFRESH_L10N=1 REFRESH_JS_BUILD=1
```
dcalhoun, line 57: After https://github.com/wordpress-mobile/GutenbergKit/pull/320, we now cache the JavaScript build as an artifact to improve speed and mitigate 429 errors for translation downloads. Can we reuse that artifact for this job?

.buildkite/release.sh
```diff
@@ -0,0 +1,22 @@
...
+fi
+
+echo "--- :robot_face: Use bot for Git operations"
+source use-bot-for-git
+
+echo "--- :rubygems: Installing gems"
+install_gems
+
+echo "--- :xcode: Building XCFramework"
+make build-resources-xcframework REFRESH_L10N=1 REFRESH_JS_BUILD=1
```
dcalhoun, line 19: Similar to my comment on the Buildkite configuration, should/can we reuse the JavaScript build artifact?

dcalhoun: Struggling to understand how these XCFramework releases coincide/integrate with GutenbergKit releases. 

The existing release script for the latter bumps the version for inclusion in the JavaScript, so that accurate version numbers are [reported in exceptions](https://github.com/wordpress-mobile/GutenbergKit/blob/cdf828849cd2fb555ea584fcd5bb61500db89268/src/utils/exception-parser.js#L96). Should the XCFramework and GutenbergKit versions align? Maybe that is only required for tagged GutenbergKit releases, and XCFramework releases for Git commits can simply use the last tagged version string? 

I had Claude capture the issue and potential solution in its own words below as well. I'm remain uncertain as to the current state, plan, or appropriate solution. 

WDYT? 

> Additionally, the JS assets baked into the XCFramework will carry whatever version is in `package.json` at the time of the build — currently `0.14.0-alpha.0` on `trunk`. This version propagates to crash report tags (`gutenberg_kit_version` in `exception-parser.js`) and the generated `GutenbergKitVersion.swift`/`.kt` files (used in User-Agent strings).
> 
> Since `package.json` is never bumped to `$version` before the build runs, the XCFramework artifact will report a stale version at runtime even though the git tag and S3 path use the correct release version.
> 
> I think this needs an `npm --no-git-tag-version version "$version"` before the `make` call, so the version flows through the same `package.json` → `generate-version.js` → Vite build pipeline that the local release script (`bin/release.sh`) already uses.

fastlane/Fastfile
```diff
@@ -0,0 +1,177 @@
...
+    if line.start_with?('let resourcesMode: DependencyMode =')
+      "let resourcesMode: DependencyMode = .release(version: \"#{version}\", checksum: \"#{xcframework_checksum}\")\n"
+    else
+      line
+    end
+  end
+  File.open(file_path, 'w') { |file| file.print lines.join }
+end
+
+lane :publish_release_to_github do |options|
```
dcalhoun, line 72: What, if any, changes should we make the project's existing `bin/release.sh` script? Does it remain valid for a different purpose? Should we update it to [align with App Infra best practices/tooling](https://github.com/wordpress-mobile/GutenbergKit/pull/151#issuecomment-3014087441)? 

This may be out of scope for this PR, but it felt worthy of discussing here nonetheless.

.gitignore
dcalhoun: I suggest we remove them. Developers can clean the files as needed.

Makefile
```diff
@@ -318,8 +334,19 @@ test-android-e2e-dev: ## Run Android E2E tests against the Vite dev server (must
...
+		| jq -r '"Build triggered: https://buildkite.com/automattic/gutenbergkit/builds/\(.number)"'
+
 .PHONY: release
-release: ## Create and publish a new release
+release: ## Create and publish a new release (local)
```
dcalhoun, line 349: I'm unsure of the `(local)` designation. It's local in the sense that this script was intended to be run a local development machine, but it does publish tags and GitHub Releases—releases intended to be downloaded and consumed by a host app. 

Should `release` and `release-on-ci` be labeled as `GutenbergKit` and `GutenbergKitXCFramework` respectively? My perception is that they "release" two, disparate things. Is that accurate or am I misunderstanding the state of things? 

Do we plan to use both targets in tandem when publishing GutenbergKit release? I suppose this relates to my other comment about coordinating/composing release version bumps/strings.

build_xcframework.sh
dcalhoun, line 1: Transparently, I'm less familiar with the inner workings of Swift/Xcode builds. That said, this script makes sense and looks appropriate to me.

dcalhoun (commented):
This looks great, @mokagio. Very excited to see this progress. Thank you! 

I left a few inline comments that are mostly to help me better understand the functionality and underlying assumptions.

.buildkite/pipeline.yml
mokagio: For sure!

https://linear.app/a8c/issue/AINFRA-2159/resue-cached-js-in-ci

.buildkite/release.sh
mokagio: Thanks for this @dcalhoun . To be honest, I'm still unsure about how to proceed, too. So far, I've just tried to mimic the wordpress-rs behavior to have CI track the checksum and revision in the `Package.swift` upon "releasing".

You bring up a good point regarding the version bump. It's frustrating that there's no way to do a version bump and `Package.swift` update atomically. 

I'll have to think about this...

mokagio:
Closing in favor of #376 because I need to do more thinking around the XCFramework and release automation.

.buildkite/pipeline.yml
mokagio: Connecting the dots. This was done via https://github.com/wordpress-mobile/GutenbergKit/pull/513

PR #320, opened 2026-02-15
ci: Mitigate translation download 429 errors

dcalhoun:
## What?

Reduces translation download requests from ~192 to ~48 per CI run by building once and sharing artifacts, removing translations from `postinstall`, and adds robust 429 rate limit handling so the remaining 48 requests recover gracefully when throttled.

## Why?

Fix CMM-1250.

Every Buildkite CI run makes ~192 HTTP requests to `translate.wordpress.org` because 4 of 6 pipeline steps independently download all 48 translation locales. This causes 429 (rate limiting) errors. The pipeline had no artifact sharing — the "Build React App" and "Publish Android Library" steps each ran a full `make build REFRESH_L10N=1`, and the lint/test steps triggered downloads via `npm ci`'s `postinstall` hook.

Even after reducing to ~48 requests, builds were still failing because the download script's retry logic was too aggressive — batch size of 10 with 1-2s backoff meant retries also got 429'd immediately.

### 1. Remove `prep-translations` from `postinstall` in `package.json`

Translation downloads are a build concern, not an install concern. The `make` targets (`make build`, `make dev-server`, etc.) already handle translations via the `prep-translations` dependency. This eliminates the translation downloads that were happening in lint/test CI steps (which run `npm ci` but don't need translations).

### 2. Restructure `.buildkite/pipeline.yml`

- **Build React App** gets a `key: build-react` and uploads `dist/` as a Buildkite artifact
- **Lint** and **Test JS** are unchanged — `npm ci` no longer triggers translation downloads
- **Publish Android Library** gets `depends_on: build-react`, downloads the pre-built `dist/` artifact instead of running its own `make build`

| Step            | Before      | After                                |
| --------------- | ----------- | ------------------------------------ |
| Build React App | 48 requests | 48 requests (unchanged, sole source) |
| Lint React App  | 48 requests | 0 (postinstall no longer downloads)  |
| Test JavaScript | 48 requests | 0 (postinstall no longer downloads)  |
| Publish Android | 48 requests | 0 (uses artifact, no npm ci)         |
| **Total**       | **~192**    | **~48**                              |

### 3. Fail build on translation fetch errors in `bin/prep-translations.js`

Previously, translation fetch failures were silently swallowed. The script now properly propagates errors so CI fails visibly rather than producing a build with missing translations.

### 4. Improve 429 rate limit handling in `bin/prep-translations.js`

- **Reduce batch size** from 10 to 5 concurrent requests
- **Add inter-batch delay** of 2s between batches to spread load
- **Detect 429 specifically** via a `RateLimitError` class that carries the parsed `Retry-After` value
- **Respect `Retry-After` header** (supports both delta-seconds and HTTP-date formats)
- **Exponential backoff for 429s** without `Retry-After`: 5s, 10s, 20s, 40s + jitter
- **Linear backoff for other errors**: 1s, 2s, 3s (unchanged behavior)
- **Increase max retries** from 3 to 5
- **Add 60s max backoff** safety limit to avoid blocking CI indefinitely
- **Log configuration at startup** for CI visibility
- **All parameters configurable** via environment variables (`L10N_BATCH_SIZE`, `L10N_BATCH_DELAY_MS`, `L10N_MAX_RETRIES`, `L10N_MAX_BACKOFF_MS`)

## Testing Instructions

1. Verify in Buildkite that:
   - Build React App step downloads translations and uploads `dist.tar.gz`
   - Lint and Test JS steps run `npm ci` without downloading translations
   - Publish Android Library step waits for build, downloads artifact, and publishes without running `npm ci`
   - Any 429 errors from `translate.wordpress.org` are recovered via backoff and retry
2. Local verification:
   - Run `npm ci` and confirm translations are **not** downloaded
   - Run `npm run prep-translations -- --force` and confirm translations download with the new config logged (`batch=5, delay=2000ms, retries=5, maxBackoff=60000ms`)
   - Run `L10N_BATCH_SIZE=2 L10N_BATCH_DELAY_MS=5000 npm run prep-translations -- --force` and confirm env var overrides are respected in the config log
   - Run `make dev-server` and confirm the editor works with translations

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -7,9 +7,6 @@ env:
 steps:
     - label: ':react: Build React App'
       key: build-react
-      env:
-          L10N_BATCH_SIZE: '2'
-          L10N_BATCH_DELAY_MS: '5000'
       command: |
           make build REFRESH_L10N=1 REFRESH_JS_BUILD=1 STRICT_L10N=1
           tar -czf dist.tar.gz dist/
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -8,7 +8,7 @@ steps:
     - label: ':react: Build React App'
       key: build-react
       command: |
-          make build REFRESH_L10N=1 REFRESH_JS_BUILD=1
+          make build REFRESH_L10N=1 REFRESH_JS_BUILD=1 STRICT_L10N=1
           tar -czf dist.tar.gz dist/
           buildkite-agent artifact upload dist.tar.gz
       plugins: &plugins
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -7,6 +7,9 @@ env:
 steps:
     - label: ':react: Build React App'
       key: build-react
+      env:
+          L10N_BATCH_SIZE: '2'
+          L10N_BATCH_DELAY_MS: '5000'
       command: |
           make build REFRESH_L10N=1 REFRESH_JS_BUILD=1 STRICT_L10N=1
           tar -czf dist.tar.gz dist/
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -6,7 +6,11 @@ env:
 
 steps:
     - label: ':react: Build React App'
-      command: make build REFRESH_L10N=1 REFRESH_JS_BUILD=1
+      key: build-react
+      command: |
+          make build REFRESH_L10N=1 REFRESH_JS_BUILD=1
+          tar -czf dist.tar.gz dist/
+          buildkite-agent artifact upload dist.tar.gz
       plugins: &plugins
           - $CI_TOOLKIT_PLUGIN
           - $NVM_PLUGIN
@@ -20,8 +24,12 @@ steps:
       plugins: *plugins
 
     - label: ':android: Publish Android Library'
+      depends_on: build-react
       command: |
-          make build REFRESH_L10N=1 REFRESH_JS_BUILD=1
+          buildkite-agent artifact download dist.tar.gz .
+          tar -xzf dist.tar.gz
+          rm -rf ./android/Gutenberg/src/main/assets/
+          cp -r ./dist/. ./android/Gutenberg/src/main/assets
           echo "--- :android: Publishing Android Library"
           ./android/gradlew -p ./android :gutenberg:prepareToPublishToS3 $(prepare_to_publish_to_s3_params) :gutenberg:publish
       agents:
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -6,7 +6,11 @@ env:
 
 steps:
     - label: ':react: Build React App'
-      command: make build REFRESH_L10N=1 REFRESH_JS_BUILD=1
+      key: build-react
+      command: |
+          make build REFRESH_L10N=1 REFRESH_JS_BUILD=1 STRICT_L10N=1
+          tar -czf dist.tar.gz dist/
+          buildkite-agent artifact upload dist.tar.gz
```
dcalhoun, line 13: Upload the build artifact for later tasks to avoid unnecessary duplicate builds.

.buildkite/pipeline.yml
```diff
@@ -20,8 +24,12 @@ steps:
       plugins: *plugins
 
     - label: ':android: Publish Android Library'
+      depends_on: build-react
       command: |
-          make build REFRESH_L10N=1 REFRESH_JS_BUILD=1
+          buildkite-agent artifact download dist.tar.gz .
+          tar -xzf dist.tar.gz
+          rm -rf ./android/Gutenberg/src/main/assets/
+          cp -r ./dist/. ./android/Gutenberg/src/main/assets
```
dcalhoun, line 32: Rely upon the single build created by an earlier task.

bin/prep-translations.js
dcalhoun, line 1: Expand the script robustness to include more sophisticated backoff logic when 429 errors occur.

package.json
```diff
@@ -16,7 +16,7 @@
 		"generate-version": "node bin/generate-version.js",
 		"lint:js": "eslint . --ext js,jsx --report-unused-disable-directives --max-warnings 0",
 		"lint:js:fix": "eslint . --ext js,jsx --report-unused-disable-directives --max-warnings 0 --fix",
-		"postinstall": "patch-package && npm run prep-translations && npm run generate-version",
```
dcalhoun, line 19: Remove the default translation downloads that occur in `npm`'s `postinstall` script. 

The downside is that someone running `npm install` rather than the project's documented recommendation of `make dev-server` or `make build` will result in absent translation strings for new clones. 

The upside is that avoiding unnecessary or duplicative translation string fetches is far simpler.

.buildkite/pipeline.yml
```diff
@@ -6,7 +6,11 @@ env:
 
 steps:
     - label: ':react: Build React App'
-      command: make build REFRESH_L10N=1 REFRESH_JS_BUILD=1
+      key: build-react
+      command: |
+          make build REFRESH_L10N=1 REFRESH_JS_BUILD=1 STRICT_L10N=1
```
dcalhoun, line 11: Add `STRICT_L10N=1` to ensure the release build fails if downloading translation strings fails. Otherwise, the release may lack expected translations.

nbradbury (approved):
All good! :shipit:

PR #322, opened 2026-02-16
fix: Upgrade @wordpress/* packages and fix Shortcode block regression

dcalhoun:
## What?

Upgrades all `@wordpress/*` dependencies to the .38 series and fixes a Shortcode block rendering regression caused by the `@wordpress/format-library` 5.36.0 → 5.38.0 bump.

## Why?

The `@wordpress/format-library@5.38.0` upgrade introduced a dependency on `@wordpress/components@^32.0.0` and `@wordpress/block-editor@^15.11.0`, while the project still pinned `@wordpress/components` at `^30.9.0`. This caused:

1. **Duplicate `@wordpress/components` installations** (v30.9.0 at root, v32.0.0 nested), leading to React context mismatches.
2. **esbuild CJS interop breakage**: WordPress packages v15.11+ switched their build output from `.js` to `.mjs`, which triggers Node-style CJS-to-ESM interop in Vite's dependency pre-bundling. This caused `react-autosize-textarea`'s default export (a `forwardRef` component used by `PlainText`) to resolve as a module object instead of the actual component, crashing the Shortcode block with: _"React.jsx: type is invalid -- expected a string or class/function but got: object"_.

## How?

1. **Upgraded all `@wordpress/*` packages** to the .38 series (notably `@wordpress/components` to `^32.0.0` and `@wordpress/block-library` to `^9.38.0`) to eliminate duplicate package installations.
2. **Regenerated existing patch files** for the new package versions (`block-library`, `editor`, `rich-text`). The underlying fixes are unchanged; only target versions and file extensions were updated.
3. **Added a new `PlainText` interop fix** to the `@wordpress/block-editor` patch that unwraps the `react-autosize-textarea` default import when esbuild's `.mjs` interop wraps it as a module object.
4. **Added deep module exports** for `inserter/media-tab/*` and `inserter/hooks/*` to the block-editor patch (moved from the old 15.11.0 patch).

## Testing Instructions

1. Run `npm ci` to install dependencies.
2. Run `npx playwright test` — all 12 E2E tests should pass, including the Shortcode block test.
3. Run `npm run test:unit` — all 72 unit tests should pass.
4. Start the dev server (`npm run dev`), navigate to `/?dev_mode=1`, and insert a Shortcode block — it should render a textarea, not an error boundary.

N/A — no UI changes; this is a dependency upgrade and build fix.


dcalhoun:
Closing in favor of two stacked PRs that separate the minimal format-library fix from the full @wordpress/* package alignment.

PR #326, opened 2026-02-18
Move editor loading into the library

jkmassel:
## What?
Moves the editor loading UI from #316 into its own PR

## Why?
The smaller change is easier to validate.

## How?
Adds loading UI – if you test with the sample app, you'll see that it was removed from the activity and added to the view.


android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -190,37 +208,134 @@ class GutenbergView : WebView {
         this.configuration = configuration
         this.coroutineScope = coroutineScope
 
+        // Initialize the asset loader now that context is available
+        assetLoader = WebViewAssetLoader.Builder()
+            .addPathHandler("/assets/", AssetsPathHandler(context))
+            .build()
```
dcalhoun, line 222: This seems duplicative of `assetLoader` initialization on lines 91-93. Should we remove one?

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -190,37 +208,134 @@ class GutenbergView : WebView {
...
+            layoutParams = LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.WRAP_CONTENT, Gravity.CENTER)
+            visibility = GONE
+        }
+        addView(errorView)
+
         if (dependencies != null) {
             this.dependencies = dependencies
 
             // FAST PATH: Dependencies were provided - load immediately
+            showSpinnerPhase()
```
dcalhoun, line 256: This may be duplicative since `showSpinnerPhase()` is invoked within `loadEditor()`. Remove it?

android/Gutenberg/src/main/java/org/wordpress/gutenberg/views/EditorErrorView.kt
```diff
@@ -0,0 +1,84 @@
...
+        // Create title
+        titleText = TextView(context).apply {
+            layoutParams = LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.WRAP_CONTENT).apply {
+                topMargin = dpToPx(16)
+                marginStart = dpToPx(16)
+                marginEnd = dpToPx(16)
+            }
+            gravity = Gravity.CENTER
+            TextViewCompat.setTextAppearance(this, android.R.style.TextAppearance_Material_Subhead)
+            text = "Failed to load editor"
```
dcalhoun, line 53: For a future PR, we should likely localize strings like this in a manner similar to iOS. 

https://github.com/wordpress-mobile/GutenbergKit/blob/5b66f8ec33edc49004dbd27b59dfd53279acdf38/ios/Sources/GutenbergKit/Sources/EditorLocalization.swift#L33-L56

dcalhoun (approved):
Tested well for me. I left a few suggestions, but non are blocking. Rebasing this atop the latest trunk branch should resolve the CI failures.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
dcalhoun: This remains relevant. Likely duplicative of lines 537-541. We can likely remove these lines 219-222.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
dcalhoun: This remains relevant as a future improvement opportunity.

dcalhoun (approved):
Latest iteration tested well. Let's merge! 🚀

PR #328, opened 2026-02-19
feat: Add local WordPress environment powered by wp-env

dcalhoun:
## What?

Adds a local WordPress development environment powered by `@wordpress/env` for testing the full editor experience with theme styles, media uploads, and plugin block assets.

## Why?

Close CMM-1265.

Testing the editor against a real WordPress instance is essential for validating theme styles, media workflows, and plugin block rendering. Previously, developers had to set up WordPress manually. This provides a one-command local environment that works with both iOS and Android demo apps.

## How?

- **wp-env infrastructure**: `.wp-env.json` config with Twenty Twenty-Five theme, Jetpack blocks plugin, and a CORS mu-plugin for cross-origin API access
- **Setup script** (`bin/wp-env-setup.sh`): Provisions application passwords and outputs credentials for the demo apps
- **Makefile targets**: `wp-env-start`, `wp-env-stop`, `wp-env-clean`, `wp-env-logs`, `wp-env-cli`
- **Demo app integration**: Both iOS and Android demo apps include a "Local WordPress" option that connects to the wp-env site with pre-configured credentials
- **Android mixed content fix**: `WebViewAssetLoader` uses `setHttpAllowed(true)` when the site is HTTP, so the editor page scheme matches and avoids mixed content errors
- **CORS mu-plugin**: Allows `appassets.androidplatform.net` origins (both HTTP and HTTPS) for Android production builds

## Testing Instructions

> [!tip]
> Using `wp-env` requires a running Docker. Tools like [Docker Desktop](https://docs.docker.com/desktop/) (paid license) and [Colima](https://colima.run) (free) provide tools for to managing this. 

1. Start the local WordPress environment:
   ```bash
   make wp-env-start
   ```
2. Configure the site URL for the Android emulator:
   ```bash
   make wp-env-cli CMD="config set WP_SITEURL http://10.0.2.2:8888"
   make wp-env-cli CMD="config set WP_HOME http://10.0.2.2:8888"
   ```
3. Build and run the Android demo app (production build, no dev server)
4. Select "Local WordPress" from the editor list
5. Verify: no mixed content or CORS errors in logcat, theme fonts and styles load correctly
6. Reset the site URL back to localhost for iOS testing:
   ```bash
   make wp-env-cli CMD="config set WP_SITEURL http://localhost:8888"
   make wp-env-cli CMD="config set WP_HOME http://localhost:8888"
   ```
7. Build and run the iOS demo app
8. Select "Local WordPress" from the editor list
9. Verify: editor loads correctly with theme styles and no errors
10. Verify HTTPS sites (e.g., WordPress.com) still work without regression on both platforms

N/A — no user-facing UI changes beyond adding a new list item in the demo apps.

## Screenshots or screencast

iOS Before | iOS After 
-- | --
ios-before | demo-ios

Android Before | Android After 
-- | --
android-before | demo-android

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -96,6 +96,29 @@ make local-android-library
 make test-android
 ```
 
+### Local WordPress Environment (wp-env)
+
+A local WordPress environment powered by `@wordpress/env` for testing the full editor experience with theme styles, media uploads, and plugin block assets.
+
+```bash
+# Start the local WordPress environment (requires Docker)
+make wp-env-start
+
+# Stop the environment (preserves data)
+make wp-env-stop
+
+# Destroy the environment and remove all data
+make wp-env-clean
+
+# View WordPress logs
+make wp-env-logs
+
+# Run a WP-CLI command
+make wp-env-cli CMD="post list"
+```
+
+See `docs/code/local-wordpress.md` for detailed setup instructions and troubleshooting.
+
 > **Note:** Most `make` targets have equivalent `npm` scripts in `package.json`. Build targets accept `REFRESH_DEPS=1`, `REFRESH_L10N=1`, and `REFRESH_JS_BUILD=1` flags to force refresh of dependencies, translations, and JavaScript builds respectively. Run `make help` for full details.
 
 ## Architecture
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,32 @@ preview: npm-dependencies ## Preview the production build locally
 	npm run preview
 
 ################################################################################
+# Local WordPress Environment Targets (wp-env)
+################################################################################
+
+.PHONY: wp-env-start
+wp-env-start: npm-dependencies ## Start the local WordPress environment (requires Docker)
+	npx wp-env start
+	@bash bin/wp-env-setup.sh
+
+.PHONY: wp-env-stop
+wp-env-stop: ## Stop the local WordPress environment
+	npx wp-env stop
+
+.PHONY: wp-env-clean
+wp-env-clean: ## Stop wp-env and remove all data (fresh start)
+	npx wp-env destroy
+	@rm -f .wp-env.credentials.json
+
+.PHONY: wp-env-logs
+wp-env-logs: ## Show wp-env WordPress logs
+	npx wp-env logs
+
+.PHONY: wp-env-cli
+wp-env-cli: ## Run a WP-CLI command in wp-env (usage: make wp-env-cli CMD="post list")
+	npx wp-env run cli wp $(CMD)
+
+################################################################################
 # Code Quality Targets
 ################################################################################
 
--- a/Makefile
+++ b/Makefile
@@ -140,9 +140,9 @@ preview: npm-dependencies ## Preview the production build locally
 ################################################################################
 
 .PHONY: wp-env-start
-wp-env-start: npm-dependencies ## Start the local WordPress environment (requires Docker)
+wp-env-start: npm-dependencies ## Start the local WordPress environment (requires Docker; RESET=1 to regenerate credentials)
 	npm run wp-env start
-	@bash bin/wp-env-setup.sh
+	@RESET=$(RESET) bash bin/wp-env-setup.sh
 
 .PHONY: wp-env-stop
 wp-env-stop: ## Stop the local WordPress environment
--- a/Makefile
+++ b/Makefile
@@ -141,25 +141,25 @@ preview: npm-dependencies ## Preview the production build locally
 
 .PHONY: wp-env-start
 wp-env-start: npm-dependencies ## Start the local WordPress environment (requires Docker)
-	npx wp-env start
+	npm run wp-env start
 	@bash bin/wp-env-setup.sh
 
 .PHONY: wp-env-stop
 wp-env-stop: ## Stop the local WordPress environment
-	npx wp-env stop
+	npm run wp-env stop
 
 .PHONY: wp-env-clean
 wp-env-clean: ## Stop wp-env and remove all data (fresh start)
-	npx wp-env destroy
+	npm run wp-env destroy
 	@rm -f .wp-env.credentials.json
 
 .PHONY: wp-env-logs
 wp-env-logs: ## Show wp-env WordPress logs
-	npx wp-env logs
+	npm run wp-env logs
 
 .PHONY: wp-env-cli
 wp-env-cli: ## Run a WP-CLI command in wp-env (usage: make wp-env-cli CMD="post list")
-	npx wp-env run cli wp $(CMD)
+	npm run wp-env run cli wp $(CMD)
 
 ################################################################################
 # Code Quality Targets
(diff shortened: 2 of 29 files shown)


kean:
This is looking really nice.

android/app/src/main/java/com/example/gutenbergkit/SiteCapabilitiesDiscovery.kt
```diff
@@ -60,13 +63,56 @@ class SiteCapabilitiesDiscovery {
...
+            }
+        }
+    }
+
+    /**
+     * Discovers capabilities by directly fetching the REST API root and inspecting
+     * the `routes` object. This works over plain HTTP, unlike WpLoginClient which
+     * may require HTTPS.
+     */
+    private fun discoverCapabilitiesViaHttp(siteApiRoot: String): SiteCapabilities {
```
dcalhoun, line 86: Added to circumvent errors related to non-TLS connections when connecting to the local wp-env server. Maybe wordpress-rs should allow this? Maybe not? 

```
uniffi.wp_api.InternalException: UnexpectedUniFFICallbackError(reason: "javax.net.ssl.SSLException: Unable to parse TLS packet header")
```

android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorAssetBundle.kt
```diff
@@ -117,6 +117,10 @@ data class EditorAssetBundle(
      * @return `true` if the asset has been cached locally.
      */
     fun hasAssetData(url: String): Boolean {
+        if(this == empty) {
+            return false
+        }
+
```
dcalhoun, line 123: Cherry picked from #326 to prevent empty bundle load error.

AGENTS.md
```diff
@@ -96,6 +96,29 @@ make local-android-library
 make test-android
 ```
 
+### Local WordPress Environment (wp-env)
```
kean, line 99: Are agents expected to run these commands for setting up docker and wp-env? If not, it probably doesn't belong to `AGENTS.md`.

Makefile
```diff
@@ -135,6 +135,32 @@ dev-tools: npm-dependencies ## Start the React Developer Tools
...
+# Local WordPress Environment Targets (wp-env)
+################################################################################
+
+.PHONY: wp-env-start
+wp-env-start: npm-dependencies ## Start the local WordPress environment (requires Docker; RESET=1 to regenerate credentials)
+	npm run wp-env start
+	@RESET=$(RESET) bash bin/wp-env-setup.sh
+
+.PHONY: wp-env-stop
+wp-env-stop: ## Stop the local WordPress environment
```
kean, line 148: It might be worth calling `wp-env` directly. You could then use the existing commands and rely on the existing documentation for it: https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/.

kean (approved):
Hey,I got it to work, and the changes look good.

Just to confirm, `wordpress/env` is a developmemnt-only dependency, correct?

I got a couple of more notes:

- It required Docker. It's not mentioned in the PR description, but is documented in the new file in docs. All I needed to do was to run it. It then downloaded and ran the image and set up the environment fully automatically.
- (dev experience) If you don't run a dev server, the editor shows an infinite spinner. 
- (suggestion) It might be worth revisiting the AGENTS.md file, especially in the light of the recent discoveries about how it affects agents performance. I'd suggest moving the basic instructions needed for developers back to README.md and maybe also factoring out some of the coding guidelines and architecture docs.

AGENTS.md
dcalhoun: Long term? Yes, to enable "E2E verification" that allows agents to verify their own work. 

Today? Since we do not have an MCP for navigating the product in a simulator/emulator, E2E verification is not really possible. So, there is not really a context where an agent would use these today; it's more relevant to simply use the E2E test commands, which themselves set up the `wp-env` environment. 

I do not believe we should remove this at this time. WDYT?

Makefile
dcalhoun: Do you mean run `npm run wp-env` rather than having bespoke `make` targets for interfacing with `wp-env`? If so, yes, the inability to easily pass `npm` script arguments is a downside of the `make` target approach. 

That said, my approach so far with the Makefile is creating a simple interface for the project's common commands. One can always run and access the lower-level commands—`npm`, `swift`, `kotlin`—as necessary or desired. 

If you find this approach confusing, I'm happy to revisit this, but I suggest we do so holistically in a separate pull request—potentially removing any/all `make` targets deemed superfluous or confusing. 

WDYT?

dcalhoun (commented):
@kean thank you for reviewing and testing. 

> Just to confirm, `wordpress/env` is a developmemnt-only dependency, correct?

Yes, it is [listed beneath](https://github.com/wordpress-mobile/GutenbergKit/blob/8d56cc4e6c34783eae43d88e8a888eb07fdec117/package.json#L113) `devDependencies` in the `package.json`. 

> It required Docker. It's not mentioned in the PR description, but is documented in the new file in docs. All I needed to do was to run it. It then downloaded and ran the image and set up the environment fully automatically.

Apologies for not noting it in the PR description. I do not interpret this as actionable, but let me know if I am mistaken. **Update:** I added a note the PR description. 

> (dev experience) If you don't run a dev server, the editor shows an infinite spinner.

This is somewhat separate from the `wp-env` environment. In the future, we could explore adding page load error handlers to the WebView to better surface the inability to load the targeted web page (i.e., the editor HTML). 

> (suggestion) It might be worth revisiting the AGENTS.md file, especially in the light of the recent discoveries about how it affects agents performance. I'd suggest moving the basic instructions needed for developers back to README.md and maybe also factoring out some of the coding guidelines and architecture docs.

I'd love to hear more about your ideas. Given this feels a bit off topic from the focus of this PR, maybe we continue in chat.

android/app/src/main/java/com/example/gutenbergkit/SiteCapabilitiesDiscovery.kt
jkmassel: WPRS should already support this, but you might need https://github.com/Automattic/wordpress-rs/blob/trunk/scripts/external-test-plugins/force-application-passwords.php to make this work locally over HTTP.

jkmassel: Trying to connect using HTTP isn't working (I think) because the server isn't actually speaking HTTPS.

We should display a nice error for this though. Filed https://github.com/Automattic/wordpress-rs/issues/1192 to track.

dcalhoun: @jkmassel to clarify, I interpret your last message means this is not solvable on the GutenbergKit side at this time. Please correct me if I misinterpreted.  

I tested adding a `mu-plugin` resembling the one you referenced, but the same TLS connection error occurred.

Makefile
kean: That sounds good.

adalpari (approved):
LGTM! 🚢 it!

PR #332, opened 2026-02-24
fix: Resolve Cover block inner UI inaccessible in GutenbergKit

dcalhoun:
## What?

Fixes the Cover block's inner UI being inaccessible after insertion, and applies a consistent maximum height to toolbar buttons.

## Why?

After inserting a Cover block, it was not possible to interact with the block's inner UI (color selection, media attachment, resizing). The drag handle was also erroneously positioned at the top of the block instead of the bottom.

The root cause was that block editor styles (`@wordpress/block-library/build-style/editor.css`) were being injected as scoped inline styles inside the editor iframe rather than as global styles. This caused certain CSS rules — particularly those for the Cover block — to be incorrectly scoped, breaking the block's interactive UI.

Additionally, toolbar buttons with text (non-icon-only) were not constrained to the same maximum height as icon buttons.

Fixes: CMM-1281

## How?

1. **Move block editor styles to global scope** (`src/utils/editor-styles.js`): Import `@wordpress/block-library/build-style/editor.css` as a global style instead of an inline scoped style within the visual editor iframe. This ensures editor styles apply correctly without scoping issues.

2. **Remove scoped inline import** (`src/components/visual-editor/index.jsx`): Remove the `blocksEditorStyles` inline import and its usage in `useEditorStyles`, since it is now loaded globally.

3. **Broaden toolbar button max-height selector** (`src/components/editor-toolbar/style.scss`): Change the selector from `.components-button.has-icon.has-icon` to `.components-button.components-button` so that all toolbar buttons (not just icon buttons) get the consistent `max-height` constraint.

4. **Add Cover block E2E test** (`e2e/cover-block.spec.js`): Add an end-to-end test that inserts a Cover block, selects a background color from the overlay color palette, and verifies the block transitions to its full state with the correct inner block structure. This test exercises the CSS scoping fix — the color button click fails if the resizable popover intercepts pointer events due to incorrectly scoped styles.

5. **Simplify local E2E testing** (`playwright.config.js`): On CI, E2E tests run against the production build (`npm run preview`). Locally, tests use the Vite dev server (`npm run dev`) so changes can be tested without a full rebuild.

## Testing Instructions

1. Open a post or page.
2. Insert a Cover block.
3. Verify you can interact with the block's inner UI — select colors, attach media, and resize the block.
4. Verify the drag handle appears at the bottom of the block (not the top).
5. Check that toolbar buttons (both icon-only and text) have a consistent height.

N/A, no navigation changes. 

## Screenshots or screencast 

Cover Before | Cover After
-- | --
cover-before | cover-after

<details><summary>Toolbar Before</summary>
<p>

https://github.com/user-attachments/assets/a0af9c73-58ba-46ac-a6ea-250c67325332

</p>
</details> 

<details><summary>Toolbar After</summary>
<p>

https://github.com/user-attachments/assets/eca9a8af-076a-433a-8e5b-afbd13162fcb

</p>
</details>

### DIFF
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -3,16 +3,21 @@
  */
 import { defineConfig, devices } from '@playwright/test';
 
+const isCI = process.env.CI === 'true';
+
+// CI uses production build, local development uses dev server
+const url = isCI ? 'http://localhost:4173' : 'http://localhost:5173';
+
 export default defineConfig( {
 	testDir: './e2e',
 	outputDir: './e2e/test-results',
 	fullyParallel: true,
-	workers: process.env.CI ? 1 : undefined,
-	retries: process.env.CI ? 2 : 0,
+	workers: isCI ? 1 : undefined,
+	retries: isCI ? 2 : 0,
 	timeout: 60_000,
-	reporter: process.env.CI ? 'list' : 'html',
+	reporter: isCI ? 'list' : 'html',
 	use: {
-		baseURL: 'http://localhost:4173',
+		baseURL: url,
 		actionTimeout: 10_000,
 		trace: 'on-first-retry',
 	},
@@ -23,9 +28,10 @@ export default defineConfig( {
 		},
 	],
 	webServer: {
-		command: 'npm run preview',
-		url: 'http://localhost:4173',
-		reuseExistingServer: true,
+		// CI uses production build, local development uses dev server
+		command: isCI ? 'npm run preview' : 'npm run dev',
+		url,
+		reuseExistingServer: ! isCI,
 		timeout: 30_000,
 	},
 } );
--- a/src/components/editor-toolbar/style.scss
+++ b/src/components/editor-toolbar/style.scss
@@ -105,7 +105,7 @@ $scroll-indicator-elevation: 32;
 }
 
 .gutenberg-kit-editor-toolbar.gutenberg-kit-editor-toolbar
-	.components-button.has-icon.has-icon {
+	.components-button.components-button {
 	min-width: $min-touch-target-size;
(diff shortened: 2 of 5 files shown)


src/components/editor-toolbar/style.scss
```diff
@@ -105,7 +105,7 @@ $scroll-indicator-elevation: 32;
 }
 
 .gutenberg-kit-editor-toolbar.gutenberg-kit-editor-toolbar
-	.components-button.has-icon.has-icon {
+	.components-button.components-button {
```
dcalhoun, line 108: Fix height discrepancies between buttons with and without icons. Now the button heights are consistent, preventing shifting occurring the in toolbar when moving selection between blocks.

src/components/visual-editor/index.jsx
```diff
@@ -94,8 +93,7 @@ function VisualEditor( { hideTitle } ) {
 		hasThemeStyles ? '' : defaultThemeStyles,
 		componentStyles,
 		blockEditorContentStyles,
-		blocksStyles,
-		blocksEditorStyles
```
dcalhoun, line 98: Hoisting the `@wordpress/block-library` editor styles to the root of the web app ensures they are not erroneously wrapped with a `:where(.editor-styles-wrapper)` selector, which prevented the styles from being correctly applied to the UI elements.

playwright.config.js
dcalhoun, line 1: Having the CI run E2E test against the production build helps avoid instability from differences between build outputs, but using the dev server for local E2E runs simplifies testing new changes.

nbradbury (approved):
Looks good! :shipit:

PR #336, opened 2026-02-25
build: reinstate nested qs dependency

dcalhoun:
## What?

1. Reinstates the nested `qs@6.14.2` dependency entries in `package-lock.json` for `body-parser` and `express`.
2. Fixes iOS UI tests that fail because the editor list row was renamed from "Default Editor" to "Standalone Editor".

## Why?

1. The `qs` dependency is required by `@wordpress/wp-env` and its transitive dependencies. It was introduced in e6cec57 but was erroneously discarded during a rebase in d12b489. Its absence causes CI failures due to an outdated lock file.
2. The row rename happened in e6cec57f but the UI test helper (`EditorUITestHelpers.swift`) was not updated, causing all tests that use `navigateToEditor` to fail (e.g. `testInsertImageBlock`).

## How?

1. Adds back the two nested `node_modules/body-parser/node_modules/qs` and `node_modules/express/node_modules/qs` entries to `package-lock.json`.
2. Updates the UI test helper to look for "Standalone Editor" instead of "Default Editor".

## Testing Instructions

1. Run `npm ci` and verify it completes without errors.
3. Verify CI passes.

### DIFF
--- a/ios/Demo-iOS/GutenbergUITests/EditorUITestHelpers.swift
+++ b/ios/Demo-iOS/GutenbergUITests/EditorUITestHelpers.swift
@@ -11,10 +11,10 @@ enum EditorUITestHelpers {
     /// the editor has loaded.
     @discardableResult
     static func navigateToEditor(app: XCUIApplication) throws -> XCUIElement {
-        // Tap the "Default Editor" row in the list.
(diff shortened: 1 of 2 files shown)


adalpari (approved):
Good catch! LGTM!

PR #339, opened 2026-02-25
chore: add WordPress.com OAuth support to iOS and Android demo apps

jkmassel:
## What?

Add WordPress.com OAuth2 authentication support to both the Android and iOS demo apps, alongside the existing Application Passwords flow for self-hosted sites.

## Why?

The demo apps previously only supported self-hosted WordPress sites via Application Passwords. This made testing WordPress.com sites an integration concerns in other apps, which means it's not getting tested as much as it needs to. This PR adds OAuth2 support so the apps can also connect to WordPress.com sites, and upgrades credential storage to use encrypted repositories on both platforms.

### Android
- Branch on `DiscoveredAuthenticationMechanism` to launch the appropriate auth flow (Application Passwords or OAuth2)
- Store WP.com accounts as `Account.WpCom` so auth type is determined by pattern matching
- Replace `SharedPreferences` with encrypted `AccountRepository` from wordpress-rs, backed by Android Keystore
- Add WP.com site namespace handling in `SitePreparationViewModel`

### iOS
- Add `AuthenticationManager` with async/throws API supporting both Application Passwords and OAuth2 flows
- Replace UserDefaults-based `ConfigurationStorage` with encrypted `AccountRepository` from wordpress-rs (`SecureEnclavePasswordTransformer`)
- Use `WpComDotOrgApiUrlResolver` for correct WP.com API URL routing
- Propagate errors from `ConfigurationStorage` instead of swallowing them

### Shared (JS/Web)
- Replace fetch-based editor asset loading with config injection via `window.GBKit.editorAssets`. (We have the data already, so why download it again?)
- Guard against duplicate site namespace insertion in the API path modifier middleware
- Move shared WP.com OAuth credentials file to project root, referenced by both platforms

## Testing Instructions

1. **Self-hosted (Android or iOS)**: Tap "Add WordPress site", enter a self-hosted site URL → Application Passwords flow → verify editor loads
2. **WordPress.com (Android or iOS)**: Copy `wp_com_oauth_credentials.json.example` to `wp_com_oauth_credentials.json` in the project root, fill in client_id/secret → enter a WP.com site URL → OAuth browser flow → verify editor loads with plugins and theme styles
3. Verify sites persist across app restarts
4. Verify site deletion works (iOS: swipe to delete, Android: long-press)
5. In a WP.com site – try adding an AI block and asking it to make you a haiku. Do the same with a Jetpack site. This validates third-party plugin support and the ability for the editor to make authenticated requests to the remote site.
6. In a self-hosted site (you can use `vanilla.wpmt.co`) load the editor and ensure that theme styles are loaded. This validates remote theme style loading.


android/gradle/libs.versions.toml
```diff
@@ -17,7 +16,7 @@ mockito = "4.1.0"
 robolectric = "4.14.1"
 kotlinx-coroutines = '1.10.2'
 androidx-recyclerview = '1.3.2'
-wordpress-rs = 'trunk-d02efa6d4d56bc5b44dd2191e837163f9fa27095'
+wordpress-rs = '1190-c2b404d9c9754b229967386fa7460d65fe87a29d'
```
jkmassel: This will need to be updated to a `trunk` revision prior to merging

android/Gutenberg/src/main/java/org/wordpress/gutenberg/RESTAPIRepository.kt
```diff
@@ -86,7 +87,7 @@ class RESTAPIRepository(
      * @return The parsed editor settings.
      */
     suspend fun fetchEditorSettings(): EditorSettings {
-        if (!configuration.themeStyles) {
+        if (!configuration.plugins && !configuration.themeStyles) {
```
dcalhoun: This was explicitly removed in https://github.com/wordpress-mobile/GutenbergKit/commit/ba7c9c295ce868a85710a5663138d2aaab49e8fa, as it led to erroneous, error-producing requests for a non-existing endpoint. We should likely revert this unless there an explicit reason to reintroduce this.

e2e/editor-error.spec.js
```diff
@@ -96,15 +94,15 @@ test.describe( 'Editor Error Handling', () => {
...
 			page.getByText(
 				'Loading plugins failed, using default editor configuration.'
 			)
-		).toBeVisible( { timeout: 10_000 } );
-
-		// Editor should still be functional despite the plugin failure.
-		await expect(
-			page.locator( '.gutenberg-kit-visual-editor' )
-		).toBeVisible();
+		).toBeHidden();
```
dcalhoun: This _might_ be a false-positive given it searches for something to be hidden. Did it ever render at all? If not and that is expected, we should merely remove this assertion.

e2e/editor-page.js
```diff
@@ -49,10 +53,14 @@ export default class EditorPage {
...
 	async setup( gbkit = {} ) {
 		const editorSettings = await getEditorSettings();
+		const editorAssets = gbkit.plugins
+			? await getEditorAssets()
+			: undefined;
 
 		const config = {
 			...DEFAULT_GBKIT,
 			editorSettings,
+			...( editorAssets && { editorAssets } ),
```
dcalhoun: I imagine this could be simplified to align with `editorSettings`. 
```suggestion
			editorAssets,
```

ios/Demo-iOS/Sources/Views/SitePreparationView.swift
```diff
@@ -57,11 +57,6 @@ struct SitePreparationView: View {
                 Toggle("Enable Native Inserter", isOn: $viewModel.enableNativeInserter)
                 Toggle("Enable Network Logging", isOn: $viewModel.enableNetworkLogging)
 
-                Picker("Network Fallback", selection: $viewModel.networkFallbackMode) {
```
dcalhoun, line 60: Given removing this isn't referenced or explained in the PR description, I presume this was an erroneous merge conflict resolution. We should reinstate this change from https://github.com/wordpress-mobile/GutenbergKit/commit/f5c91cce5acc9a78351fafd8ab639f0f4a1c0c0f that improves the offline editor experience.

src/utils/bridge.js
```diff
@@ -4,7 +4,6 @@
 import parseException from './exception-parser';
 import { debug, error } from './logger';
 import { isDevMode } from './dev-mode';
-import { basicFetch } from './fetch';
```
dcalhoun, line 7: This was the only usage of `basicFetch`. We should remove that module entirely.

.gitignore
```diff
@@ -194,8 +195,7 @@ local.properties
 # /ios/Sources/GutenbergKit/Gutenberg/index.html
 
 # Translation files
-src/translations/*
-!src/translations/.gitkeep
```
dcalhoun, line 201: Presuming this is a bad merge conflict resolution. We should reinstate this change from https://github.com/wordpress-mobile/GutenbergKit/commit/3079431469e90cd98da9277d3054a41496e7b3e8. This ensures the translations directory always exists to avoid Node.js module load errors.

.gitignore
```diff
@@ -205,7 +205,6 @@ playwright-report/
 .wp-env.credentials.json
 wp-env/mu-plugins/*
 !wp-env/mu-plugins/gutenbergkit-cors.php
-!wp-env/mu-plugins/gutenbergkit-jetpack-blocks.php
```
dcalhoun, line 211: Presuming this is a bad merge conflict resolution. We should reinstate this change from https://github.com/wordpress-mobile/GutenbergKit/commit/aa8eaf1fb19d037a71fc72f351acd70b132b7afb. It ensure our custom MU plugin is tracked, while others are excluded.

wp_com_oauth_credentials.json.example
dcalhoun, line 1: We should add project documentation linking to relevant documentation for creating WordPress.com Apps for [OAuth](https://developer.wordpress.com/docs/api/oauth2/) and noting the required Redirect URL value.

dcalhoun (commented):
The implementation tested well for me. 🎉 

Code looks good overall, but I left several inline notes we should address before merging.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/RESTAPIRepository.kt
dcalhoun: Addressed in https://github.com/wordpress-mobile/GutenbergKit/pull/400.

e2e/editor-error.spec.js
dcalhoun: Addressed in https://github.com/wordpress-mobile/GutenbergKit/pull/400.

e2e/editor-page.js
dcalhoun: Addressed in https://github.com/wordpress-mobile/GutenbergKit/pull/400.

ios/Demo-iOS/Sources/Views/SitePreparationView.swift
dcalhoun: Addressed in https://github.com/wordpress-mobile/GutenbergKit/pull/400.

src/utils/bridge.js
dcalhoun: Addressed in https://github.com/wordpress-mobile/GutenbergKit/pull/400.

.gitignore
dcalhoun: Addressed in https://github.com/wordpress-mobile/GutenbergKit/pull/400.

.gitignore
dcalhoun: Addressed in https://github.com/wordpress-mobile/GutenbergKit/pull/400.

wp_com_oauth_credentials.json.example
dcalhoun: Addressed in https://github.com/wordpress-mobile/GutenbergKit/pull/400.

dcalhoun (approved):
The latest iteration tested well for me on both iOS and Android. 🎉 

All feedback/issues were addressed in https://github.com/wordpress-mobile/GutenbergKit/pull/400 or subsequent commits.

PR #351, opened 2026-03-04
fix: resolve failures from @wordpress/edit-post bump (#346)

dcalhoun:
## What?

Fixes issues introduced by the Dependabot PR #346 (`@wordpress/edit-post` 8.39.0 → 8.40.1), which was merged with failing E2E tests.

## Why?

PR #346 bumped `@wordpress/edit-post`, which transitively bumped `@wordpress/block-library` from 9.39.0 to 9.41.0. This upstream change migrated the Button block's Width control from a custom `InspectorControls` panel to the standard `dimensions.width` block support ([WordPress/gutenberg#74242](https://github.com/WordPress/gutenberg/pull/74242)). As a result:

1. **Missing inspector tabs**: The Button block no longer has fills in the default `InspectorControls` slot (Settings tab), leaving only the Styles tab. With a single tab, Gutenberg renders a flat layout instead of tabbed Settings/Styles panels. This is the intended upstream behavior.
2. **Broken fieldset styles**: The fieldset reset styles were scoped to `.gutenberg-kit-visual-editor`, but the block settings panel renders in a `Popover` outside that element, so the styles were not applied.

## How?

1. **E2E test fix**: Removed the `Styles` tab click from the `color-gradient` spec since the Button block inspector no longer renders tabs.
2. **Fieldset style fix**: Moved the fieldset reset styles from `.gutenberg-kit-visual-editor` to the `.gutenberg-kit` root scope so they apply globally, including to popovers.

## Testing Instructions

Verify the CI checks pass. 

N/A — no UI changes; the flat inspector layout is the upstream Gutenberg behavior.

## Screenshots or screencast 

Before | After
-- | -- 
before-1 | after-1
before-2 | after-2


nbradbury (approved):
> Verify the CI checks pass.

CI is passing so this is good to go :shipit:

PR #353, opened 2026-03-05
fix: skip editor settings fetch when theme styles are disabled

dcalhoun:
## What?

Skip the `/wp-block-editor/v1/settings` fetch when `shouldUseThemeStyles` is disabled, regardless of the `shouldUsePlugins` flag.

## Why?

Fix CMM-1921. Fix CMM-1163. 

The guard in `fetchEditorSettings()` previously required **both** `shouldUsePlugins` and `shouldUseThemeStyles` to be `false` before skipping the fetch. This meant a site with plugins enabled but no support for the block editor settings endpoint would still attempt the fetch, resulting in a 404 error that could cause the editor to fail to load.

The editor settings endpoint is only relevant for theme styles — plugins have their own separate asset endpoint (`/wpcom/v2/editor-assets`).

## How?

Changed the guard condition in `fetchEditorSettings()` from:
```
if !shouldUsePlugins && !shouldUseThemeStyles → skip
```
to:
```
if !shouldUseThemeStyles → skip
```

Applied to both iOS (`RESTAPIRepository.swift`) and Android (`RESTAPIRepository.kt`), with updated and additional test coverage on both platforms.

## Testing Instructions

1. Configure a site that supports plugins (`/wpcom/v2/editor-assets` route exists) but does **not** support the block editor settings endpoint (`/wp-block-editor/v1/settings` route is absent).
2. Open the editor for that site.
3. Verify the editor loads without a 404 error for the settings endpoint.
4. Verify that a site with both plugins and theme styles enabled still fetches and applies editor settings correctly.

N/A — no UI changes.

## Screenshots or screencast

N/A — no UI changes.

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/RESTAPIRepository.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/RESTAPIRepository.kt
@@ -86,7 +86,7 @@ class RESTAPIRepository(
      * @return The parsed editor settings.
      */
     suspend fun fetchEditorSettings(): EditorSettings {
-        if (!configuration.plugins && !configuration.themeStyles) {
+        if (!configuration.themeStyles) {
             return EditorSettings.undefined
         }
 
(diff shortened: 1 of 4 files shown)


jkmassel (approved):
With these changes, I was able to load the site that was erroring before the changes.

PR #356, opened 2026-03-06
fix: resolve patch-package version mismatch for @wordpress/format-library

dcalhoun:
## What?

Regenerates the `@wordpress/format-library` patch file to match the currently installed version and adds `--error-on-warn` to prevent future mismatches from going unnoticed.

## Why?

Running `make npm-dependencies` produces the following warning:

```
Warning: patch-package detected a patch file version mismatch

  Don't worry! This is probably fine. The patch was still applied
  successfully. Here's the deets:

  Patch file created for

    @wordpress/format-library@5.38.0

  applied to

    @wordpress/format-library@5.41.0

  At path

    node_modules/@wordpress/format-library

  This warning is just to give you a heads-up. There is a small chance of
  breakage even though the patch was applied successfully. Make sure the package
  still behaves like you expect (you wrote tests, right?) and then run

    patch-package @wordpress/format-library

  to update the version in the patch file name and make this warning go away.
```

This was introduced by #349 (and subsequent dependency bumps) which upgraded the package without regenerating the patch file.

## How?

1. Regenerated the patch file for `@wordpress/format-library@5.41.0` using `npx patch-package @wordpress/format-library --exclude='^$'` (the `--exclude` flag is needed because the patch modifies `package.json`, which patch-package excludes by default).
2. Added `--error-on-warn` to the `postinstall` script so that patch version mismatches cause a hard failure instead of a silent warning. This will cause CI (and local installs) to fail if a dependabot PR bumps a patched package without regenerating the patch.
3. Documented the `--exclude='^$'` caveat in `AGENTS.md` for future reference.

## Testing Instructions

Verify CI checks pass.


nbradbury (approved):
CI passes :shipit:

PR #357, opened 2026-03-09
feat: proxy media uploads through native delegate for processing

dcalhoun:
## What?

Adds a native media upload pipeline that routes file uploads through a local HTTP server on iOS and Android, enabling the host app to process files (e.g., resize images, transcode video) before they are uploaded to WordPress.

## Why?

Ref CMM-1249.

Gutenberg's built-in upload path sends files directly from the WebView to the WordPress REST API with no opportunity for native processing. Host apps need to resize images, enforce upload size limits, or apply other transformations before upload. This pipeline gives the native layer full control over media processing while keeping the existing Gutenberg upload UX (blob previews, save locking, entity caching) unchanged.

## How?

**Architecture:** A localhost HTTP server runs on each platform, built on the `GutenbergKitHTTP` library (iOS) and `HttpServer` (Android) from #367. The library handles TCP binding, HTTP/1.1 parsing, bearer token authentication (`Relay-Authorization`), multipart form-data parsing, connection limits, and disk-backed body buffering. The upload server is a thin handler on top.

**JS layer:**

- `nativeMediaUploadMiddleware` in `api-fetch.js` intercepts `POST /wp/v2/media` requests when `nativeUploadPort` is configured in `window.GBKit`, forwarding the original request body (the file plus every sibling field — `post`, additionalData) and query string (e.g. `?_embed`) to the local server. The native server relays WordPress's response verbatim: on success the middleware returns WordPress's attachment object unchanged (preserving `media_details.sizes`, `_embedded`, distinct raw/rendered fields, etc.), so the existing Gutenberg upload pipeline works unchanged
- On a non-2xx it mirrors `@wordpress/api-fetch` and rejects with the parsed WordPress error body (`{ code, message, data }`) so `@wordpress/media-utils` surfaces WordPress's real message and code, falling back to an `invalid_json` error on a non-JSON body
- Uses exact endpoint matching (`/wp/v2/media` but not `/wp/v2/media/123` or `/wp/v2/media-categories`) to avoid intercepting non-upload requests

**Native layer:**

- `MediaUploadDelegate` protocol/interface with `processFile` (resize/transcode) and optional `uploadFile` for a custom upload, which returns the raw WordPress response (`MediaUploadResponse?`); returning `nil` falls back to the default uploader
- `DefaultMediaUploader` as fallback, uploading to `/wp/v2/media` via the host's HTTP client, with site API namespace support for namespaced sites. It returns WordPress's raw bytes and status without throwing on non-2xx, so the server can relay the exact response
- Upload server starts whenever a delegate can handle uploads — no resources wasted otherwise. The default uploader is built only when a site API root and auth header are present; a cookie-auth host (empty auth header) with a delegate implementing `uploadFile` still starts the server
- Server-generated errors (413, delegate/parse failures, "no uploader configured") emit a `{ code, message }` JSON body with CORS headers so they normalize like a relayed WordPress error; oversized bodies are drained before the 413 so the WebView receives a clean response rather than a connection reset
- Filename sanitization prevents path traversal from malicious `Content-Disposition` values

**Demo apps:**

- Both iOS and Android demo apps include a delegate that resizes images to 2000px max
- "Enable Native Media Upload" toggle (defaults to true) controls whether the delegate is set

**Reliability & lifecycle:**

- A transport-layer failure reaching the loopback server does **not** fall back to a direct re-upload: retrying a non-idempotent `POST /wp/v2/media` could duplicate an attachment the server had already relayed. Reachability is gated proactively upstream instead (the middleware skips the native path when no port is advertised, and the native side only advertises a reachable port). A caller-initiated cancellation propagates as the cancellation (detected via `signal.aborted`) rather than being retried
- iOS stops the upload server on `deinit` (not `viewDidDisappear`, which also fires when another view controller is presented over the editor) and holds the delegate weakly
- Android re-advertises the server port and token into `window.GBKit` after a restart (e.g. a Compose host constructing a new delegate per recomposition) and uses a 60s upload timeout to match `EditorHTTPClient`, since WordPress generates image sub-sizes synchronously inside `POST /wp/v2/media`

### Key design decisions

- **Localhost HTTP server over platform-specific bridges:** Provides a uniform interception point for all upload paths (file picker, paste, drag-and-drop, programmatic) without requiring per-path bridge wiring. Builds on the cross-platform HTTP library from #367 rather than hand-rolling TCP/HTTP/multipart handling
- **Delegate is opt-in:** The server doesn't start and no configuration is injected unless the host provides a `MediaUploadDelegate` that can handle uploads, keeping the default behavior unchanged
- **`api-fetch` middleware over `mediaUpload` editor setting:** Ideally, media uploads would be handled via the `mediaUpload` editor setting (see the [Gutenberg Framework guides](https://wordpress.org/gutenberg-framework/docs/intro/)), but GutenbergKit uses Gutenberg's `EditorProvider` which [overwrites that setting internally](https://github.com/WordPress/gutenberg/blob/29914e1d09a344edce58d938fa4992e1ec248e41/packages/editor/src/components/provider/use-block-editor-settings.js#L340). Until GutenbergKit is refactored to use `BlockEditorProvider`, the `api-fetch` middleware approach is necessary.

### Alternatives considered

1. **JS Canvas resize + native inserter resize** — Two separate implementations: `createImageBitmap()` + `OffscreenCanvas` in JS for web uploads, `CGImageSourceCreateThumbnailAtIndex` in `MediaFileManager.import()` for the native inserter. Ships fastest and lowest complexity, but Canvas resize quality is lower than native, two codepaths to maintain, and a dead-end for video (client-side transcoding in a WebView is impractical).

2. **Native upload pipeline via local HTTP server (this PR)** — A single `api-fetch` middleware intercepts all `POST /wp/v2/media` requests and routes files through a localhost server for native processing. Covers every upload path (file picker, drag-and-drop, paste, programmatic, plugin blocks) with native-quality processing. Scales to video transcoding. More upfront work than option 1.

3. **Replace `MediaPlaceholder` via `withFilters` hook** — Use `editor.MediaPlaceholder` and `editor.MediaReplaceFlow` filters with `handleUpload={false}` to deliver raw `File` objects to `onSelect`, then route to native. Incomplete coverage: misses block drag-and-drop re-uploads (`handleBlocksDrop` calls `mediaUpload` directly), direct `mediaUpload` calls from plugins, and loses blob previews when `handleUpload` is false. `instanceof FileList` checks are fragile in WebView contexts.

4. **Redirect to native UI on large files** — Keep the web upload button, but show a native dialog when files exceed limits. Awkward UX (user already picked a file, now asked to pick again differently). On iOS, the already-selected JS `File` can't be handed to native for optimization. Two parallel upload paths add complexity.

5. **JS resize for images + hide web upload for video blocks** — JS Canvas resize for images, hide the "Upload" button on video-accepting blocks via `editor.MediaPlaceholder` filter (forcing users to Media Library for video). Users can't drag-and-drop videos, blocks accepting both image and video (Cover) get complicated, and it's a dead-end architecture.

6. **Client-side processing via `@wordpress/upload-media` (WASM libvips)** — Gutenberg's experimental [`@wordpress/upload-media`](https://github.com/WordPress/gutenberg/tree/4da891757d7e32199faca1de025a2d13fcad8206/packages/upload-media) package includes a WASM build of libvips for high-quality client-side image resizing, rotation, format transcoding, and thumbnail generation. Quality is comparable to server-side ImageMagick. However, it requires `SharedArrayBuffer` for WASM threading, which is only available in cross-origin isolated contexts — WKWebView loads GutenbergKit's HTML locally with no HTTP headers, so `SharedArrayBuffer` is unavailable. WebKit also lacks `credentialless` iframe support, meaning cross-origin isolation would break third-party embeds (YouTube, Twitter, etc.). Single-threaded WASM fallback is unvalidated, and the package's memory footprint (50-100MB+ per image) is a concern under iOS jetsam pressure. Not viable today, but worth revisiting if the package decouples its store/queue management from WASM processing (tracked upstream).

A key constraint is **platform asymmetry**: Android can intercept web `<input type="file">` via `onShowFileChooser()`, but iOS cannot — `WKWebView` handles file selection internally. This rules out purely native interception strategies for web-originated uploads and motivated the localhost server approach, which works identically on both platforms.

## Testing Instructions

1. Open the iOS or Android demo app connected to a WordPress site (e.g., wp-env)
2. Verify the "Enable Native Media Upload" toggle is present and defaults to on
3. Insert an Image block and upload a large image (>2000px)
4. Verify the upload succeeds and the image displays in the editor
5. Check logs for "Resized image from WxH to fit 2000px" confirming native processing
6. Toggle "Enable Native Media Upload" off, restart the editor, and upload again — verify the upload still works (via standard Gutenberg path, no resize log)

The toggle follows the same pattern as the existing "Enable Native Inserter" toggle — no new UI beyond that.

## Screenshots or screencast

N/A — backend/infrastructure change with no visible UI changes beyond the demo app toggle.


src/utils/api-fetch.js
```diff
@@ -125,6 +130,117 @@ function filterEndpointsMiddleware( options, next ) {
...
+		},
+		body: formData,
+		signal: options.signal,
+	} )
+		.then( ( response ) => {
+			if ( ! response.ok ) {
+				return response.text().then( ( body ) => {
+					const message =
+						response.status === 413
+							? `The file is too large to upload. Please choose a smaller file.`
```
jkmassel: We should probably print whatever the server sends back – it _should_ be `WP_Error`-shaped, but some hosts might have messaging like "The max is ${SOME_NUMBER}" or "You've reached your quota".

WDYT?

dcalhoun: Yes, that makes sense. 

This was strictly implemented to handle the 250 MB maximum upload restriction of the local server, but I agree it should be made more robust. If we do not add specific handling for 413, the default user-facing message is something like "Unable to get a valid response from the server."

dcalhoun: Addressed in 8c78c1a9ab6ef51ce4b1750f08ab7d591cb1713e and a3ff56cee774b34b8a9aa92b1659c02ffbe00c1a by removing this JS logic entirely.

ios/Sources/GutenbergKit/Sources/Media/MediaUploadServer.swift
```diff
@@ -0,0 +1,383 @@
...
+    private let siteApiNamespace: String?
+
+    init(httpClient: EditorHTTPClientProtocol, siteApiRoot: URL, siteApiNamespace: [String] = []) {
+        self.httpClient = httpClient
+        self.siteApiRoot = siteApiRoot
+        self.siteApiNamespace = siteApiNamespace.first
+    }
+
+    func upload(fileURL: URL, mimeType: String, filename: String) async throws -> MediaUploadResult {
+        let fileData = try Data(contentsOf: fileURL)
```
dcalhoun: **Known issue:** This loads the entire file into memory to build the multipart body. Fine for the image-resize use case, but we may need a streaming approach for videos.

A fix might be to write the multipart body to a temp file on disk and upload via `URLSession.upload(for:fromFile:)`. That requires either adding an `upload(_:fromFile:)` method to `EditorHTTPClientProtocol` or having `DefaultMediaUploader` own a `URLSession` directly. Android already avoids this via OkHttp's `file.asRequestBody()`.

Noted in the PR description's "Known issues" section.

src/utils/api-fetch.js
```diff
@@ -131,6 +136,112 @@ function filterEndpointsMiddleware( options, next ) {
...
+				mime_type: result.mime,
+				media_type: result.type,
+				media_details: {
+					width: result.width || 0,
+					height: result.height || 0,
+				},
+				link: result.url,
+			};
+		} )
+		.catch( ( err ) => {
```
dcalhoun, line 254: **Known issue (non-blocking):** When a file exceeds the HTTP library's `maxRequestBodySize` (4 GB default), the library rejects the TCP connection while the browser is still streaming the request body. The `fetch()` throws a generic `TypeError` ("Load failed" / "Failed to fetch") that lands here — not the structured error from the `.then` branch above.

There's no good workaround at this layer — the server can't reliably send an HTTP 413 response while the client is still writing (HTTP/1.1 limitation).  Uploads rejected by WordPress's own size limits do produce a proper error response that hits the `!response.ok` branch correctly.

Noted in the PR description's "Known issues" section.

dcalhoun:
@jkmassel this now relies upon the GBK HTTP server library. This is ready for another review.

ios/Sources/GutenbergKit/Sources/Media/MediaUploadServer.swift
jkmassel: HTTPServer.Request can partially help us here!  For the passthrough case (ie – we're not touching the file), we can just do:

```swift
// Stream the body directly — no intermediate copy.
if let body = parsed.body {
  urlRequest.setValue("\(body.count)", forHTTPHeaderField: "Content-Length")
  urlRequest.httpBodyStream = try body.makeInputStream()
}
```

This avoids the file using any extra memory at all.

The "we modified the file and now we need to upload that" case is a bit trickier. We could make a custom input stream handler that takes the original request and a reference to the new file and emits a new input stream. That would keep memory usage flat, and disk usage would only be `${SIZE_OF_ORIGINAL_FILE} + ${SIZE_OF_RESIZED_FILE}`. `RequestBody.Storage. fileSlice` has all the data we need for this. It could even be a convenience method on `RequestBody` so you could do something like:

```swift
if let body = parsed.body {
  urlRequest.setValue("\(body.count)", forHTTPHeaderField: "Content-Length")
  
  let newFilePath = doSomeMediaProcessing(of: body)
  
  urlRequest.httpBodyStream = try body.replacingMultipartData(named: "first-file-part", with: newFilePath)
}
```

LMK if this doesn't make sense, or if you'd like a hand implementing `replacingMultipartData`?

src/utils/api-fetch.js
jkmassel: Fun! 

I think we can relax the server a bit here:

> **RFC 9110 §15.5.14 (413 Content Too Large)**
> The 413 (Content Too Large) status code indicates that the server is refusing to process a request because the request content is larger than the server is willing or able to process. The server MAY terminate the request, if the protocol version in use allows it; otherwise, the server MAY close the connection.

Instead of closing the connection immediately, we could just drop the bytes on the floor as they come in, then send the proper `413` response. For even the largest files, this should be _very_ fast and doesn't use any disk or memory resources. WDYT?

dcalhoun: Addressed by #419 and #441.

ios/Sources/GutenbergKit/Sources/Media/MediaUploadServer.swift
dcalhoun: Addressed by #419 and #441.

dcalhoun:
@jkmassel the latest changes tested well for me. I believe this is ready for another review. 🙇🏻‍♂️

android/Gutenberg/src/main/java/org/wordpress/gutenberg/HttpServer.kt
```diff
@@ -341,12 +341,6 @@ class HttpServer(
             // Phase 1: receive headers only.
             readUntil(parser, input, buffer, deadlineNanos) { it.hasHeaders }
 
-            // Drain oversized body before throwing so the
-            // client receives the 413 (RFC 9110 §15.5.14).
-            if (parser.state == HTTPRequestParser.State.DRAINING) {
-                readUntil(parser, input, buffer, deadlineNanos) { it.isComplete }
-            }
```
dcalhoun: Relocated to avoid unauthorized requests causing body drains. Similar iOS changes added around `ios/Sources/GutenbergKitHTTP/HTTPServer.swift:285`.

ios/Sources/GutenbergKitHTTP/HTTPServer.swift
```diff
@@ -120,6 +120,12 @@ public final class HTTPServer: Sendable {
     /// If no data arrives within this interval, the connection is closed with a 408 response.
     public static let defaultIdleTimeout: Duration = .seconds(5)
 
+    /// The default maximum time to wait for the listener to become ready (5 seconds).
+    /// Binding to loopback normally completes in milliseconds; the bound exists so a
+    /// listener stuck in the `.waiting` state (which emits no further updates)
+    /// cannot suspend its caller indefinitely.
+    public static let defaultStartTimeout: Duration = .seconds(5)
```
dcalhoun, line 132: Added in https://github.com/wordpress-mobile/GutenbergKit/pull/357/changes/7da893b350f8bf37beb3717a3365a427273e4d85 to avoid media upload server start delays from indefinitely postponing editor launch. I welcome feedback on the implementation.

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -243,8 +243,16 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
...
+            // FAST PATH: Dependencies were provided at init() - load immediately.
+            //
+            // Deliberately NOT tracked in `dependencyTaskHandle`: `viewDidDisappear`
+            // cancels that handle to abort the async dependency *fetch*, but the
+            // fast path is cheap local work that must run to completion — a
+            // transient disappearance (e.g. a modal presented over the editor)
+            // cancelling it mid `startUploadServer()` silently disabled native
+            // uploads for the session. `[weak self]` still makes it a no-op once
+            // the controller is torn down.
+            Task(priority: .userInitiated) { [weak self] in
```
dcalhoun, line 262: Added in https://github.com/wordpress-mobile/GutenbergKit/pull/357/changes/f1ce3cccfe76273082a5aff57399817abce8d499. Based on my understanding, this seems correct. I welcome help confirming that.

ios/Sources/GutenbergKit/Sources/Media/MediaUploadServer.swift
```diff
@@ -34,8 +38,12 @@ final class MediaUploadServer: Sendable {
...
         maxRequestBodySize: Int64 = HTTPRequestParser.defaultMaxBodySize
     ) async throws -> MediaUploadServer {
-        // Sweep temp files orphaned by a prior crash before starting.
-        cleanOrphanedUploads()
+        // Sweep temp files orphaned by a prior crash, off the editor-startup
+        // path — the sweep only deletes stale files (>1 hour old), so it cannot
+        // race this server's own in-flight uploads and nothing below depends on it.
+        let cleanupTask = Task.detached(priority: .utility) {
+            cleanOrphanedUploads()
+        }
```
dcalhoun, line 46: Added in https://github.com/wordpress-mobile/GutenbergKit/pull/357/changes/45219181e3aa60b5ce3119367758d90e6b6beb0b. I believe moving this to a `Task` is sound for improving editor startup time, but I welcome a second glance to ensure this doesn't cause issue.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/HttpServer.kt
```diff
@@ -328,6 +385,38 @@ class HttpServer(
                 }
             }
 
+            // Drain the oversized body before responding so the (authenticated)
+            // client receives the 413 instead of a connection reset
+            // (RFC 9110 §15.5.14).
+            if (parser.state == HTTPRequestParser.State.DRAINING) {
+                readUntil(parser, input, buffer, deadlineNanos) { it.isComplete }
+            }
```
dcalhoun, line 393: Relocated in https://github.com/wordpress-mobile/GutenbergKit/pull/357/changes/acea3f54dfdff0bf4fee4e76d5cca5b469bf904f to avoid unauthorized requests causing body drains. Similar iOS changes added around `ios/Sources/GutenbergKitHTTP/HTTPServer.swift:285`.

dcalhoun:
**Note to self — coordinate with #573 before merging.**

#573 (plain-permalink support) fixes native REST URL building for sites whose only root is `https://site/?rest_route=/`. It merges with this branch without textual conflict (only shared file: `RESTAPIRepository.kt`), but two things need handling — whichever lands second rebases.

**1. Android: a clean merge silently reverts #573.** This PR rewrites `buildNamespacedUrl` to delegate to `RestUrlBuilder.namespaced` and drops the `apiRoot`/`namespace` fields — exactly where #573's fix lives. Git takes this PR's version cleanly and Android goes back to broken, with no conflict to catch it.

Fix is small — have `namespaced` delegate its join instead of concatenating:

```kotlin
fun namespaced(siteApiRoot: String, siteApiNamespace: String?, path: String): String =
    siteApiRoot.appendingRestPath(namespacedPath(siteApiNamespace, path))
```

`appendingRestPath` is the shared join from #573. It has to stay in the join, not move inside `namespaced` — both asset libraries build URLs without namespacing and need the same handling.

iOS needs nothing: `WordPressRESTURL.namespaced` already calls `apiRoot.appending(rawPath:)` on every path, which is what #573 fixes.

**2. `mediaEndpointURL` is broken on a query root, both platforms.** This is the divergence #563 flagged and assumed was moot until the base was fixed — it isn't. With a corrected base, tested live with `?_embed=wp:featuredmedia`:

| | Result | |
|---|---|---|
| Android — `namespaced(...) + query` | `?rest_route=/wp/v2/media?_embed=…` | **404** (double `?`) |
| iOS — `percentEncodedQuery = …` | `https://site/?_embed=…` | **`rest_route` destroyed** |
| correct | `?rest_route=/wp/v2/media&_embed=…` | 200 |

iOS is worse — `percentEncodedQuery` overwrites the whole query, so the endpoint vanishes and the request hits the site root.

Both need the rule the web layer already uses: when the base carries a query, merge with `&` rather than appending or overwriting. #573 mirrors `createRootURLMiddleware` natively, so this can reuse it rather than reimplement.

jkmassel (approved):

PR #363, opened 2026-03-10
fix: include block library editor styles for correct placeholder UI

dcalhoun:
## What?

Include `@wordpress/block-library` editor styles (`editor.css`) in the visual editor's style bundle.

## Why?

Fix CMM-1939.

The editor-specific styles from `@wordpress/block-library` were not being loaded after being erroneously removed in #332. This caused styling issues in blocks that rely on editor-only CSS — notably the Featured Image block, whose placeholder button rendered at 100% width instead of the intended max 48px.

## How?

Import `editor.css` from `@wordpress/block-library/build-style/` and pass it to `useEditorStyles` alongside the existing block library styles.

Note: This means the stylesheet is loaded into the DOM twice, but this matches the web editor's current implementation, albeit odd.

## Testing Instructions

1. Open the demo editor.
2. Add a Featured Image block.
3. Verify the placeholder buttons (e.g., "Upload") are compact (~48px) rather than stretching to full width.

No accessibility changes — this is a CSS-only fix restoring intended button dimensions.

## Screenshots or screencast 

Before | After
-- | --
before | after

### DIFF
--- a/src/components/visual-editor/index.jsx
+++ b/src/components/visual-editor/index.jsx
@@ -20,6 +20,7 @@ import '@wordpress/format-library';
 import componentStyles from '@wordpress/components/build-style/style.css?inline';
 import blockEditorContentStyles from '@wordpress/block-editor/build-style/content.css?inline';
 import blocksStyles from '@wordpress/block-library/build-style/style.css?inline';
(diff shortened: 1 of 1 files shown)


nbradbury (approved):
Fix is good! :shipit:

PR #366, opened 2026-03-10
feat: improve iOS offline editor experience

dcalhoun:
## What?

Improve the offline editor experience by adding network fallback mode support, an offline indicator, and graceful degradation when the device is offline.

## Why?

Fix CMM-1206.

When opening the iOS GutenbergKit editor while offline without a cached bundle, the editor fails to load entirely. The expected behavior is to fall back to the bundled editor and display a notice indicating that the user is offline.

## How?

- **Network fallback mode**: Add a `NetworkFallbackMode` enum and configuration property that enables automatic fallback to bundled resources when network requests fail (`EditorService`, `EditorConfiguration`).
- **Offline indicator**: Create an `OfflineIndicator` component that displays a "Working Offline" status bar with an offline icon when the device loses connectivity, using browser online/offline events. Uses `speak()` from `@wordpress/a11y` to announce "Network connection lost, working offline" for screen readers.
- **iOS demo app improvements**: Add a Network Fallback picker to the site configuration view. When the device is offline, fall back to offline defaults instead of showing an error, allowing the editor to still be opened.
- **Documentation**: Add a Network Fallback Mode section to the preloading docs.

## Testing Instructions

1. Open the iOS demo app.
2. Navigate to a configured site's Editor Configuration.
3. Enable "Automatic" for the Network Fallback picker.
4. Enable airplane mode.
5. Tap "Start" to open the editor.
6. Verify the editor loads with the bundled editor and displays a "Working Offline" indicator at the top with a wifi-off icon.
7. Disable airplane mode.
8. Verify the "Working Offline" indicator disappears.

1. Enable VoiceOver on the device.
2. With the editor open, enable airplane mode.
3. Verify VoiceOver announces "Network connection lost, working offline."
4. Navigate to the offline indicator and verify it reads "Working Offline."

## Screenshots or screencast

https://github.com/user-attachments/assets/935de5e6-1b38-434f-8b29-a46d7a053895

### DIFF
--- a/ios/Demo-iOS/Sources/Views/SitePreparationView.swift
+++ b/ios/Demo-iOS/Sources/Views/SitePreparationView.swift
@@ -292,8 +292,15 @@ class SitePreparationViewModel {
             siteURL: URL(string: config.siteUrl)!,
             siteApiRoot: URL(string: config.siteApiRoot)!
         )
-        .setShouldUseThemeStyles(false)
-        .setShouldUsePlugins(false)
+        // Optimistically enable theme styles and plugins so that
+        // previously-cached assets from an earlier online session can still be
+        // used. For sites that don't support these features, this is safe
+        // because EditorService won't be able to fetch the remote manifests
+        // while offline and the automatic network fallback will gracefully
+        // degrade to an empty asset bundle.
+        .setShouldUseThemeStyles(true)
+        .setShouldUsePlugins(true)
+        .setNetworkFallbackMode(.automatic)
         .setAuthHeader(config.authHeader)
         .setLogLevel(.debug)
         .build()
--- a/ios/Demo-iOS/Sources/Views/SitePreparationView.swift
+++ b/ios/Demo-iOS/Sources/Views/SitePreparationView.swift
@@ -57,6 +57,11 @@ struct SitePreparationView: View {
                 Toggle("Enable Native Inserter", isOn: $viewModel.enableNativeInserter)
                 Toggle("Enable Network Logging", isOn: $viewModel.enableNetworkLogging)
 
+                Picker("Network Fallback", selection: $viewModel.networkFallbackMode) {
+                    Text("Disabled").tag(NetworkFallbackMode.disabled)
+                    Text("Automatic").tag(NetworkFallbackMode.automatic)
+                }
+
                 if viewModel.postTypes.isEmpty {
                     HStack {
                         Text("Post Type")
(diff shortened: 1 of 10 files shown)


src/components/offline-indicator/index.jsx
```diff
@@ -0,0 +1,72 @@
...
+		window.addEventListener( 'offline', handleOffline );
+
+		return () => {
+			window.removeEventListener( 'online', handleOnline );
+			window.removeEventListener( 'offline', handleOffline );
+		};
+	}, [] );
+
+	return { isConnected };
+}
```
dcalhoun, line 72: Using `navigator.onLine` and the `online`/`offline` events is proving highly unreliable. Namely, they don't ever report an "offline" state in Android Chrome when airplane mode is activated (and wi-fi disabled). So, this needs to be replaced. 

That said, the primary target of this work is ensuring the iOS editor _opens_ when offline. Currently, it displays a full-screen error message. 

I consider this PR a quick iteration to improve the UX. I plan to follow up later, refactoring this "Working Offline" indicator to rely upon native OS capabilities that are far more capable of detecting internet—either within the GBK library or from bridge communication with the host app. 

In the meantime, I suggest we merge this given it unblocks the editor on iOS while offline.

ios/Demo-iOS/Sources/Views/SitePreparationView.swift
```diff
@@ -257,6 +278,27 @@ class SitePreparationViewModel {
...
+    }
+
+    private static func buildOfflineConfiguration(for config: ConfiguredEditor) -> EditorConfiguration {
+        EditorConfigurationBuilder(
+            postType: .post,
+            siteURL: URL(string: config.siteUrl)!,
+            siteApiRoot: URL(string: config.siteApiRoot)!
+        )
+        .setShouldUseThemeStyles(false)
+        .setShouldUsePlugins(false)
```
kean: Shouldn't you still be able to use plugins offline assuming all required assets are available locally? What happens if you get reconnected? Do you need to close and re-open the editor to get the "full" version?

kean (commented):
I smoke-tested the changes. The editor did load and indicator did show up.
Left one comment for clarification.

ios/Demo-iOS/Sources/Views/SitePreparationView.swift
dcalhoun: > Shouldn't you still be able to use plugins offline assuming all required assets are available locally? What happens if you get reconnected?

Yes, that makes sense. I addressed this in 396185237f9cc1e3e957ee48c1c4a900d314dcbb. 

It's OK to optimistically set these values to `true` in this specific scenario because a network request will never be made to the site. So, it has no negative impact if the site doesn't support these features. 

You can now use themes/plugins while offline when...

1. Open the site configuration view. 
2. Tap Preload Editor. 
3. Enable airplane mode. 
4. Start the editor. 

> What happens if you get reconnected? Do you need to close and re-open the editor to get the "full" version?

I did not handle this scenario specifically. Currently you would need to close and reopen the editor to fetch missing themes, plugins, etc.

kean (approved):

PR #370, opened 2026-03-16
Fix flaky "clear removes all entries" test

mokagio:
## What?

Run into flaky tests working on #318. Got an agent to fix it. Took a few iterations, but got there in the end. This fixes flaky `EditorURLCacheTests.clearRemovesAll` test.

## Why?

`URLCache.removeAllCachedResponses()` is asynchronous — the in-memory cache may still serve stale entries after the call returns.
The existing 0.2s sleep is not enough under CI load, causing intermittent failures.

## How?

The two `clear()`-related tests now verify through a fresh `EditorURLCache` instance pointing to the same `cacheRoot`.
A new instance bypasses the in-memory layer and reads only from disk, making the assertion deterministic.

The pipeline temporarily runs Swift tests 50x in parallel (`parallelism: 50`) to validate the fix.
Remove before merging.

## Testing Instructions

1. Check that all 50 parallel Swift test jobs pass in CI.
2. The `clear removes all entries` and `store succeeds after clear` tests should no longer flake.

---

*Posted by Claude Code (Opus 4.6) on behalf of @mokagio with approval.*

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -55,6 +55,7 @@ steps:
 
     - label: ':swift: Test Swift Package'
       command: swift test
+      parallelism: 50
       plugins: *plugins
 
     - label: ':ios: Test iOS E2E'
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -55,7 +55,6 @@ steps:
 
     - label: ':swift: Test Swift Package'
       command: swift test
-      parallelism: 50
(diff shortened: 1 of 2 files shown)


jkmassel (approved):
> waitForClearToTakeEffect

Welp I surely hate this. 

But it works, so let's ship it :) 

I think this is a pretty strong sign that we shouldn't use `URLCache` for anything though...

Thanks for taking care of this Gio!! 🙇

PR #382, opened 2026-03-17
build(deps-dev): upgrade to Vite 8 and @vitejs/plugin-react 6

dcalhoun:
## What?

Upgrade Vite from v7 to v8 and `@vitejs/plugin-react` from v5 to v6.

Supersedes https://github.com/wordpress-mobile/GutenbergKit/pull/373 and #380.

## Why?

Vite 8 replaces esbuild and Rollup with [Rolldown and Oxc](https://vite.dev/blog/announcing-vite8), the new Rust-based build toolchain. This keeps us on the latest major version and unblocks upgrading `@vitejs/plugin-react` to v6, which was [required by the Dependabot PR](https://github.com/wordpress-mobile/GutenbergKit/pull/373#issuecomment-4076188720).

### Dependency changes (`package.json`)

- `vite`: `^7.3.1` → `^8.0.0`
- `@vitejs/plugin-react`: `^5.1.4` → `^6.0.1`
- Removed `vite-plugin-node-polyfills` (does not support Vite 8, and was replaced by a lighter approach)

### Replace `vite-plugin-node-polyfills` with resolve stubs (`vite.config.js`, `src/stubs/node-modules.js`)

PostCSS imports Node.js modules (`path`, `fs`, `url`, `source-map-js`) that don't exist in browsers. PostCSS marks these as `false` in its `browser` field, but `@wordpress/block-editor` imports PostCSS via direct file paths (e.g., `postcss/lib/processor`), bypassing that field and causing Vite to emit "Module has been externalized for browser compatibility" warnings.

The `vite-plugin-node-polyfills` plugin suppressed these by providing full browser polyfills (~600 KB in dependencies), but PostCSS never actually calls these modules at runtime—it guards usage behind `Boolean()` checks. A single empty stub module aliased via `resolve.alias` achieves the same result with zero bundle impact and no external dependencies.

### Patch mousetrap global-bind plugin (`patches/mousetrap+1.6.5.patch`)

Vite 8's Rolldown bundler may reorder side-effect imports relative to value imports ([rolldown/rolldown#6436](https://github.com/rolldown/rolldown/issues/6436), [vitejs/vite#5142](https://github.com/vitejs/vite/issues/5142)). This caused mousetrap's `global-bind` plugin to execute before the main mousetrap module, so `window.Mousetrap` was not yet set and `bindGlobal` was never added to the prototype—breaking `@wordpress/keyboard-shortcuts`.

The patch adds a `require("mousetrap")` fallback so the plugin resolves Mousetrap via the module system instead of relying on the global. This is preferred over Rolldown's [`output.strictExecutionOrder`](https://rolldown.rs/reference/OutputOptions.strictExecutionOrder) option, which would increase bundle size by injecting runtime helpers across the entire bundle.

## Testing Instructions

1. Run the dev server (`make dev-server`) and verify the editor loads without errors in the browser console.
2. Run `make build` and verify it completes successfully.
3. Run `make test-js` and verify all tests pass.
4. Test with plugins that register keyboard shortcuts (e.g., Jetpack AI Assistant[^1]) and verify no `mousetrap[bindGlobal]` errors appear.

[^1]: Requires adding a site to the demo app that has an active Jetpack connection—e.g., WoW/Atomic. 

N/A, no user-facing changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -75,7 +75,7 @@
 				"@testing-library/user-event": "^14.6.1",
 				"@types/react": "^19.2.7",
 				"@types/react-dom": "^19.2.3",
-				"@vitejs/plugin-react": "^5.1.4",
+				"@vitejs/plugin-react": "^6.0.1",
 				"@wordpress/dependency-extraction-webpack-plugin": "^6.41.0",
 				"@wordpress/env": "^11.2.0",
 				"@wordpress/eslint-plugin": "^24.3.0",
@@ -91,8 +91,7 @@
 				"react-devtools": "^7.0.1",
 				"react-devtools-core": "^7.0.1",
 				"sass-embedded": "^1.98.0",
-				"vite": "^7.3.1",
-				"vite-plugin-node-polyfills": "^0.25.0",
+				"vite": "^8.0.0",
 				"vitest": "^4.1.0"
 			}
 		},
@@ -259,6 +258,7 @@
 			"integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
 			"dev": true,
 			"license": "MIT",
+			"peer": true,
 			"dependencies": {
 				"@babel/code-frame": "^7.29.0",
 				"@babel/generator": "^7.29.0",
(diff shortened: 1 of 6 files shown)


dcalhoun:
@kean friendly reminder to review this. No worries if you simply need more time due to other priorities. Let me know if you have any questions.

kean (approved):
LGTM!

PR #387, opened 2026-03-18
fix: improve offline indicator accuracy with HTTP probe

dcalhoun:
## What?
Fixes the offline indicator not appearing in Android WebView, and improves its accuracy across all platforms.

## Why?
Two issues:

1. **Android WebView never fires `online`/`offline` events** — a known long-standing WebView limitation. Unlike Chrome (a full browser with its own network stack), Android WebView does not synthesize these events when OS network state changes, so the offline banner never appeared.
2. **`navigator.onLine` and browser `online`/`offline` events are unreliable** on other platforms — they only confirm a network interface is active, not that the internet is reachable. Chrome is also known to not fire the `offline` event in all network loss scenarios.

Fix CMM-1945. Follow up on https://github.com/wordpress-mobile/GutenbergKit/pull/366#discussion_r2915245274.

### Android native polyfill
Adds a `ConnectivityManager.NetworkCallback` inside `GutenbergView` that monitors OS network state and synthesizes `window.dispatchEvent(new Event('online'|'offline'))` into the WebView — a transparent polyfill requiring no host app changes. The callback is registered in `onAttachedToWindow` and unregistered in `onDetachedFromWindow`. Also adds the required `ACCESS_NETWORK_STATE` permission to the library manifest (a normal permission; no runtime grant needed).

### JS probe-based connectivity strategy
Replaces the simple event listener approach with an asymmetric strategy:

- **Initial state** reads `navigator.onLine`. A `false` value is trusted immediately (definitive signal the device has no interface). A `true` value triggers a mount-time probe to catch misreported online status before any event fires.
- **`offline` event**: trusted immediately — this signal is reliable.
- **`online` event**: verified by a lightweight `HEAD` probe before clearing the indicator, guarding against false positives.

The probe targets `siteApiRoot` from the GBKit config, testing actual WordPress API reachability. Falls back to `/favicon.ico` in local dev environments.

## Testing Instructions
1. If set, remove `GUTENBERG_EDITOR_URL` from `local.properties`. 
2. Run `make build` to build the latest JavaScript. 
1. Open the Android demo app editor on a device or emulator.
4. Toggle airplane mode on — verify the offline banner appears.
5. Toggle airplane mode off — verify the banner disappears.

The offline indicator uses `wp.a11y.speak()` to announce connectivity loss to screen readers. Verify the announcement is made when going offline.

## Screenshots or screencast 

<details><summary>Screen recording</summary>
<p>

https://github.com/user-attachments/assets/06076541-e06f-4435-9e20-7d18fa77a54a

</p>
</details> 

<details><summary>Screenshot</summary>
<p>

android-working-offline

</p>
</details>

### DIFF
--- a/android/Gutenberg/src/main/AndroidManifest.xml
+++ b/android/Gutenberg/src/main/AndroidManifest.xml
@@ -1,4 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+
 </manifest>
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
@@ -4,6 +4,10 @@ import android.annotation.SuppressLint
 import android.content.Context
 import android.content.Intent
 import android.graphics.Bitmap
+import android.net.ConnectivityManager
+import android.net.Network
+import android.net.NetworkCapabilities
+import android.net.NetworkRequest
 import android.net.Uri
 import android.os.Bundle
 import android.os.Handler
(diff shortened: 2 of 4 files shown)


adalpari (approved):
Works great!! 🚢 it!

PR #401, opened 2026-03-25
Add XCFramework build infrastructure and CI

mokagio:
## Summary

Adds the automation for building, signing, and uploading the resources XCFramework build. This will allow having a smaller follow up PR focused only on updating the release automation to handle version bumping, `Package.swift` update, and GitHub release.

This lands the build and upload pipeline without wiring into `Package.swift` — consumers still use the local source target.

I pushed a one-off `Package.swift` update branch to use the latest artifact built here and [tested it in WordPress iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/25452) via CI-made Prototype Build.

- Add `build_xcframework.sh` script that archives `GutenbergKitResources` for device + simulator, links `.o` into a dylib, assembles an XCFramework with resource bundles and dSYMs, and outputs a zip with SPM checksum
- Add `make build-resources-xcframework` target
- Add Fastlane setup scoped to `publish_to_s3` and `xcframework_sign` lanes (release orchestration lanes deferred)
- Add Buildkite steps: XCFramework build on every push, S3 publish on every build under a tag-aware version (`NEW_VERSION` override → `BUILDKITE_TAG` → `BUILDKITE_COMMIT`), with `if_exists: :replace` so retries are idempotent

## Test plan

- [ ] Verify `make build-resources-xcframework` produces a valid zip + checksum locally
- [ ] Verify Buildkite build step succeeds and uploads artifacts
- [ ] Verify S3 publish writes under the commit SHA on a push build, and overwrites cleanly on a same-commit retry

🤖 Generated with [Claude Code](https://claude.ai/code)

---

_Posted by Claude Code (Opus 4.6) on behalf of @mokagio with approval._

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -67,6 +67,24 @@ steps:
           make test-swift-package
       plugins: *plugins
 
+    - label: ':xcode: Build XCFramework'
+      key: build-xcframework
+      depends_on: build-react
+      command: |
+          buildkite-agent artifact download dist.tar.gz .
+          tar -xzf dist.tar.gz
+          make build-resources-xcframework
+      artifact_paths:
+          - '*.xcframework.zip'
+          - '*.xcframework.zip.checksum.txt'
+      plugins: *plugins
+
+    - label: ':s3: Publish XCFramework to S3'
+      if: build.env("NEW_VERSION") != null
+      depends_on: build-xcframework
+      command: bundle exec fastlane publish_to_s3 version:$NEW_VERSION
+      plugins: *plugins
+
     - label: ':ios: Test iOS E2E'
       depends_on: build-react
       command: |
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -96,16 +96,15 @@ steps:
       plugins: *plugins
 
     - label: ':s3: Publish XCFramework to S3'
-      # Eventually, we might want to gate this based on an env var
-      # to be set by the process that starts the pipeline
-      #
-      # if: build.env("NEW_VERSION") != null
       depends_on: build-xcframework
       command: |
           buildkite-agent artifact download '*.xcframework.zip' .
           buildkite-agent artifact download '*.xcframework.zip.checksum.txt' .
           install_gems
-          bundle exec fastlane publish_to_s3 version:${NEW_VERSION:-$BUILDKITE_COMMIT}
+          # Version precedence: explicit `NEW_VERSION` override wins, then a
+          # tag build publishes under the tag, otherwise fall back to the
+          # commit SHA so every push gets a stable artifact URL.
+          bundle exec fastlane publish_to_s3 version:${NEW_VERSION:-${BUILDKITE_TAG:-$BUILDKITE_COMMIT}}
       plugins: *plugins
 
     - label: ':ios: Test iOS E2E'
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -73,7 +73,11 @@ steps:
       command: |
           buildkite-agent artifact download dist.tar.gz .
           tar -xzf dist.tar.gz
+          install_gems
+          bundle exec fastlane set_up_signing_release
           make build-resources-xcframework
+          bundle exec fastlane xcframework_sign
+          ./package_xcframework.sh
       artifact_paths:
           - '*.xcframework.zip'
           - '*.xcframework.zip.checksum.txt'
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -85,7 +85,11 @@ steps:
       command: |
           buildkite-agent artifact download dist.tar.gz .
           tar -xzf dist.tar.gz
+          install_gems
+          bundle exec fastlane set_up_signing_release
           make build-resources-xcframework
+          bundle exec fastlane xcframework_sign
+          ./package_xcframework.sh
       artifact_paths:
           - '*.xcframework.zip'
           - '*.xcframework.zip.checksum.txt'
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -96,9 +96,16 @@ steps:
       plugins: *plugins
 
     - label: ':s3: Publish XCFramework to S3'
-      if: build.env("NEW_VERSION") != null
+      # Eventually, we might want to gate this based on an env var
+      # to be set by the process that starts the pipeline
+      #
+      # if: build.env("NEW_VERSION") != null
       depends_on: build-xcframework
-      command: bundle exec fastlane publish_to_s3 version:$NEW_VERSION
+      command: |
+          buildkite-agent artifact download '*.xcframework.zip' .
+          buildkite-agent artifact download '*.xcframework.zip.checksum.txt' .
+          install_gems
+          bundle exec fastlane publish_to_s3 version:${NEW_VERSION:-$BUILDKITE_COMMIT}
       plugins: *plugins
 
     - label: ':ios: Test iOS E2E'
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -84,9 +84,16 @@ steps:
       plugins: *plugins
 
     - label: ':s3: Publish XCFramework to S3'
-      if: build.env("NEW_VERSION") != null
+      # Eventually, we might want to gate this based on an env var
+      # to be set by the process that starts the pipeline
+      #
+      # if: build.env("NEW_VERSION") != null
       depends_on: build-xcframework
-      command: bundle exec fastlane publish_to_s3 version:$NEW_VERSION
+      command: |
+          buildkite-agent artifact download '*.xcframework.zip' .
+          buildkite-agent artifact download '*.xcframework.zip.checksum.txt' .
+          install_gems
+          bundle exec fastlane publish_to_s3 version:${NEW_VERSION:-$BUILDKITE_COMMIT}
       plugins: *plugins
 
     - label: ':ios: Test iOS E2E'
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -77,6 +77,26 @@ steps:
           make test-swift-library
       plugins: *plugins
 
+    - label: ':xcode: Build XCFramework'
+      key: build-xcframework
+      depends_on:
+        - build-react
+        - swift-test-library
+      command: |
+          buildkite-agent artifact download dist.tar.gz .
+          tar -xzf dist.tar.gz
+          make build-resources-xcframework
+      artifact_paths:
+          - '*.xcframework.zip'
+          - '*.xcframework.zip.checksum.txt'
+      plugins: *plugins
+
+    - label: ':s3: Publish XCFramework to S3'
+      if: build.env("NEW_VERSION") != null
+      depends_on: build-xcframework
+      command: bundle exec fastlane publish_to_s3 version:$NEW_VERSION
+      plugins: *plugins
+
     - label: ':ios: Test iOS E2E'
       depends_on: build-react
       command: |
--- a/build_xcframework.sh
+++ b/build_xcframework.sh
@@ -156,16 +156,17 @@ copy_resource_bundles() {
 
     echo "--- Copying resource bundles for ${sdk}"
 
-    if [ -d "${bundle_path}" ]; then
-        find "${bundle_path}" -name "*.bundle" -maxdepth 1 -type d -print0 | while IFS= read -r -d '' bundle; do
-            bundle_name=$(basename "${bundle}")
-            echo "  ${bundle_name} -> ${framework_path}/"
-            rm -rf "${framework_path:?}/${bundle_name}"
-            cp -R "${bundle}" "${framework_path}/"
-        done
-    else
-        echo "  Warning: bundle path not found: ${bundle_path}"
+    if [ ! -d "${bundle_path}" ]; then
+        echo "Error: bundle path not found: ${bundle_path}" >&2
+        exit 1
     fi
+
+    find "${bundle_path}" -name "*.bundle" -maxdepth 1 -type d -print0 | while IFS= read -r -d '' bundle; do
+        bundle_name=$(basename "${bundle}")
+        echo "  ${bundle_name} -> ${framework_path}/"
+        rm -rf "${framework_path:?}/${bundle_name}"
+        cp -R "${bundle}" "${framework_path}/"
+    done
 }
 
 # Slice directory names depend on the architectures actually built — e.g.
--- a/build_xcframework.sh
+++ b/build_xcframework.sh
@@ -1,0 +1,207 @@
+#!/usr/bin/env bash
+
+# Builds a GutenbergKitResources XCFramework from the local source target.
+#
+# Prerequisites:
+#   - Built web assets in ios/Sources/GutenbergKitResources/Gutenberg/
+#   - Package.swift must have `resourcesMode` set to `.local`
+#
+# Output:
+#   - GutenbergKitResources-<git-sha>.xcframework.zip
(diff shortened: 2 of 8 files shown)


.buildkite/pipeline.yml
```diff
@@ -67,6 +67,35 @@ steps:
...
+      key: build-xcframework
+      depends_on: build-react
+      command: |
+          buildkite-agent artifact download dist.tar.gz .
+          tar -xzf dist.tar.gz
+          install_gems
+          bundle exec fastlane set_up_signing_release
+          make build-resources-xcframework
+          bundle exec fastlane xcframework_sign
+          ./package_xcframework.sh
```
mokagio, line 80: I admit this is quite confusing with a mash of Fastlane, make, and scripts. I'll tidy up in a follow up, my preference being to use make as the single entry point.

.buildkite/pipeline.yml
```diff
@@ -67,6 +67,35 @@ steps:
...
+      # Eventually, we might want to gate this based on an env var
+      # to be set by the process that starts the pipeline
+      #
+      # if: build.env("NEW_VERSION") != null
+      depends_on: build-xcframework
+      command: |
+          buildkite-agent artifact download '*.xcframework.zip' .
+          buildkite-agent artifact download '*.xcframework.zip.checksum.txt' .
+          install_gems
+          bundle exec fastlane publish_to_s3 version:${NEW_VERSION:-$BUILDKITE_COMMIT}
```
mokagio, line 96: Android publishes every version but does not have the issue of having to update `Package.swift` to match the checksum generated by the build.

Given storage is cheap, it might be okay for us to publish the XCFramework on every build, too, just to continue exercising the infra and avoid surprises at release time.

Or, we could be frugal and only publish when we need. I'm leaning toward publishing one every build, though. Faster feedback. And we could add additional automation to easily push a commit with updated `Package.swift` so that any XCFramework in S3 can be tested in WordPress iOS in a few clicks.

.buildkite/pipeline.yml
mokagio: https://linear.app/a8c/issue/AINFRA-1983/followup-gutenbergkitresources-setup-move-script-away-from-root

build_xcframework.sh
```diff
@@ -0,0 +1,193 @@
...
+build_framework() {
+    local sdk="$1"
+    local destination="$2"
+    local archive_path="${BUILD_DIR}/${SCHEME}-${sdk}.xcarchive"
+
+    echo "--- Building ${SCHEME} for ${sdk}"
+
+    rm -rf "${archive_path}"
+
+    xcodebuild archive \
```
jkmassel, line 84: Should we add `ARCHS="arm64 x86_64"` to this instead of relying on the default? 

IMHO we should probably ship `x86_64` for as long as the minimum deployment target overlaps with it, which might be longer than it remains the Xcode default?

build_xcframework.sh
```diff
@@ -0,0 +1,193 @@
...
+
+SCHEME="GutenbergKitResources"
+MINIMUM_IOS_VERSION="17.0"
+BUILD_DIR="$(pwd)/build"
+DERIVED_DATA_PATH="${BUILD_DIR}/DerivedData"
+
+GIT_SHA="$(git rev-parse HEAD)"
+XCFRAMEWORK_NAME="${SCHEME}-${GIT_SHA}.xcframework"
+
+link_dylib() {
```
jkmassel, line 26: I was surprised to see the use of dynamic linking instead of static – I can see some benefits (namely avoiding multiple copies of the library if it's used in an extension) at the cost of a more complex build process, but I wasn't sure if this was done deliberately or just how it panned out during implementation?

fastlane/Fastfile
```diff
@@ -0,0 +1,101 @@
...
+
+  sync_code_signing(
+    team_id: APPLE_TEAM_ID,
+    api_key: app_store_connect_api_key,
+    type: 'appstore',
+    storage_mode: 's3',
+    s3_region: 'us-east-2',
+    s3_bucket: 'a8c-fastlane-match',
+    readonly: readonly,
+    app_identifier: 'com.automattic.hostmgr',
```
jkmassel, line 63: I assume we'll update this at some point? 😁

build_xcframework.sh
```diff
@@ -0,0 +1,193 @@
...
+    local intermediates="${DERIVED_DATA_PATH}/Build/Intermediates.noindex/ArchiveIntermediates/${SCHEME}"
+    local build_products="${intermediates}/BuildProductsPath/Release-${sdk}"
+    local generated_maps="${intermediates}/IntermediateBuildFilesPath/GeneratedModuleMaps-${sdk}"
+
+    local framework_path="${BUILD_DIR}/frameworks/${sdk}/${SCHEME}.framework"
+    rm -rf "${framework_path}"
+    mkdir -p "${framework_path}/Modules"
+    mkdir -p "${framework_path}/Headers"
+
+    # Binary: SPM produces a .o object file — link it into a dylib so that
```
jkmassel, line 108: If we're using Xcode to produce this anyway (since XCFrameworks are an Xcode-ism), could we modify `Package.swift` to use:

```swift
.library(name: "GutenbergKitResources", type: .dynamic, targets: ["GutenbergKitResources"]),
```

the `.dynamic` should make it so we can delete `link_dylib` entirely and allow Xcode to handle `Info.plist` generation stuff too, I think?

fastlane/Fastfile
```diff
@@ -0,0 +1,101 @@
...
+
+lane :publish_to_s3 do |options|
+  version = required_version!(options)
+
+  upload_to_s3(
+    bucket: 'a8c-apps-public-artifacts',
+    key: File.join('gutenbergkit', version, 'GutenbergKitResources.xcframework.zip'),
+    file: xcframework_file_path,
+    auto_prefix: false,
+    if_exists: :fail
```
jkmassel, line 30: If the publishing step somehow fails, we'd need to cut a new release right? 

I don't think that's the end of the world, just double-checking that we're all ok with that :)

build_xcframework.sh
```diff
@@ -0,0 +1,193 @@
...
+
+    if [ -d "${bundle_path}" ]; then
+        find "${bundle_path}" -name "*.bundle" -maxdepth 1 -type d -print0 | while IFS= read -r -d '' bundle; do
+            bundle_name=$(basename "${bundle}")
+            echo "  ${bundle_name} -> ${framework_path}/"
+            rm -rf "${framework_path:?}/${bundle_name}"
+            cp -R "${bundle}" "${framework_path}/"
+        done
+    else
+        echo "  Warning: bundle path not found: ${bundle_path}"
```
jkmassel, line 157: Should the script fail instead of just emitting a warning here?

fastlane/Fastfile
mokagio: We can but don't need to. See what wordpress-rs does: https://github.com/Automattic/wordpress-rs/blob/1cb0f48eb08879bf8e0f03e2ac38885c2515c9b8/fastlane/Fastfile#L650-L675

What I should do, however, is document this.

build_xcframework.sh
mokagio: I remember going back and forth on what type of linking to use during the implementation, with a change here resulting in issues elsewhere. Definitely something to try, though.

build_xcframework.sh
mokagio: Good point!

fastlane/Fastfile
mokagio: > If the publishing step somehow fails, we'd need to cut a new release right?

The automation at this point doesn't concern itself with the release in the GitHub/npm/tag release meaning of the word. Maybe we can be strict at the beginning and build in some retry-ability for certain failures (e.g. build fine but can't hit S3 because it's down) in the future?

build_xcframework.sh
mokagio: To be honest, I don't remember exactly. It's likely a byproduct of various trial and errors. I'll try updating with static linking and see how we go...

mokagio: https://linear.app/a8c/issue/AINFRA-2256/verify-behavior-with-static-linking-for-gutenbergkitresources

build_xcframework.sh
mokagio: Will try as part of https://linear.app/a8c/issue/AINFRA-2256/verify-behavior-with-static-linking-for-gutenbergkitresources

.buildkite/pipeline.yml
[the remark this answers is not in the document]
mokagio: Addressed in df8e988c — fixed in the Makefile rather than the pipeline. `build-resources-xcframework` now calls `make copy-dist-ios` explicitly (mirroring what `test-ios-e2e` already does), so the recipe is correct regardless of whether the chained `make build` short-circuits.

*Posted by Claude Code (Opus 4.7) on behalf of @mokagio with approval.*

build_xcframework.sh
```diff
@@ -0,0 +1,193 @@
...
+# Copy dSYMs into xcframework slices
+DEVICE_DSYMS="${BUILD_DIR}/${SCHEME}-iphoneos.xcarchive/dSYMs"
+SIM_DSYMS="${BUILD_DIR}/${SCHEME}-iphonesimulator.xcarchive/dSYMs"
+
+if [ -d "${DEVICE_DSYMS}" ]; then
+    cp -r "${DEVICE_DSYMS}" "${XCFRAMEWORK_PATH}/ios-arm64/"
+fi
+if [ -d "${SIM_DSYMS}" ]; then
+    cp -r "${SIM_DSYMS}" "${XCFRAMEWORK_PATH}/ios-arm64_x86_64-simulator/"
+fi
```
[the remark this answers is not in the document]
mokagio: Addressed in b4de7fae. Replaced the hardcoded `ios-arm64/` and `ios-arm64_x86_64-simulator/` paths with a `resolve_slice_dir` helper that walks the XCFramework's actual `ios-*` directories and requires exactly one match per SDK — so it works whether the simulator slice is universal or arm64-only.

*Posted by Claude Code (Opus 4.7) on behalf of @mokagio with approval.*

package_xcframework.sh
```diff
@@ -0,0 +1,39 @@
...
+#   - GutenbergKitResources-<git-sha>.xcframework.zip
+#   - GutenbergKitResources-<git-sha>.xcframework.zip.checksum.txt
+
+set -euo pipefail
+
+BUILD_DIR="$(pwd)/build"
+OUTPUT_DIR="${1:-$(pwd)}"
+
+XCFRAMEWORK_PATH=$(find "${BUILD_DIR}" -maxdepth 1 -name "*.xcframework" -type d | head -1)
+
```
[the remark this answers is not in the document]
mokagio: Addressed in aa63175f and 634ef3cd. Two-part fix:

1. `build_xcframework.sh` now removes any prior `*.xcframework`, `*.xcframework.zip`, and `*.xcframework.zip.checksum.txt` from `build/` at the top of the recipe, so re-runs are hermetic.
2. `package_xcframework.sh` addresses the artifact by exact SHA-suffixed name (`build/${SCHEME}-${GIT_SHA}.xcframework`) instead of globbing — so the packager fails loudly if the expected build is missing, rather than silently picking whichever `.xcframework` happens to be sitting in `build/`.

*Posted by Claude Code (Opus 4.7) on behalf of @mokagio with approval.*

fastlane/Fastfile
```diff
@@ -0,0 +1,101 @@
...
+end
+
+def xcframework_dir_path
+  Dir.glob(File.join(PROJECT_ROOT, 'build', 'GutenbergKitResources-*.xcframework')).first \
+    || UI.user_error!('XCFramework not found. Run build_xcframework.sh first.')
+end
+
+def xcframework_file_path
+  Dir.glob(File.join(PROJECT_ROOT, 'GutenbergKitResources-*.xcframework.zip')).first \
+    || UI.user_error!('XCFramework zip not found. Run package_xcframework.sh first.')
```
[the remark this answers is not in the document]
mokagio: Partially addressed in aa63175f. `build_xcframework.sh` now removes prior `*.xcframework` / `*.xcframework.zip` / `*.xcframework.zip.checksum.txt` before each build, so by the time these Fastlane helpers run the globs always match exactly one file in practice.

The Ruby helpers themselves still use `Dir.glob(...).first` — so the explicit `require_single_glob_match!` pattern from your suggestion isn't implemented. Happy to add it as a defensive layer if you'd prefer the helpers be robust on their own rather than relying on the build script's cleanup.

*Posted by Claude Code (Opus 4.7) on behalf of @mokagio with approval.*

fastlane/Fastfile
```diff
@@ -0,0 +1,103 @@
...
+    team_id: APPLE_TEAM_ID,
+    api_key: app_store_connect_api_key,
+    type: 'appstore',
+    storage_mode: 's3',
+    s3_region: 'us-east-2',
+    s3_bucket: 'a8c-fastlane-match',
+    readonly: readonly,
+    # We don't need provisioning profile, only the certificate.
+    # An empty identifier list does the trick and avoids confusion.
+    app_identifier: [],
```
mokagio, line 65: @jkmassel I think this is a cleaner option that what I had before. What do you think?

It results in only the cert being fetched from S3 without provisioning profile confusion

Image

And of course code signing works as expected because only the cert is needed:

Image

fastlane/Fastfile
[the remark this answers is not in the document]
mokagio: Upgraded to fully addressed in 3e9a8806 — added a `require_single_glob_match!` helper and routed the three path helpers through it, so they now error explicitly on zero or multiple matches rather than relying on the build script's cleanup to guarantee a single match. Verified end-to-end via `bundle exec fastlane xcframework_sign`: single-match resolves to the expected path, multi-match aborts the lane with both paths listed.

*Posted by Claude Code (Opus 4.7) on behalf of @mokagio with approval.*

.buildkite/pipeline.yml
```diff
@@ -67,6 +67,35 @@ steps:
...
+      plugins: *plugins
```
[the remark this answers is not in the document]
mokagio: Addressed in 7a496695, though intentionally not via the gating you suggested. Decided to commit to per-build publishes for now and make the behavior consistent rather than partly-implemented:

- Removed the commented-out `if:` guard.
- Switched the version expression to `${NEW_VERSION:-${BUILDKITE_TAG:-$BUILDKITE_COMMIT}}` — explicit override wins, then a tag (so tagged builds publish under the tag), otherwise the commit SHA.
- Flipped both `upload_to_s3` calls to `if_exists: :replace` so same-commit retries no longer fail the lane.

Updated the PR description and test plan to match. Proper gating will land with the release-orchestration follow-up PR mentioned in the summary.

*Posted by Claude Code (Opus 4.7) on behalf of @mokagio with approval.*

build_xcframework.sh
mokagio: Addressed in b962cf8e57f29933a9b206316b1ebcbf67e77903, including switching the logic to fail early

build_xcframework.sh
mokagio: Sure can, but I'm not sure it's a clear win. As such, I implemented it in #482 for further discussion.

mokagio:
@jkmassel  I think I addressed or tracked all your and Copilot's suggestions. 

In its current state, this publishes the XCFramework on every build, like Android does, but doesn't wire it in `Package.swift` for consumption. That is something we shall address in the next PR, in a way that can ideally make the package version bump, tag, and `Package.swift` update happen in lockstep.

If something happens in the meantime that requires us to switch to the XCFramework consumption mode, it should be straightforward to push a commit here that changes `Package.swift` and have the consumers pin the library to it.

.buildkite/pipeline.yml
jkmassel: We package every build for https://github.com/automattic/wordpress-rs/ which is far larger without issue, so I think this is ok too :)

jkmassel (approved):
Looks like we're in good shape, thanks Gio!!

PR #413, opened 2026-04-02
ci: update CodeQL workflow actions and add Dependabot monitoring

jkmassel:
## Summary
- Bump `actions/checkout` from v4 to v5
- Bump `github/codeql-action` (init, autobuild, analyze) from v3 to v4
- Add `github-actions` ecosystem to Dependabot to automatically track future action version updates

Resolves GitHub Actions deprecation warnings for Node.js 20. Both v5 (checkout) and v4 (codeql-action) are drop-in replacements with no breaking changes for this workflow.

## Test plan
- [ ] Verify CodeQL workflow runs successfully on this PR
- [ ] Confirm no Node.js deprecation warnings in workflow logs

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -22,13 +22,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                include:
-                    - language: actions
-                      build-mode: none
-                    - language: java-kotlin
-                      build-mode: autobuild
-                    - language: javascript-typescript
-                      build-mode: none
+                language: [actions, javascript-typescript]
 
         steps:
             - name: Checkout repository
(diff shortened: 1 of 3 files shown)


dcalhoun (commented):
While I do not see any deprecation warnings in the CI logs, it appears the Swift CodeQL task continues to cancel due to a timeout. Here is Claude's analysis. WDYT is the most appropriate solution?

<details><summary>Claude analysis</summary>

## Root Cause

The `analyze-swift` job is hitting its **30-minute timeout** (`codeql.yml:67`). The `swift build --build-tests` step alone takes ~29.5 minutes, leaving no time for the subsequent "Build Demo app" step. The build got to 277/279 compilations before being killed.

Two factors contribute to the long build time:

1. **SPM dependency resolution is slow** — computing the version for `SVGView.git` alone took **531 seconds** (~9 minutes). This is a known issue with SPM when using `from:` ranges on repos with many tags — SPM checks every tag to find compatible versions.

2. **The build itself is large** — compiling all targets including tests (279 compilation units total) plus all dependencies (SwiftSoup, SVGView, GutenbergKitHTTP, etc.) takes ~20 minutes.

And the "Build Demo app" xcodebuild step would add even more time on top of that.

## Fix Options

**Option 1 (simplest): Increase the timeout**

Bump `timeout-minutes` from 30 to 60 on the `analyze-swift` job. This gives enough headroom for both build steps plus the CodeQL analysis step.

**Option 2 (also helps): Remove `--build-tests`**

Change `swift build --build-tests` to just `swift build`. CodeQL only needs the source code compiled for analysis — test code is typically not what you want to scan for security vulnerabilities. This would cut out ~50 compilation units and save several minutes.

**Option 3 (complementary): Remove the Demo app build**

The Demo app build adds more compilation time. If the Swift package build already covers the main library code, the Demo app step may be redundant for CodeQL's purposes.

</details>

.github/workflows/codeql.yml
```diff
@@ -22,46 +22,78 @@ jobs:
...
+              uses: github/codeql-action/init@v4
               with:
                   languages: swift
 
             - name: Build Swift package
-              run: swift build --target GutenbergKit --target GutenbergKitHTTP
+              run: swift build
+
+            - name: Build Demo app
+              run: xcodebuild build -project ios/Demo-iOS/Gutenberg.xcodeproj -scheme Gutenberg -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO
```
iangmaia, line 94: 💭 Just wondering, would it make sense to check in the Demo app `Package.resolved`, now that this job builds the Xcode project? That would make dependency resolution for this step reproducible.

PR #415, opened 2026-04-02
fix(android): use ACTION_OPEN_DOCUMENT to avoid ERR_UPLOAD_FILE_CHANGED

dcalhoun:
## What?

Replace `ACTION_GET_CONTENT` with `ACTION_OPEN_DOCUMENT` for the WebView file chooser and remove the `FileCache` workaround that was copying cloud-backed files to local cache.

## Why?

On Android 13+, the system Photo Picker intercepts `ACTION_GET_CONTENT` for image/video MIME types, returning proxy `content://media/picker/...` URIs. These proxy URIs report inconsistent `last_modified` metadata between queries, which triggers Chromium's `ERR_UPLOAD_FILE_CHANGED` validation and fails the upload. The existing `FileCache` workaround ([#224](https://github.com/wordpress-mobile/GutenbergKit/pull/224)) mitigated this by copying files to local cache before handing them to the WebView, but at the cost of doubling storage usage per upload. With the addition of #357, that memory pressure will continue to increase, as a local cache is used there as well. 

Additional details from the #224 workaround description:

> Due to a [Chrome bug](https://issues.chromium.org/issues/40123366), uploading media often failed if the file was provided by a cloud-based content provider (e.g., the Google Drive app). The error is almost always thrown, but generally only causes video (or larger) file uploads to fail. 
> 
> This issue is also [noted in the `tus-js-client` project](https://github.com/tus/tus-js-client/issues/255#issuecomment-828386273), which is used by VideoPress. The `ERR_UPLOAD_FILE_CHANGE` error occurs for most all media uploads (images, videos, etc), but seemingly mostly causes VideoPress uploads to fail, likely due to the `tus` usage. 
> 
> The `ERR_UPLOAD_FILE_CHANGE` error is also _not_ thrown when using GutenbergKit/VideoPress from the Chrome browser directly. It only occurs in the Android WebView. 

## How?

`ACTION_OPEN_DOCUMENT` routes directly to DocumentsUI (the system file browser), which returns stable `content://com.android.providers.media.documents/...` URIs that do not trigger the Chromium bug. This eliminates the root cause rather than working around it, allowing removal of the `FileCache` workaround entirely.

Changes:
- Switch `onShowFileChooser` intent from `ACTION_GET_CONTENT` to `ACTION_OPEN_DOCUMENT`
- Remove the `Intent.createChooser()` wrapper (unnecessary for `ACTION_OPEN_DOCUMENT`)
- Delete `FileCache.kt` and `FileCacheTest.kt`
- Remove `processFileUris()` from `GutenbergView`
- Remove `FileCache.clearCache()` from `onDetachedFromWindow`
- Simplify `EditorActivity` file picker handler
- Clean up stale detekt baseline entries

## Testing Instructions

1. Open the GutenbergKit demo app on an Android 13+ device.
2. Insert an **Image block** → tap Upload → select an image → verify upload succeeds.
3. Insert a **Gallery block** → tap Upload → select multiple[^1] images → verify upload succeeds.
4. Insert a **Video block** → tap Upload → select a video → verify upload succeeds.
5. Insert an **Audio block** → tap Upload → select an audio file → verify upload succeeds.
6. Insert a **File block** → tap Upload → select any file → verify upload succeeds.
7. For each case, verify no `ERR_UPLOAD_FILE_CHANGED` errors appear in Logcat.

[^1]: Selecting multiple files may require long pressing a file to begin multi-selection. 

N/A — no UI changes within the editor. The system DocumentsUI file picker is used instead of the Photo Picker, but both are system-provided accessible components.

## Screenshots or screencast

Before | After 
-- | -- 
before-2 | after-2


nbradbury (approved):
@dcalhoun It's a shame this change was necessary because the Photo Picker is much friendlier, but it seems there's no alternative. I'll approve this, but note there's a conflict that needs to be resolved before merging :shipit:

dcalhoun:
Thanks, @nbradbury. I agree this is quite unfortunate. I'm discussing this specific workaround with @jkmassel before merging.

jkmassel (approved):

PR #418, opened 2026-04-02
Add Lockdown Mode warning sheet for iOS editor

jkmassel:
## Summary
Ref CMM-2014. 

- Detects iOS Lockdown Mode via `WKWebpagePreferences.isLockdownModeEnabled` after WebView navigation completes
- Presents a dismissible half-sheet warning with a "Learn More" link to Apple's support article on excluding apps
- Re-checks lockdown mode status when the app returns to foreground to handle Settings changes
- All user-facing strings routed through `EditorLocalization` so host apps can provide translations

IMG_0003

## Test plan
- [x] Verify the warning sheet appears when Lockdown Mode is enabled on device
- [x] Verify "Learn More" opens the Apple support article
- [x] Verify "Dismiss" closes the sheet and focuses the editor
- [x] Verify toggling Lockdown Mode exclusion in Settings and returning to the app updates state correctly
- [x] Verify editor autofocus is skipped when lockdown mode is active so the keyboard doesn't overlap the sheet
- [x] Verify normal editor behavior is unaffected when lockdown mode is disabled
- [x] Run `LockdownModeMonitorTests` (CI will do this)

## Known issue
Type some text then switch to settings and back – the contents of the editor are lost.


ios/Sources/GutenbergKit/Sources/Views/LockdownModeSheet.swift
```diff
@@ -0,0 +1,82 @@
+import SwiftUI
+
+#if canImport(UIKit)
+import UIKit
+
+/// A sheet that warns users about Lockdown Mode potentially affecting editor functionality.
+///
+/// Lockdown Mode applies additional security restrictions to WebKit that can
+/// impact the performance and functionality of the Gutenberg editor.
+struct LockdownModeSheet: View {
```
dcalhoun, line 10: Should we rename this file from `Banner` to `Sheet` to match the terminology in the file?

ios/Sources/GutenbergKit/Sources/Services/LockdownModeMonitor.swift
```diff
@@ -0,0 +1,201 @@
...
+        onResetReadyState?()
+
+        // Dismiss the sheet if presented, then reload to re-run detection
+        dismissSheetIfPresented { [weak self] in
+            // Reload triggers navigation delegate which calls detectLockdownMode()
+            // If Lockdown Mode state changed:
+            //   - Now enabled: sheet will be shown
+            //   - Now disabled: editor will fade in normally without sheet
+            self?.detectable?.reloadForLockdownMode()
+        }
```
dcalhoun: > Type some text then switch to settings and back – the contents of the editor are lost.

I'm unsure this is an acceptable trade-off for ensuring the editor state is always accurate regarding Lockdown Mode. This results in the editor always reloading when foregrounded, regardless of whether Lockdown Mode is modified, enabled, or disabled. This feels like a significant downgrade in the UX. 

At best, the editor flashes while reloading and loses any transient state (open modals,  popover placement, text selection, etc). At worst, content is loss. Presumably the latter will not occur if proper content syncing occurs in the host app, but still... 

<details><summary>Editor reload flash</summary>

https://github.com/user-attachments/assets/92c80d88-01b7-4966-8b21-9ad225d0d5bd

</details>

Instead of reloading, should we consider updating the copy to include something like "and re-open the editor" after excluding the app? 

What are your thoughts on this?

ios/Sources/GutenbergKit/Sources/Views/LockdownModeBannerView.swift
```diff
@@ -0,0 +1,82 @@
...
+                    Text(EditorLocalization.localize(.lockdownModeDismiss))
+                        .font(.body)
+                        .foregroundStyle(.primary)
+                        .frame(maxWidth: .infinity)
+                        .padding(.vertical, 14)
+                }
+                .buttonStyle(.plain)
+            }
+            .padding(24)
+            .background(Color(UIColor.systemBackground))
```
dcalhoun: I'm unsure why exactly, but the current sheet design feels off to me. Part of it may be the different background colors between top and bottom sections, where it feels as though the sheet is upside down or imbalanced due to the different "weights" the colors/opacities convey. 

Also, I feel as though the first line of text should have title treatment, with slightly larger text size. 

Should we consider aligning with the layout and styles in Apple's own Lockdown Mode sheet? Just a thought. Not a blocker by any means. 

Our Notice | Lockdown Mode Top | Lockdown Mode Bottom
-- | -- | --
Image  |  Image | Image

ios/Sources/GutenbergKit/Sources/Services/LockdownModeMonitor.swift
```diff
@@ -0,0 +1,201 @@
...
+        let hostingController = UIHostingController(rootView: sheetView)
+        hostingController.modalPresentationStyle = .pageSheet
+        hostingController.isModalInPresentation = true
+
+        if let sheet = hostingController.sheetPresentationController {
+            sheet.detents = [.medium()]
+            sheet.prefersGrabberVisible = false
+        }
+
+        presentingViewController.present(hostingController, animated: true)
```
dcalhoun, line 154: I'm unsure what leads to this, but the sheet is not all that accessible via VoiceOver. When it opens, the focus is moved to the status bar instead of the sheet itself. Afterwards, you cannot swipe to move focus to the sheet; you can only tap or drag your finger atop the sheet to begin reading its contents.

ios/Sources/GutenbergKit/Sources/Views/LockdownModeBannerView.swift
dcalhoun: Also, I just noticed the sheet copy can be cut off  on my iPhone SE. The image above includes "[...] to restore the full functional..." Oddly, I can drag my finger to reveal the full text. 

<details><summary>Drag to reveal full text</summary>
<p>


https://github.com/user-attachments/assets/b5e13a5a-05a0-4e3e-81cd-dc6933e5e9c9



</p>
</details> 

I wonder if this is a flaw in the design of the sheet or possibly related to another oddity I encounter intermittently, where the inserter's media count is truncated into a menu. 

Expected | Truncated 1 | Truncated 2
-- | -- | -- 
IMG_0646 | IMG_0647 | IMG_0648

dcalhoun (commented):
Implementation looks solid, but I shared a few concerns regarding the experience. I'm curious of your thoughts on those.

ios/Sources/GutenbergKit/Sources/Services/LockdownModeMonitor.swift
dcalhoun: Noting that Jeremy and I discussed this some on a call... 

The content loss in the iOS demo app likely occurs because the demo app does not implement a persistence layer alongside using the library's [`editorDidRequestLatestContent`](https://github.com/wordpress-mobile/GutenbergKit/blob/e0d335b310f57606f75cfe20eb54854f1ea291fe/ios/Sources/GutenbergKit/Sources/EditorViewControllerDelegate.swift#L63-L74) function—[WordPress-iOS does](https://github.com/wordpress-mobile/WordPress-iOS/pull/25144). 

Regardless, reloading the editor on every foreground is a bug; it should only reload when Lockdown Mode status changed. 

We should address the unexpected reload, or replace the automatic reload with a note directing the user to re-open the editor.

ios/Sources/GutenbergKit/Sources/Services/LockdownModeMonitor.swift
jkmassel: This should be resolved.

ios/Sources/GutenbergKit/Sources/Views/LockdownModeBannerView.swift
jkmassel: Fixed this by wrapping the text in a scroll view. 

Then just for completeness sake, made the sheet full-screen to hold all the text (since it'll likely be longer after translation).

dcalhoun (approved):
Latest changed tested well for me. The previously reported issues are no longer present. 🚀

PR #419, opened 2026-04-02
fix: improve upload server error handling and memory efficiency

dcalhoun:
> [!note]
> These are improvements to #357 and the PR targets merging into that in-progress branch. 

## What?

Improves the upload server's handling of oversized requests and reduces memory pressure during file uploads.

## Why?

Addresses [PR #357 review feedback](https://github.com/wordpress-mobile/GutenbergKit/pull/357#pullrequestreview-4023619364) from @jkmassel:

1. **413 responses were lost**: When `Content-Length` exceeded the server's limit, the connection was closed immediately — the WebView's `fetch()` never received the 413 response.
2. **413 responses lacked CORS headers**: Even after fixing delivery, the browser blocked the 413 because the server built the response directly, bypassing the handler (which adds CORS headers).
3. **iOS uploads loaded entire files into memory**: `DefaultMediaUploader` used `Data(contentsOf:)` + `httpBody`, causing peak memory of ~2x file size.
4. **Unnecessary multipart re-encoding**: When the delegate didn't modify the file, the pipeline still extracted it to a temp file and re-encoded it into a new multipart body for WordPress.

### Drain oversized request bodies (RFC 9110 §15.5.14)
Adds a `.draining` parser state that reads and discards body bytes without buffering them. The server continues reading until `Content-Length` is satisfied, then sends a clean 413 response. Both iOS and Android parsers.

### Route 413 through the handler for CORS headers
Instead of the server building 413 responses directly, `parseRequest()` now returns parsed headers for `payloadTooLarge` (a non-fatal error — valid headers, rejected body) and exposes the error via a `parseError` property. The server passes it to the handler via a new `serverError` field on the request, letting `MediaUploadServer` build the response with CORS headers — consistent with how OPTIONS preflight is handled.

### Stream multipart upload body from disk (iOS)
Replaces `Data(contentsOf:)` + `httpBody` with a streaming `InputStream` via `httpBodyStream`. The multipart body (preamble, file content, epilogue) is written through a bound stream pair on a background thread, keeping peak memory at ~65 KB regardless of file size. Android already streams via OkHttp's `file.asRequestBody()`.

### Passthrough upload when delegate doesn't modify the file
When `processFile` returns the original file unchanged (e.g., GIFs, non-images, files already within size limits), the original request body is forwarded directly to WordPress via `passthroughUpload()` — skipping multipart re-encoding and the extra file read. Both iOS and Android.

## Testing Instructions

1. Open the iOS and Android demo apps with the Vite dev server running
2. Insert an Image block and select a file larger than the server's max body size
3. Verify the upload error snackbar shows a human-readable message (not a CORS error or connection reset)
4. Upload a normal-sized image and verify it succeeds as before (processed path)
5. Upload a GIF and verify it succeeds (passthrough path — check debug logs for "Passthrough: forwarding original request body to WordPress")

## Screenshots or screencast
N/A — behavioral/performance changes, no UI modifications.

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/MediaUploadServer.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/MediaUploadServer.kt
@@ -78,7 +78,6 @@ internal class MediaUploadServer(
             externallyAccessible = false,
             requiresAuthentication = true,
             cacheDir = cacheDir,
-            errorResponseHeaders = corsHeaders,
             handler = { request -> handleRequest(request) }
         )
         server.start()
@@ -191,6 +190,11 @@ internal class MediaUploadServer(
 
     // MARK: - Response Building
 
+    private val corsHeaders: Map<String, String> = mapOf(
+        "Access-Control-Allow-Origin" to "*",
+        "Access-Control-Allow-Headers" to "Relay-Authorization, Content-Type"
+    )
+
     private fun corsPreflightResponse(): HttpResponse = HttpResponse(
         status = 204,
         headers = corsHeaders + mapOf(
@@ -236,11 +240,6 @@ internal class MediaUploadServer(
 
     companion object {
         private const val TAG = "MediaUploadServer"
-
-        private val corsHeaders: Map<String, String> = mapOf(
-            "Access-Control-Allow-Origin" to "*",
-            "Access-Control-Allow-Headers" to "Relay-Authorization, Content-Type"
-        )
     }
 }
 
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/MediaUploadServer.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/MediaUploadServer.kt
@@ -187,13 +187,7 @@ internal class MediaUploadServer(
                     // Delegate didn't modify the file — forward the original
                     // request body to WordPress without re-encoding.
                     Log.d(TAG, "Passthrough: forwarding original request body to WordPress")
-                    val body = request.body
-                        ?: throw MediaUploadException("Missing request body for passthrough")
-                    val contentType = request.header("Content-Type")
-                        ?: throw MediaUploadException("Missing Content-Type for passthrough")
-                    val uploader = defaultUploader
-                        ?: throw MediaUploadException("No default uploader configured")
-                    uploader.passthroughUpload(body, contentType)
+                    performPassthroughUpload(request)
                 }
             }
             return successResponse(media)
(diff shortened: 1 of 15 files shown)


android/Gutenberg/src/main/java/org/wordpress/gutenberg/MediaUploadServer.kt
```diff
@@ -91,6 +92,16 @@ internal class MediaUploadServer(
...
     private suspend fun handleRequest(request: HttpRequest): HttpResponse {
+        // Server-detected error (e.g., payload too large) — build the
+        // error response here so it includes CORS headers.
+        request.serverError?.let { error ->
+            val message = when (error) {
+                HTTPRequestParseError.PAYLOAD_TOO_LARGE -> "The file is too large to upload in the editor."
+                else -> error.errorId
+            }
+            return errorResponse(error.httpStatus, message)
+        }
```
dcalhoun, line 104: The approach of passing server errors to the handler was taken rather than the adding a new `errorResponseHeaders` configuration (added in 0b6c67be77c1b9b5adb0eda8e5c37d1dca46c0d8, reverted in 0563a94e8dd30d70aeff3cdff1c708863ec39e51). The configuration approach leaks host/caller concerns into the library. A similar [consideration occurred](https://github.com/wordpress-mobile/GutenbergKit/pull/367#discussion_r2956420503) when implementing [authentication bypass](https://github.com/wordpress-mobile/GutenbergKit/blob/09ae7b9016234d2bc6c547a0dfcebd3c0394940f/ios/Sources/GutenbergKitHTTP/HTTPServer.swift#L278-L286) for preflight `OPTIONS` requests. 

Ultimately, the 413 responses need to include the correct CORS headers, otherwise the error code and message do not arrive to the client. It instead receives a CORS error. 

@jkmassel what do you think of this approach?

ios/Sources/GutenbergKit/Sources/Media/MediaUploadServer.swift
```diff
@@ -282,15 +292,11 @@ class DefaultMediaUploader: @unchecked Sendable {
...
 
-        var body = Data()
-        body.append("--\(boundary)\r\n")
-        body.append("Content-Disposition: form-data; name=\"file\"; filename=\"\(filename)\"\r\n")
-        body.append("Content-Type: \(mimeType)\r\n\r\n")
-        body.append(fileData)
-        body.append("\r\n--\(boundary)--\r\n")
+        let (bodyStream, contentLength) = try Self.multipartBodyStream(
+            fileURL: fileURL, boundary: boundary, filename: filename, mimeType: mimeType
+        )
```
dcalhoun, line 344: @jkmassel does this align with your [vision for handling modified](https://github.com/wordpress-mobile/GutenbergKit/pull/357#discussion_r3025707145) files?

src/utils/api-fetch.js
```diff
@@ -202,7 +202,7 @@ export function nativeMediaUploadMiddleware( options, next ) {
 			if ( ! response.ok ) {
 				return response.text().then( ( body ) => {
 					const error = new Error(
-						`Native upload failed (${ response.status }): ${
```
dcalhoun, line 205: Exposing the term "native" to users feels inappropriate. It's an implementation detail.

jkmassel (approved):
I have some minor suggestions in #419. Let's land that, then this, then I think #357 will be ready

dcalhoun:
> I have some minor suggestions in #419. Let's land that, then this, then I think #357 will be ready

Sounds good. I believe you mean #441. That is next on my list to act on today. I'll work to move all three forward.

PR #422, opened 2026-04-03
test(ios): increase URLCache clear polling timeout for CI stability

dcalhoun:
## What?

Increase the `waitForClearToTakeEffect` polling timeout from 1s to 5s and add `.timeLimit(.minutes(1))` to the two clear-related tests in `EditorURLCacheTests`.

## Why?

The "clear removes all entries" and "store succeeds after clear" tests are flaky in CI. `URLCache.removeAllCachedResponses()` is internally asynchronous and CI environments with slower I/O can exceed the previous 1s polling window.

This is the third attempt to address this flakiness:
1. Feb 2026 (`8ac1c734`): Increased `clear()` sleep from 50ms → 200ms
2. Mar 2026 (`dc1c2fad`, #370): Added exponential backoff polling with 1s timeout
3. This PR: Extended polling timeout to 5s, added time limit safety net

## How?

- Increased the polling timeout in `waitForClearToTakeEffect()` from `1.0` to `5.0` seconds
- Added `.timeLimit(.minutes(1))` to both `clearRemovesAll()` and `storeAfterClear()` tests to prevent indefinite hangs

## Testing Instructions

1. Run `swift test --filter EditorURLCacheTests` — all 36 tests should pass.

### DIFF
--- a/ios/Tests/GutenbergKitTests/Stores/EditorURLCacheTests.swift
+++ b/ios/Tests/GutenbergKitTests/Stores/EditorURLCacheTests.swift
@@ -142,7 +142,7 @@ struct EditorURLCacheTests {
 
     // MARK: - clear()
 
-    @Test("clear removes all entries")
+    @Test("clear removes all entries", .timeLimit(.minutes(1)))
     func clearRemovesAll() throws {
         try cache.store(makeResponse(), for: testURL, httpMethod: .GET)
         let otherURL = URL(string: "https://example.com/other")!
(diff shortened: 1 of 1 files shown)


jkmassel (approved):
Root cause here is that `URLCache` is non-deterministic and we should probably just replace it. This seems like a good workaround for now though.

dcalhoun:
Noting there appears to be another flaky test that intermittently fails in CI:

>✖ "maxAge policy: re-storing refreshes the expiration"

PR #392, opened 2026-03-21
ci: automate release notes and PR labeling

dcalhoun:
## What?
Adds GitHub configuration and documentation to automate release notes generation and PR labeling based on Conventional Commits.

## Why?
Release notes were noisy — Dependabot dependency bump PRs dominated the changelog, making it hard to see meaningful changes. There was also no consistent labeling of PRs by type, and the commit/PR conventions lived only in AGENTS.md (AI agent guidance) rather than human-facing docs.

## How?
- **`.github/release.yml`**: Configures GitHub's automatic release notes to exclude Dependabot PRs and group changes into four categories — \"Breaking Changes\", \"Features & Enhancements\", \"Bug Fixes\", and \"Other Changes\" — using existing repository type labels. Noise labels (`[Type] Automated Testing`, `[Type] Build Tooling`, `[Type] Task`, `[Type] Developer Documentation`, `dependencies`) are excluded from \"Other Changes\" to keep the changelog focused on user-visible changes.
- **`.github/workflows/label-pr.yml`**: On PR open or edit, parses the Conventional Commits prefix and breaking change indicator (`!`) from the PR title and applies the matching repository label. When `!` is present (e.g. `feat!:`, `fix(scope)!:`), only `[Type] Breaking Change` is applied. Skips if a conflicting type label is already present, so developers can override the auto-assigned label.
- **`bin/release.sh`**: Release commits now follow the Conventional Commits format (`chore(release): x.y.z`).
- **`docs/code/developer-workflows.md`** (new): Human-facing documentation covering Conventional Commits usage, PR guidelines, and the automatic labeling rules.
- **`docs/releases.md`**: Added a \"Release Notes\" section explaining category groupings, Dependabot exclusions, and linking to the labeling rules.
- **`docs/code/README.md`**: Linked to the new `developer-workflows.md`.
- **`AGENTS.md`**: Replaced inline commit/PR guidelines with a pointer to the new doc.

Conventional Commits → label mapping:
| Prefix | Label |
|---|---|
| `feat` | `[Type] Enhancement` |
| `fix` | `[Type] Bug` |
| `perf` | `[Type] Performance` |
| `test` | `[Type] Automated Testing` |
| `docs` | `[Type] Developer Documentation` |
| `build`, `ci` | `[Type] Build Tooling` |
| `refactor`, `task`, `chore`, `style` | `[Type] Task` |
| any type with `!` | `[Type] Breaking Change` |

## Testing Instructions
[...]

PR #423, opened 2026-04-03
ci: add PR validation check for type label or Conventional Commits title

jkmassel:
## Summary

- Add a required status check (`validate-pr.yml`) that fails if a PR has neither a valid Conventional Commits title prefix nor a `[Type]` label

This closes the gap where PRs without a recognized prefix (e.g. "Move editor loading into the library") silently skip auto-labeling and fall through the cracks in release notes. With this check, every PR is guaranteed to be categorized — including breaking changes.

## Test plan

- [ ] Open a PR with a valid prefix (e.g. `feat: test`) — check should pass
- [ ] Open a PR without a prefix (e.g. "Test PR") — check should fail
- [ ] Manually add a `[Type]` label to the failing PR — check should re-run and pass
- [ ] After merge, add as a required status check in branch protection settings


dcalhoun:
With https://github.com/wordpress-mobile/GutenbergKit/pull/392, I attempted to automate change groups and avoid the need to manually type change log entries, including breaking changes. If a PR is labeled appropriately, its title is included in the GitHub Release notes in the corresponding group. If a PR title follows Conventional Commits spec, a GitHub Action automatically labels the PR appropriately. I.e., the PR title becomes the release note. 

Curious your thoughts: Is there value in a `CHANGELOG.md` over GitHub Releases? PR titles are required for the GitHub flow, is there value in also having separate release note entries? If both are worth keeping, we might consider ensuring manual notes are never forgotten via [scripts like Jetpack's](https://github.com/Automattic/jetpack/blob/c265c6e63c9ee21188bfc0f3123c171750da013d/tools/js-tools/git-hooks/pre-push-hook.mjs#L136-L270). Lasly, should we align the `CHANGELOG.md` organization with the existing GitHub Release organization or remove the latter?

oguzkocer (commented):
I tried testing this with #454 & #455 but I don't think it worked. I branched out of `jkmassel/add-changelog` and opened the PRs against it as I don't see anything in the config about it needing to be against `trunk`. Did I miss something?

jkmassel:
@oguzkocer – sorry about that. I fixed the issue and rebase your test PRs atop this one. They now pass/fail as expected.

oguzkocer (approved):
This looks good to me. Claude brought up some considerations as shared below, but I don't think they are blocking. I'll leave it to you whether these are worth doing and it should be done in this PR.

### Claude's review

1. **Breaking change detection is captured but unused.** `isBreaking` is computed (line 29) but never referenced again. If it's intended for future use, I'd remove it now to avoid dead code — it can be re-added when needed. If it's supposed to do something (like warn or require a `[Type] Breaking Change` label), that logic is missing.

2. **`style` prefix is included but has no corresponding `[Type]` label.** If someone uses `style:` in their title, the check passes, but the auto-labeler (presumably) won't assign a type label. Is that intentional? Same question for `chore:` — is there a label mapping for it?

3. **`typeLabels` list doesn't include all label categories.** For example, there's no `[Type] Code Style` or `[Type] Chore` in the array. If those labels exist, PRs with them wouldn't be recognized by this check. If they don't exist, then the `style` and `chore` prefixes pass validation but can't be categorized by label alone.

jkmassel:
The `chore:` and `style:` ones are dropped from release notes, so it should be fine. Thanks for flagging!

PR #432, opened 2026-04-07
fix(android): include restBase and restNamespace in GBKit post payload

dcalhoun:
## What?

- Adds `restBase` and `restNamespace` to the Android `GBKitGlobal.Post` payload, populated from the configured post type via a small heuristic (`post`→`posts`, `page`→`pages`, custom types pluralized).
- Hardens `filterEndpointsMiddleware` so it falls back to `wp/v2`/`posts` when those fields are missing from `window.GBKit.post`, matching the contract `getPost()` already provides.

## Why?

`filterEndpointsMiddleware` (in `src/utils/api-fetch.js`) failed blocking server fetches for the post being edited. This resulted in the WordPress core data fetched via `GET /wp/v2/posts/{id}` clearing the title and content set from the native host.

iOS already populates these fields from `EditorConfiguration.postType` (`PostTypeDetails`). This PR brings Android in line and adds a defensive fallback in the JS middleware so any host that omits the fields still gets the filter applied.

## How?

- `android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/GBKitGlobal.kt` — adds the two new fields to `Post`, populates them in `fromConfiguration`, and introduces a private `restBaseFor(postType)` helper with the pluralization heuristic.
- `src/utils/bridge.js` — extracts a new exported `POST_FALLBACKS` constant so `getPost()` and the middleware share one source of truth for default values.
- `src/utils/api-fetch.js` — `filterEndpointsMiddleware` now bails out only when there is no post or no post id, and falls back to `POST_FALLBACKS` for `restBase`/`restNamespace`.
- `src/utils/api-fetch.test.js` — switches the bridge mock to a partial mock so `POST_FALLBACKS` survives, and adds four new cases covering the filter with/without explicit fields and the bail-out paths.
- `android/Gutenberg/src/test/java/org/wordpress/gutenberg/model/GBKitGlobalTest.kt` — asserts `restBase`/`restNamespace` for `post`, `page`, and a custom post type.

## Testing Instructions

Since the Android demo app cannot yet edit existing posts, a manual patch is required to verify the fix end-to-end.

1. Apply a manual patch with test title/content and valid post ID from a site. 
    ```diff
    diff --git a/android/app/src/main/java/com/example/gutenbergkit/SitePreparationViewModel.kt b/android/app/src/main/java/com/example/gutenbergkit/SitePreparationViewModel.kt
    index bb1a6c81..35a8dcf3 100644
    --- a/android/app/src/main/java/com/example/gutenbergkit/SitePreparationViewModel.kt
    +++ b/android/app/src/main/java/com/example/gutenbergkit/SitePreparationViewModel.kt
    @@ -231,6 +231,7 @@ class SitePreparationViewModel(
                arrayOf()
            }
    
    +        val postId: UInt = 0000u;
            return EditorConfiguration.builder(
                siteURL = config.siteUrl,
                siteApiRoot = siteApiRoot,
    @@ -241,8 +242,9 @@ class SitePreparationViewModel(
                .setSiteApiNamespace(siteApiNamespace)
                .setNamespaceExcludedPaths(arrayOf())
                .setAuthHeader(config.authHeader)
    -            .setTitle("")
    -            .setContent("")
    +            .setPostId(postId)
    +            .setTitle("Test")
    +            .setContent("<!-- wp:paragraph --><p>Hello</p><!-- /wp:paragraph -->")
                .setHideTitle(false)
                .setCookies(emptyMap())
                .setEnableNetworkLogging(true)

    ```
1. Replace `0000` with a valid post ID for your test site. 
1. Open the editor for your test site via the Start button. 
1. Wait a few seconds for initial fetches to complete.
1. Verify the editor title and content is not cleared. 

No UI changes; not applicable.

## Screenshots or screencast

N/A — bug fix with no visible UI changes.

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/GBKitGlobal.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/GBKitGlobal.kt
@@ -74,6 +74,10 @@ data class GBKitGlobal(
         val id: Int, // TODO: Instead of the `-1` trick, this should just be `null` for new posts
         /** The post type (e.g., `post`, `page`). */
         val type: String,
+        /** The REST API base path for this post type (e.g., `posts`, `pages`). */
+        val restBase: String,
+        /** The REST API namespace (e.g., `wp/v2`). */
+        val restNamespace: String,
         /** The post status (e.g., `draft`, `publish`, `pending`). */
         val status: String,
         /** The post title (URL-encoded). */
@@ -110,6 +114,8 @@ data class GBKitGlobal(
                 post = Post(
                     id = postId ?: -1,
                     type = configuration.postType,
+                    restBase = restBaseFor(configuration.postType),
+                    restNamespace = "wp/v2",
                     status = configuration.postStatus,
                     title = configuration.title.encodeForEditor(),
                     content = configuration.content.encodeForEditor()
@@ -125,6 +131,18 @@ data class GBKitGlobal(
                     }
                 }
             )
+        }
+
+        /**
+         * Maps a post type slug to its WordPress REST API base path.
+         *
+         * Defaults to pluralizing the slug for unknown types (e.g., `product` → `products`),
+         * which matches the WordPress convention for most post types.
+         */
+        private fun restBaseFor(postType: String): String = when (postType) {
+            "post" -> "posts"
+            "page" -> "pages"
+            else -> if (postType.endsWith("s")) postType else "${postType}s"
         }
     }
 
--- a/src/utils/api-fetch.js
+++ b/src/utils/api-fetch.js
@@ -111,6 +111,18 @@ function tokenAuthMiddleware( options, next ) {
  * Middleware to filter out requests to specific endpoints.
  *
  * @type {APIFetchMiddleware}
+ *
+ * @todo Properly seed the post entity and remove this middleware.
+ *
+ * This was added to prevent re-fetching entity content provided by the native
+ * host app, which can lead to content loss. However, we can likely avoid the
+ * need for this middleware by ensuring we properly seed the entity content into
+ * the store on initialization.
+ *
+ * This requires hoisting the relevant logic from `useEditorSetup` to occur
+ * before we render the editor, and invoking `finishResolution`.
+ *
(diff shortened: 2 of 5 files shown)


android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/GBKitGlobal.kt
```diff
@@ -126,6 +132,18 @@ data class GBKitGlobal(
...
+         * Maps a post type slug to its WordPress REST API base path.
+         *
+         * Defaults to pluralizing the slug for unknown types (e.g., `product` → `products`),
+         * which matches the WordPress convention for most post types.
+         */
+        private fun restBaseFor(postType: String): String = when (postType) {
+            "post" -> "posts"
+            "page" -> "pages"
+            else -> if (postType.endsWith("s")) postType else "${postType}s"
+        }
```
dcalhoun, line 146: This will be removed and replaced with fetching of `PostTypeDetails` in a future PR focused on editing existing posts (and align with iOS). In the interim, this ensures Android correctly sets the `restBase` value for posts and pages.

src/utils/api-fetch.js
```diff
@@ -111,16 +111,31 @@ function tokenAuthMiddleware( options, next ) {
...
+ *
+ * This was added to prevent re-fetching entity content provided by the native
+ * host app, which can lead to content loss. However, we can likely avoid the
+ * need for this middleware by ensuring we properly seed the entity content into
+ * the store on initialization.
+ *
+ * This requires hoisting the relevant logic from `useEditorSetup` to occur
+ * before we render the editor, and invoking `finishResolution`.
+ *
+ * See: https://github.com/wordpress-mobile/GutenbergKit/commit/c9b4fc9978a3760ba97f3f5d4359c2bc2155bb80
```
dcalhoun, line 125: In the future, we can improve entity seeding and negate the need this middleware that manually disables the entity fetch requests. See https://github.com/wordpress-mobile/GutenbergKit/commit/c9b4fc9978a3760ba97f3f5d4359c2bc2155bb80 as a proof-of-concept.

jkmassel (approved):
This fixes the immediate problem and we'll revisit the further improvements

PR #433, opened 2026-04-08
feat(demo): edit existing posts (Android + iOS)

dcalhoun:
## What?

Adds a cross-platform "edit existing posts" feature to the iOS and Android demo apps:

- A **Posts list** screen for browsing existing posts of the currently selected post type
- A **Post type picker** on the site preparation screen (Post / Page, plus any custom post types discovered via REST)
- A **Save button** in the editor toolbar that PUTs the edited title/content back to the site via the WordPress REST API
- **PostTypeDetails** — a Kotlin data class mirroring the iOS `PostTypeDetails` struct, carrying the slug + REST base + namespace together. Threaded through `EditorConfiguration`, `EditorPreloadList`, `GBKitGlobal`, and `EditorService` so custom post types get the correct REST routing instead of a heuristic pluralization
- Demo-app cleanup: removes permanently-disabled toolbar items and overflow menu placeholders on both platforms

## Why?

GutenbergKit's demo apps previously only supported creating new posts — there was no way to load an existing post, edit it, and persist the changes back to the server. That makes it hard to dogfood the editor against real content and hard to reproduce bugs that only occur on existing posts.

### Save flow (REST-only in this PR)

Both demos persist title/content via `wordpress-rs`' `posts.update()` directly. 

### Android library changes

- **New:** `PostTypeDetails.kt` data class with `post`/`page` companion constants, mirroring iOS 1:1.
- **Breaking:** `EditorConfiguration.postType` is now `PostTypeDetails` instead of `String`. `EditorConfiguration.builder(siteURL, siteApiRoot, postType)` now takes a `PostTypeDetails` parameter. Host apps will need to pass `PostTypeDetails.post` / `PostTypeDetails.page` (or a custom-constructed instance) instead of a string slug. This **mirrors iOS' [current state](https://github.com/wordpress-mobile/GutenbergKit/blob/eaba3dd7d66920867de6e638e12ac2409eaaa5b2/ios/Sources/GutenbergKit/Sources/Model/EditorConfiguration.swift#L29)** after [#299](https://github.com/wordpress-mobile/GutenbergKit/pull/299). 
- `EditorPreloadList.buildPostPath()` now uses the post type's `restNamespace`/`restBase` instead of hardcoded `/wp/v2/posts/\$id`, fixing a latent bug where editing a page would preload the wrong path.
- `GBKitGlobal.Post` populates `restBase`/`restNamespace` directly from the post type details instead of the `restBaseFor()` heuristic introduced in #432.

### Android demo changes

[...]

Issue #434, opened 2026-04-08
Prevent post data refetch with finishResolution instead of middleware filtering


jkmassel:
## Summary

`filterEndpointsMiddleware` in `src/utils/api-fetch.js` exists to block Gutenberg's automatic `GET /wp/v2/{restBase}/{id}` fetch, which overwrites the title and content injected by the native host. This middleware requires `restBase` and `restNamespace` to construct the path to block, which adds complexity to the native-to-JS bridge contract (see #432).

The root cause is that `receiveEntityRecords` in `useEditorSetup` puts post data into the store but does **not** mark the resolution as finished. Gutenberg's data layer treats "data exists" and "resolution is finished" as independent — so when a component calls `getEntityRecord('postType', type, id)`, the resolver fires a network fetch even though the data is already present.

## Proposed fix

After pre-seeding the post data in `src/components/editor/use-editor-setup.js`, call `finishResolution` to tell the data layer the record is already resolved:

```js
receiveEntityRecords('postType', post.type, post);
finishResolution('getEntityRecord', ['postType', post.type, post.id]);
```

WordPress core uses this same pattern internally — see `@wordpress/core-data/src/resolvers.js` where `finishResolution` is called after `receiveEntityRecords` for navigation fallbacks and templates.

## Expected outcome

- Gutenberg's `getEntityRecord` resolver skips the network fetch entirely since the resolution is already marked as finished.
- `filterEndpointsMiddleware` can be removed (or simplified), along with the JS-side dependency on `restBase`/`restNamespace` in the middleware.
- The `POST_FALLBACKS` constant in `bridge.js` may be simplified since the middleware no longer needs fallback values for path construction.

## Context

- #432 added `restBase`/`restNamespace` to the Android payload and hardened the middleware with fallback defaults. That fix is correct and addresses the immediate bug. This issue tracks the deeper fix that eliminates the need for the middleware approach.
- The Android library also has hardcoded `/wp/v2/posts/` paths in `RESTAPIRepository.buildPostUrl()` and `EditorPreloadList.buildPostPath()` — those are separate from this issue but related parity work (iOS already uses dynamic `restBase`/`restNamespace` for both).

dcalhoun:
[...]

PR #435, opened 2026-04-08
feat: trigger editor savePost() lifecycle from native hosts

dcalhoun:
## What?

> [!note]
> This stacks atop #433. The ability to edit existing posts is required for testing the `savePost()` lifecycle. 

Adds a `savePost()` lifecycle bridge so native hosts can trigger the WordPress `core/editor` store's save flow without persisting content via the editor.

## Why?

Ref CMM-2024. 

Plugins like VideoPress need to fire side-effect API calls (e.g., metadata syncing) when the editor's save lifecycle runs. Given the native host app, not the web editor, manages saving, those side-effects never ran. 

Exposing the lifecycle as a bridge method that hosts can call allows host-driven saves to trigger the same `core/editor` store flow that plugins subscribe to, so plugin side-effects fire as expected.

The bridge is **lifecycle-only**: it does not persist post content. Hosts remain responsible for reading content via `getTitleAndContent()` and PUTting it through their own REST API calls. The existing `filterEndpointsMiddleware` (hardened in #432) already swallows the editor's own post-save HTTP request, so calling `savePost()` does not cause a duplicate REST round-trip.

## How?

- **JS** dispatches `dispatch('core/editor').savePost()` and removes the `editor-save` notice in a `finally` block. The call is added to the host bridge's exposed `window.editor` API and cleaned up on hook unmount.
- **iOS** uses `callAsyncJavaScript` so the Promise returned by `editor.savePost()` is awaited natively. A documented contract clarifies that lifecycle ≠ persistence and that plugin failures must not block saves.
- **Android** can't await Promises through `evaluateJavascript`, so the bridge dispatches the call with a UUID request ID, captures completion in `editorDelegate.onSavePostComplete(requestId, success, error)`, and routes it back to the per-request callback. `onDetachedFromWindow()` drains any pending callbacks with `success = false` so coroutines awaiting the bridge don't hang on view teardown.

## Testing Instructions

This PR is best tested by editing a post that contains a VideoPress block, since VideoPress is the canonical plugin that depends on the save lifecycle for metadata sync.

**Prerequisites:**
- A WordPress.com (or Jetpack-connected) site with VideoPress enabled
- An existing post on that site containing at least one VideoPress block
- The site added to the demo app via OAuth[^1] (WP.com) or Application Passwords (self-hosted)
- Temporarily allow the VideoPress block type:

    ```diff
    diff --git a/src/utils/blocks.js b/src/utils/blocks.js
    index 52a39f97..fd47f2ab 100644
    --- a/src/utils/blocks.js
    +++ b/src/utils/blocks.js
    @@ -17,7 +17,10 @@ export function unregisterDisallowedBlocks( allowedBlockTypes ) {
    
      const unregisteredBlocks = [];
      getBlockTypes().forEach( ( block ) => {
    -		if ( ! allowedBlockTypes.includes( block.name ) ) {
    +		if (
    +			! allowedBlockTypes.includes( block.name ) &&
    +			block.name !== 'videopress/video'
    +		) {
          unregisterBlockType( block.name );
          unregisteredBlocks.push( block.name );
        }

    ```

[^1]: Certain VideoPress requests fail with when using the dev server due to CORS, use `make build` instead. 

**Steps:**

1. Open the demo app (iOS or Android) and authenticate with the test site
2. From the site preparation screen, select **Post** as the Post Type and tap **Browse** 
3. From the posts list, tap the post containing the VideoPress block
4. In the editor, tap the VideoPress block to select it, then open the block inspector
5. Change the **Rating** from its current value (e.g., G → PG-13)
6. Tap the **Save** button in the top-right toolbar
7. Wait for the save to complete (the button briefly disables)
8. Open the same post in the WordPress site's web editor 
9. Inspect the VideoPress block — **verify the Rating reflects the value you set in step 5**

N/A, no navigation changes. 

## Screenshots or screencast

https://github.com/user-attachments/assets/d5d826a5-ea48-488e-90b2-946fb23dc680

### DIFF
--- a/src/components/editor/use-host-bridge.js
+++ b/src/components/editor/use-host-bridge.js
@@ -5,6 +5,7 @@ import { useEffect, useCallback, useRef } from '@wordpress/element';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { store as coreStore } from '@wordpress/core-data';
 import { store as editorStore } from '@wordpress/editor';
+import { store as noticesStore } from '@wordpress/notices';
 import { parse, serialize, getBlockType } from '@wordpress/blocks';
 import { store as blockEditorStore } from '@wordpress/block-editor';
 import { insert, create, toHTMLString } from '@wordpress/rich-text';
@@ -20,6 +21,7 @@ export function useHostBridge( post, editorRef, markBridgeReady ) {
 	const { editEntityRecord } = useDispatch( coreStore );
 	const { undo, redo, switchEditorMode, savePost } =
 		useDispatch( editorStore );
+	const { removeNotice } = useDispatch( noticesStore );
 	const { getEditedPostAttribute, getEditedPostContent } =
 		useSelect( editorStore );
 	const { updateBlock, selectionChange } = useDispatch( blockEditorStore );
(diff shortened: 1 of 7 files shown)


src/components/editor/use-host-bridge.js
```diff
@@ -88,6 +91,18 @@ export function useHostBridge( post, editorRef, markBridgeReady ) {
 			redo();
 		};
 
+		window.editor.triggerSaveLifecycle = async () => {
+			try {
+				// Await the lifecycle so hosts can sequence persistence after
+				// plugin side-effects settle, do not return the `Promise` return value
+				// to avoid host errors.
+				await savePost();
```
dcalhoun, line 104: Noting that invoking `savePost()` may mean we _cannot_ remove the `filterEndpointsMiddleware` as referenced in https://github.com/wordpress-mobile/GutenbergKit/pull/432#discussion_r3051853903 and https://github.com/wordpress-mobile/GutenbergKit/issues/434. 

It may still be worthwhile to properly seed the content, but I believe invoking `savePost()` triggers a `PUT` request for the same endpoint. There may also be defense in depth by retaining the middleware.

dcalhoun:
Rebased now that #433 merged.

Issue #439, opened 2026-04-08
iOS: MediaUploadServer tests silently skip when server can't bind


jkmassel:
## Summary

`MediaUploadServerTests` uses `.enabled(if: _canStartUploadServer)` to skip the entire suite if `HTTPServer` can't bind a port. This means test failures due to binding issues are invisible — the suite reports as skipped rather than failing.

## Location

`ios/Tests/GutenbergKitTests/Media/MediaUploadServerTests.swift` (lines 7-32)

## Current behavior

A static check tries to start and stop a server at import time. If it fails, all tests in the suite are silently skipped via Swift Testing's `.enabled(if:)` trait.

## Suggested behavior

Tests should fail, not skip, when the server can't start. If there are known environments where binding is expected to fail (e.g., specific CI sandboxes), those should be identified explicitly rather than using a blanket skip.

Issue #438, opened 2026-04-08
iOS: bound stream writer threads silently swallow errors


jkmassel:
## Summary

The background writer threads used for bound stream pairs silently swallow all errors. If a `FileHandle.read()` fails mid-stream or `OutputStream.write()` returns an error, the thread exits without logging or propagating the failure. The output stream closes, the input stream sees EOF, and `URLSession` receives fewer bytes than `Content-Length` promised — resulting in a network error with no indication of the root cause.

## Locations

**On `trunk`:**
- `RequestBody.makePipedFileSliceStream` — `ios/Sources/GutenbergKitHTTP/RequestBody.swift` (~line 217)

**On `feat/leverage-host-media-processing` (PR #419):**
- `DefaultMediaUploader.multipartBodyStream` — `ios/Sources/GutenbergKit/Sources/Media/MediaUploadServer.swift` (same pattern, copied from above)

## The pattern

```swift
Thread.detachNewThread {
    defer {
        output.close()
        try? fileHandle.close()
    }

    var remaining = length
    while remaining > 0 {
        let chunkSize = min(65_536, remaining)
        // FileHandle.read error silently swallowed by try?
        guard let chunk = try? fileHandle.read(upToCount: chunkSize),
              !chunk.isEmpty else {
            break  // silent exit — no logging
        }
        // write error — silent exit
        guard Self.writeAll(chunk, to: output) else { return }
        remaining -= chunk.count
    }
}
```

## Problems

1. **No logging**: When the writer thread fails, there is no log message indicating what went wrong (I/O error, stream closed, etc.), making production issues hard to diagnose.
2. **No error propagation**: The caller has no way to distinguish "stream completed successfully" from "stream failed after 50% of the file." Both look like the output stream closing.
3. **Root cause masked**: `URLSession` will report a generic network/content-length mismatch error, hiding the actual cause (disk I/O failure, file deleted, permissions changed, etc.).

## Suggested improvements

- Log errors from `FileHandle.read()` and `OutputStream.write()` before exiting
- Consider using `os.Logger` consistent with other upload server logging
- Optionally, track whether the writer completed successfully so callers can report a more specific error

Issue #437, opened 2026-04-08
iOS: OutputStream.write return value 0 treated as fatal in bound stream writers


jkmassel:
## Summary

The `writeAll` pattern used in bound-stream-pair writers treats `OutputStream.write(_:maxLength:)` returning `0` as a fatal error. Apple's documentation defines a `0` return as "a fixed-length stream has reached its capacity." It is unclear whether a bound stream pair (from `Stream.getBoundStreams(withBufferSize:)`) qualifies as a "fixed-length stream," meaning `write()` could return `0` when the internal buffer is full rather than blocking.

If `0` is returned as a backpressure signal, the current code silently aborts the write, closes the stream, and leaves `URLSession` waiting for `Content-Length` bytes that never arrive — causing the upload to hang.

## Locations

1. **`RequestBody.makePipedFileSliceStream`** — `ios/Sources/GutenbergKitHTTP/RequestBody.swift` (~line 236)
2. **`DefaultMediaUploader.writeAll`** — `ios/Sources/GutenbergKit/Sources/Media/MediaUploadServer.swift` (introduced in #419, same pattern)

Both use:
```swift
let result = output.write(base.advanced(by: written), maxLength: ...)
if result <= 0 { return }
```

## Apple documentation

- [`OutputStream.write(_:maxLength:)`](https://developer.apple.com/documentation/foundation/outputstream/1410720-write): returns 0 for "a fixed-length stream that has reached its capacity"
- [`CFWriteStreamWrite`](https://developer.apple.com/documentation/corefoundation/1539680-cfwritestreamwrite): "If the stream is not full, this call blocks until at least one byte is written"
- Neither document explicitly states whether bound stream pairs block or return 0 when the buffer is full

## Suggested fix

Distinguish `0` (retry) from `-1` (error):

```swift
if result == -1 { return false }
if result == 0 {
    Thread.sleep(forTimeInterval: 0.001)
    continue
}
```

## Risk

Low probability in practice — uploads would need to outpace `URLSession`'s read rate long enough to fill the 65KB buffer. More likely with large files on slow connections.

PR #441, opened 2026-04-08
fix(ios): review adjustments for #419

jkmassel:
Suggested changes from reviewing #419.

### 1. Use `Int64` for `HTTPRequestParser.bytesWritten`

`bytesWritten` was `Int` while `expectedContentLength` and `maxBodySize` are already `Int64`. This is consistent on all Apple platforms today (where `Int` is 64-bit), but using `Int64` explicitly matches the convention used throughout the parser and avoids unnecessary `Int64(...)` casts at comparison sites.

### 2. Add end-to-end test for 413 + CORS headers

The core motivation for the drain+handler routing in #419 is that 413 responses include CORS headers so the WebView's `fetch()` can read them. This adds an integration test that sends an oversized request to `MediaUploadServer` and verifies both the 413 status code and the `Access-Control-Allow-Origin` header.

To support this, `MediaUploadServer.start()` now exposes a `maxRequestBodySize` parameter (defaulting to the existing 4 GB), so the test can use a small limit without sending gigabytes of data.

## Related issues

- #437 — pre-existing: `OutputStream.write` return value 0 treated as fatal in bound stream writers
- #438 — pre-existing: bound stream writer threads silently swallow errors
- #439 — pre-existing: `MediaUploadServer` tests silently skip when server can't bind

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/http/HTTPRequestParser.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/http/HTTPRequestParser.kt
@@ -126,10 +126,7 @@ class HTTPRequestParser(
         // whether the full Content-Length has been consumed.
         if (_state == State.DRAINING) {
             bytesWritten += data.size.toLong()
-            val offset = headerEndOffset
-            if (offset != null && bytesWritten - offset >= expectedContentLength) {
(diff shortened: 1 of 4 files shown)


dcalhoun (approved):
Changes tested well for me—both successful uploads and 413 error messages. Thank you!

PR #462, opened 2026-04-22
feat!: seed user capabilities from host to avoid cross-origin canUser regression

jkmassel:
## What?

Introduces a `useSeededPermissions` hook that reads host-supplied `userCapabilities` from `window.GBKit` and preseeds the `@wordpress/core-data` store so `canUser( ..., { kind: 'postType', name: 'attachment' } )` returns `true` without an OPTIONS request. Extends the native iOS/Android SDKs with a matching `UserCapabilities` type that flows through `EditorConfiguration` into `GBKitGlobal`, and extends the `wp-env` CORS mu-plugin to expose the `Allow` header and route REST `OPTIONS` through `WP_REST_Server` so the local environment matches production CORS behaviour.

**Breaking change**: `userCapabilities` is a required parameter on `EditorConfiguration.init` / `EditorConfigurationBuilder` (iOS) and `EditorConfiguration.Builder` / `EditorConfiguration.builder(...)` (Android). Every call site must pass `UserCapabilities(uploadFiles: <Bool>)` explicitly — a missing argument becomes a compile error rather than a silent `uploadFiles = false` default that hides the upload button.

## Why?

`@wordpress/core-data` 7.42.0 ([Gutenberg #76307](https://github.com/WordPress/gutenberg/pull/76307)) changed `canUser` to return `false` instead of `undefined` when the REST `Allow` header is missing. This breaks `@wordpress/editor`'s `hasUploadPermissions = canUser(...) ?? true` fallback — when `canUser` returns `false`, `??` doesn't fall through, `settings.mediaUpload` becomes `undefined`, `MediaUploadCheck` evaluates falsy, and **the Upload button disappears from every MediaPlaceholder**.

GutenbergKit hits this in production, not just in wp-env. The editor is a local bundle (`bundle://` / `appassets.androidplatform.net`) and the REST API lives on the user's WordPress site, so every request is cross-origin. Browsers hide response headers from JS unless the server lists them in `Access-Control-Expose-Headers`, and WordPress core's default `rest_send_cors_headers()` only exposes `X-WP-Total, X-WP-TotalPages, Link` — not `Allow`. Confirmed empirically against `https://vanilla.wpmt.co/wp-json/wp/v2/media`.

Fixing WP core's CORS allowlist is a Trac-level change with a multi-release timeline, and reverting #76307 upstream is unlikely. This PR must land before the pending Dependabot bump at #456 (`@wordpress/editor` 14.41.0 → 14.44.0), otherwise every cross-origin GBKit host silently regresses.

Required-vs-optional: an earlier iteration made `userCapabilities` optional with a `uploadFiles = false` default. That silently opted every un-updated host into the broken behaviour — the opposite of the fix. Making the parameter required forces every integration to consciously declare what the signed-in user can do, which is the whole point of the seeded-permissions pathway.

### What we explored

1. **Patch `@wordpress/core-data` to restore the `undefined` behaviour.** ❌ Fragile, drifts on every upstream bump, ignores upstream intent.
2. **Land the bump and fix via a WP core CORS change.** ❌ Trac-level change on a multi-release timeline; only helps sites that update.
3. **Seed `canUser` from host-declared capabilities.** ✅ The host already authenticates the user and knows their capabilities via `GET /wp/v2/users/me?context=edit&_fields=capabilities`. Passing `userCapabilities` into the editor via the existing `window.GBKit` bridge and dispatching `receiveUserPermissions` + `finishResolutions` into the core-data store bypasses the OPTIONS inference entirely.

### Changes

**JS**

- **`src/components/editor/use-seeded-permissions.js`**: new hook reading `userCapabilities.uploadFiles` from `getGBKit()` and dispatching seed data into the core-data store on mount. Uses only public core-data actions (`receiveUserPermissions`, `finishResolutions`) — no patches. Additive: if the host doesn't supply capabilities, the editor falls back to today's path.
- **`src/components/editor/index.jsx`**: calls `useSeededPermissions()` before `<EditorProvider>` mounts.
- **`src/utils/bridge.js`**: documents the new `userCapabilities` field on the `GBKitConfig` typedef.
- **`wp-env/mu-plugins/gutenbergkit-cors.php`**: exposes `Allow` to cross-origin JS; routes REST `OPTIONS` through WP's dispatcher (non-REST `OPTIONS` are still short-circuited).

**Native SDKs**

- **`UserCapabilities` type** on both iOS (Swift) and Android (Kotlin) with a single required `uploadFiles: Bool` field (designed to grow).
- **`EditorConfiguration`** on both platforms carries a required `userCapabilities`, which flows through `GBKitGlobal` into the JS bridge.
- **`EditorConfigurationBuilder`** (iOS) and `EditorConfiguration.Builder` / `EditorConfiguration.builder(...)` (Android) require `userCapabilities` as a constructor argument — a host app that forgets to supply it gets a compile error telling it exactly what to add.
- Demo apps, tests, and `docs/integration.md` / `docs/code/preloading.md` updated to pass `userCapabilities` at every call site.

Host-app changes (fetching capabilities after auth and injecting into the `window.GBKit` payload) are out of scope here and will ship as separate coordinated PRs on WordPress iOS/Android and Jetpack iOS/Android.

## Testing Instructions

- [ ] `make test-js` — all 136 existing tests plus 5 new `useSeededPermissions` tests pass
- [ ] `make test-swift-package` — passes clean
- [ ] `make test-android` — passes clean
- [ ] `make lint-js-fix` — clean
- [ ] `make build` — builds without errors
- [ ] E2E: `npx playwright test image-upload audio-upload video-upload file-upload cover-block gallery-block` — all pass
- [ ] Manual: set `window.GBKit.userCapabilities = { uploadFiles: true }` via devtools before the editor mounts; verify Upload button appears in the Image block
- [ ] Manual: omit `userCapabilities` entirely; verify editor falls back to existing `canUser` resolver path without crashing
- [ ] Verify no OPTIONS request is made to `/wp/v2/media` when the seed is active (Network tab)
- [ ] Verify the iOS/Android demo apps pass a realistic `UserCapabilities(uploadFiles: true)` through the authenticated flow and `uploadFiles: false` through the bundled/offline flow

## Related issues

- Precedes the Dependabot bump at #456 (must land first).
- Upstream: https://github.com/WordPress/gutenberg/pull/76307

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorConfiguration.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorConfiguration.kt
@@ -29,7 +29,7 @@ data class EditorConfiguration(
     val editorAssetsEndpoint: String? = null,
     val enableNetworkLogging: Boolean = false,
     var enableOfflineMode: Boolean = false,
-    val userCapabilities: UserCapabilities = UserCapabilities()
+    val userCapabilities: UserCapabilities
 ): Parcelable {
 
     /**
@@ -47,15 +47,30 @@ data class EditorConfiguration(
     }
     companion object {
         @JvmStatic
-        fun builder(siteURL: String, siteApiRoot: String, postType: PostTypeDetails = PostTypeDetails.post): Builder = Builder(siteURL, siteApiRoot, postType = postType)
+        fun builder(
+            siteURL: String,
+            siteApiRoot: String,
+            userCapabilities: UserCapabilities,
+            postType: PostTypeDetails = PostTypeDetails.post
+        ): Builder = Builder(siteURL, siteApiRoot, userCapabilities, postType = postType)
 
         @JvmStatic
-        fun bundled(): EditorConfiguration = Builder("https://example.com", "https://example.com/wp-json/", PostTypeDetails.post)
+        fun bundled(): EditorConfiguration = Builder(
+            "https://example.com",
+            "https://example.com/wp-json/",
+            UserCapabilities(uploadFiles = false),
+            PostTypeDetails.post
+        )
             .setEnableOfflineMode(true)
             .build()
     }
 
-    class Builder(private var siteURL: String, private var siteApiRoot: String, private var postType: PostTypeDetails) {
+    class Builder(
+        private var siteURL: String,
+        private var siteApiRoot: String,
+        private var userCapabilities: UserCapabilities,
+        private var postType: PostTypeDetails
+    ) {
         private var title: String = ""
         private var content: String = ""
         private var postId: UInt? = null
@@ -74,7 +89,6 @@ data class EditorConfiguration(
         private var editorAssetsEndpoint: String? = null
         private var enableNetworkLogging: Boolean = false
         private var enableOfflineMode: Boolean = false
-        private var userCapabilities: UserCapabilities = UserCapabilities()
 
         fun setTitle(title: String) = apply { this.title = title }
         fun setContent(content: String) = apply { this.content = content }
@@ -130,7 +144,7 @@ data class EditorConfiguration(
      *
      * This allows modifying specific fields while preserving others.
      */
-    fun toBuilder(): Builder = Builder(siteURL, siteApiRoot, postType)
+    fun toBuilder(): Builder = Builder(siteURL, siteApiRoot, userCapabilities, postType)
         .setTitle(title)
         .setContent(content)
         .setPostId(postId)
@@ -149,7 +163,6 @@ data class EditorConfiguration(
         .setEditorAssetsEndpoint(editorAssetsEndpoint)
         .setEnableNetworkLogging(enableNetworkLogging)
         .setEnableOfflineMode(enableOfflineMode)
-        .setUserCapabilities(userCapabilities)
 
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorConfiguration.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorConfiguration.kt
@@ -28,7 +28,8 @@ data class EditorConfiguration(
     val cachedAssetHosts: Set<String> = emptySet(),
     val editorAssetsEndpoint: String? = null,
     val enableNetworkLogging: Boolean = false,
-    var enableOfflineMode: Boolean = false
+    var enableOfflineMode: Boolean = false,
+    val userCapabilities: UserCapabilities = UserCapabilities()
 ): Parcelable {
 
     /**
@@ -73,6 +74,7 @@ data class EditorConfiguration(
         private var editorAssetsEndpoint: String? = null
         private var enableNetworkLogging: Boolean = false
         private var enableOfflineMode: Boolean = false
+        private var userCapabilities: UserCapabilities = UserCapabilities()
 
         fun setTitle(title: String) = apply { this.title = title }
         fun setContent(content: String) = apply { this.content = content }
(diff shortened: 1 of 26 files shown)


dcalhoun:
Noting that I'm landing `@wordpress` dependency updates in https://github.com/wordpress-mobile/GutenbergKit/pull/691. That implements the "Patch @wordpress/core-data" approach discarded by this PR. 

I believe the seeding approach of this PR is still worth exploring. The patch in https://github.com/wordpress-mobile/GutenbergKit/pull/691 unblocks updates and assumes the user has upload permissions; seeding would improve correctness by checking user permissions. When we re-engage seeding, we should...

1. Rebase it atop trunk;
2. Drop the wp-env `mu-plugin` allowing the `Allows` header for CORS (diverges from production environments);
3. And document that the seeding approach remains a _proximation_ given it uses a different mechanism than Gutenberg core (site-wide `upload_files` user property vs REST API endpoint `permission_callback`).  

It's worth investigating if the third item impacts discussed efforts to lower post saving down from native host → GBK's WebView (rely upon Gutenberg core saving mechanisms). E.g., will "this Author may edit their own post but not someone else's" pose a problem. If it does, options 3, 4, and 5 outlined in https://github.com/wordpress-mobile/GutenbergKit/pull/691 may present alternatives for addressing it. 

More details on various approaches considered can be found in the "Agent details" section of https://github.com/wordpress-mobile/GutenbergKit/pull/691.

dcalhoun:
Follow up to https://github.com/wordpress-mobile/GutenbergKit/pull/462#issuecomment-5734898059: I ultimately [replaced the patch with a middleware](https://github.com/wordpress-mobile/GutenbergKit/pull/691#issuecomment-5764942679). Doing so addressed [review findings](https://github.com/wordpress-mobile/GutenbergKit/pull/691#pullrequestreview-5266554750) where our patch approach may have erroneously flipped valid denials to allows. It also helps us avoid adding yet another patch to maintain. 

This PR's seeding approach remains worthwhile for providing _accurate_ (not optimistic) capability values for cross-origin contexts. However, it appears the resolver parameters this PR utilizes for seeding have already changed in upstream Gutenberg. We'll need to reconcile that if we continue forward with that approach. 

**An alternative worth considering:** we could preload the capabilities alongside the existing preloading, that would negate the need for new bridge APIs. 

I captured the current state of things and alternative details below. 

<details><summary>AI-generated details</summary>
<p>

**The store seeding will stop working after the next `@wordpress` release.** WordPress/gutenberg#82638 (merged to Gutenberg trunk on Sept 17) changes `canUser`'s resolution key from `[ action, resource, id ]` to `[ resource, id ]`, and callers of metadata actions now have to pass the new key themselves. This PR's `finishResolutions( 'canUser', [ [ 'create', { kind: 'postType', name: 'attachment' } ], … ] )` would mark keys nothing reads. The resolver would still send `OPTIONS`, find `Allow` hidden cross-origin, and overwrite the seeded `true` with `false`. The unit tests mock `@wordpress/data`, so they'd stay green.

**Alternative: preload `OPTIONS /wp/v2/media` with its `Allow` header, as core does.** Core's editor preloads that request (`wp-admin/edit-form-blocks.php`), and `rest_preload_api_request()` keeps every response header, `Allow` included. `canUser` then gets real permissions from the preloaded response through `apiFetch`'s preloading middleware, a public path that doesn't depend on store internals. For GBK that would mean:

- `EditorService` preloads `OPTIONS /wp/v2/media` on both platforms, alongside the existing `OPTIONS /wp/v2/settings`.
- Preloaded responses keep `Allow`. Both platforms currently keep only `Accept` and `Link` ([`EditorPreloadList.swift`](https://github.com/wordpress-mobile/GutenbergKit/blob/5d5e5718aa349cfc9c0fb38bc6cb7c74851352f5/ios/Sources/GutenbergKit/Sources/Model/EditorPreloadList.swift#L131), [`EditorURLResponse.kt`](https://github.com/wordpress-mobile/GutenbergKit/blob/5d5e5718aa349cfc9c0fb38bc6cb7c74851352f5/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/http/EditorURLResponse.kt#L50)). Their comment says that matches core, but core keeps every header.

That gives real per-user upload permission on every app surface with no bridge API change, and real entity permissions for the preloaded post as a bonus. Both WordPress apps appear to go through `EditorService` already.

**One thing to verify first:** what the namespaced API returns for an authenticated native `OPTIONS` on the media route. #691 saw `allow: GET`. If that's the real authenticated answer rather than the unauthenticated CORS preflight's, preloading it would hide Upload on namespaced sites.

Meanwhile, #691 covers upload with an `apiFetch` middleware that fills in `Allow: GET, POST` on that request only when the browser hides it. The dev server, E2E and hosts that don't preload keep working either way.

</p>
</details>

PR #463, opened 2026-04-22
fix(js): route showBlockInserter through per-platform dispatch

jkmassel:
## Summary

Fixes a bug where `showBlockInserter` would reach the Android `@JavascriptInterface` as `"[object Object]"`. `bridge.js` now stringifies the payload for Android (single `String` arg) and preserves the structured body for iOS — matching the pattern already used by `openMediaLibrary`, `onNetworkRequest`, and `logException`.

iOS receives the same structured body as before.

This PR is a little awkward because it contains #464 – only bc59eea3e170889b72988886f4c8deadface36a7 requires review. Sorry for the confusion.

## Root cause

The generic `dispatchToBridge` helper at `src/utils/bridge.js` flattens `args` with `Object.values` and passes them positionally to the `@JavascriptInterface`. That works for primitives but `@JavascriptInterface` can't receive arrays or objects — they arrive as the literal string `"[object Object]"`. The block inserter payload is a nested object, so dispatching it through the generic helper silently corrupted it on Android.

## Fix

`showBlockInserter` no longer routes through `dispatchToBridge`. Its Android branch now calls `window.editorDelegate.showBlockInserter(JSON.stringify(payload))`; its iOS branch posts the structured payload as before.

## Test plan

- [x] `make test-js` — new `showBlockInserter` tests cover the no-inserter, Android-only, iOS-only, and both-bridges paths

## Scope note

This PR was split. The Kotlin payload model + parser tests moved to #465 — landed as strict (non-nullable `List<String>` on `BlockPattern`), which assumes the JS side emits `[]` for absent array fields. The matching JSX fix (`?? null` → `?? []` in `native-inserter/index.jsx`) lands in a follow-up along with shape-assertion tests for the producer side.


oguzkocer (approved):
I only reviewed bc59eea3e170889b72988886f4c8deadface36a7 and it looks good.

PR #465, opened 2026-04-23
feat(android): add block inserter payload model

jkmassel:
## Summary

Groundwork for the native block inserter feature on Android: `@Serializable` Kotlin data classes mirroring the JSON shape produced by `BlockInserterBridge` in `src/components/native-inserter/index.jsx`. Parser tests cover full payload, minimal payload, fallback category, explicit nulls on nullable strings, and malformed input.

Extracted from #463 so each PR stays narrow. The original #463 mixed this model with a separate JS bridge fix, and feedback on the stack prompted a split.

## Scope

- **`BlockInserter.kt`** — `BlockInserterPayload`, `BlockInserterSection`, `BlockType`, `BlockPattern`, `BlockPatternCategory`, `SourceRect`.
- **`BlockInserterTest.kt`** — parser tests.

Not referenced from production code yet; the Android `@JavascriptInterface` receiver lands in a later PR in the stack.

The three `List<String>` fields on `BlockPattern` — `blockTypes`, `categories`, `keywords` — are non-nullable with `= emptyList()` defaults. The JS bridge emits an empty array (never `null`) for these fields, so the parser relies on that contract.

## Test plan

- [x] `./gradlew :Gutenberg:testDebugUnitTest detekt :Gutenberg:lintDebug` — passes locally

## Related

- #463 — the original combined PR this was split out of (will be rescoped to JS-only)

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/BlockInserter.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/BlockInserter.kt
@@ -1,0 +1,92 @@
+package org.wordpress.gutenberg.model
+
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+
+/**
+ * Payload delivered to the host app when the web editor requests the native
+ * block inserter. Mirrors the shape produced by
+ * `preprocessBlockTypesForNativeInserter` in `src/utils/blocks.js` and the
+ * `BlockInserterBridge` component at `src/components/native-inserter/index.jsx`.
+ */
+@Serializable
+data class BlockInserterPayload(
+    val sections: List<BlockInserterSection> = emptyList(),
+    val patterns: List<BlockPattern> = emptyList(),
+    val patternCategories: List<BlockPatternCategory> = emptyList(),
+    val sourceRect: SourceRect? = null,
+) {
+    companion object {
+        private val json = Json {
+            ignoreUnknownKeys = true
+            isLenient = true
+        }
+
+        fun fromJson(jsonString: String): BlockInserterPayload =
+            json.decodeFromString(serializer(), jsonString)
+    }
+}
+
+@Serializable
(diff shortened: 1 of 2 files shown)


android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/BlockInserter.kt
```diff
@@ -0,0 +1,92 @@
...
+@Serializable
+data class BlockPattern(
+    val name: String,
+    val title: String,
+    /** HTML with Gutenberg block delimiters. */
+    val content: String,
+    val blockTypes: List<String> = emptyList(),
+    val categories: List<String> = emptyList(),
+    val description: String? = null,
+    val keywords: List<String> = emptyList(),
```
oguzkocer, line 69: > The three `List<String>` fields on `BlockPattern` — `blockTypes`, `categories`, `keywords` — are non-nullable with `= emptyList()` defaults. The JS bridge emits an empty array (never `null`) for these fields, so the parser relies on that contract.

[The JS code](https://github.com/wordpress-mobile/GutenbergKit/blob/4d733c3beb7f7b4737efc1d08d49024ede0eb73d/src/components/native-inserter/index.jsx#L372-L378) seems to set these to `null` which I think might result in parsing errors. It might be worth covering this with a unit test if the JS does produce `null` values.

oguzkocer (commented):
Besides a question about nullability, it looks good to me.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/BlockInserter.kt
jkmassel: That's handled by https://github.com/wordpress-mobile/GutenbergKit/pull/467, which will enforce empty lists instead of `null`

oguzkocer (approved):

PR #461, opened 2026-04-22
feat(android): add native block inserter

jkmassel:
## Summary

Third of three stacked PRs for the native block inserter. Adds the Android-side handler for the `showBlockInserter` JavaScript bridge call set up in #463, gated by the `enableNativeBlockInserter` flag added in #464. `GutenbergView` renders a programmatic `BottomSheetDialog` internally — headless-into-library, so no Compose leaks into the Gutenberg library.

The dialog is a stub — flat list of section headers and tappable block rows. Icons, patterns, search, and richer layout will replace it in follow-up PRs (see **Downstream**).

Rebased directly onto `trunk` so CI picks up the Android instrumented-test job from #471.

## Flow

`bridge.js` `showBlockInserter` → Android `@JavascriptInterface` → `BlockInserterPayload.fromJson` → `BlockInserterDialog` (sections + blocks). Tapping a block fires `window.blockInserter.insertBlock(id)`. Any dismiss — selection or swipe — fires `window.blockInserter.onClose()` so the JS picker state always resets. Bad payloads surface a Toast (from `R.string.gbk_block_inserter_failure`) so a broken tap isn't silent.

## Test plan

- [x] `make test-js` — 135/135
- [x] `./gradlew :Gutenberg:test detekt :Gutenberg:lintDebug :app:compileDebugKotlin` — BUILD SUCCESSFUL
- [x] Manually verified on a Pixel 9 Pro XL: toggle **Enable Native Inserter** → open editor → tap **+** → library sheet opens, tapping a block inserts + dismisses, reopening works on subsequent taps.
- [ ] Reviewer: verify iOS behavior is unchanged. The structured body for iOS is preserved, but I haven't run this branch on iOS myself.

## Related

**Upstream**:
- #463 — `fix(js): route showBlockInserter through per-platform dispatch` (contains the squash-merge of #464's feature flag).
- #465 (merged to trunk) — Kotlin `BlockInserterPayload` model.

**Downstream stacked PRs**:
- #467 — `refactor(js): harden native inserter pattern payload shape` (formatter extraction + shape tests)
- #468 — `feat(android): render branded block icons with contrast-aware tinting`
  - #469 — `feat(android): organize block inserter into cards with search` (stacked on #468)
- #470 — `feat(android): add HTML-to-Bitmap renderer with two-layer cache`

Mirrors the iOS implementation under `ios/Sources/GutenbergKit/Sources/Views/BlockInserter/`.

PR #467, opened 2026-04-23
refactor(js): harden native inserter pattern payload shape

jkmassel:
## Summary

- Extracts inline pattern/pattern-category serialization from `NativeBlockInserterButton` into named helpers in `src/utils/blocks.js`.
- Flips array fields (`blockTypes`, `categories`, `keywords`) from `?? null` to `?? []` to match the non-nullable `List<String>` contract in the Kotlin `BlockPattern` model (#465). Emitting `null` there would throw `SerializationException` on Android.
- Adds a JSON Schema at `schemas/block-inserter-payload.schema.json` that mirrors the Kotlin `BlockInserterPayload` contract, plus unit tests that validate the formatters against it via `ajv`.

## Dependencies

Targets `trunk` directly. Self-contained — all files it touches exist on `trunk` and #465 (the Kotlin `BlockInserterPayload` model the schema mirrors) has merged.

Landing this **early** rather than stacked on #461 is deliberate: the formatter fix and shape-validating tests act as a defensive guardrail, so when the Android native inserter (#461) flips to ready nothing lands on `trunk` already broken. The `?? []` path has no consumer until #461 lands — technically dead code today — but the cost of landing early is low and the payoff (no broken-on-trunk intermediate state) is real.

- **#465** `feat(android): add block inserter payload model` — **merged**. Kotlin `BlockInserterPayload` is now on `trunk`.
- **#461** `feat(android): add native block inserter` — _still draft_. This PR is a guardrail for it but no longer blocks on it.

## Root Cause

`NativeBlockInserterButton` produced `pattern.blockTypes ?? null`, `pattern.categories ?? null`, and `pattern.keywords ?? null`. The Kotlin model declares those fields as non-nullable `List<String>` with `emptyList()` defaults, which means `null` on the wire fails strict deserialization. The bug only surfaces once the bridge is wired up end-to-end, so the shape tests here are the earliest place to catch it.

## Fix

1. **`src/utils/blocks.js`**: Add `formatPatternsForNativeInserter` and `formatPatternCategoriesForNativeInserter`. Array fields default to `[]`; nullable scalars (`description`, `source`, `viewportWidth`) stay `null`.
2. **`src/components/native-inserter/index.jsx`**: Replace ~20 lines of inline mapping with two function calls.
3. **`schemas/block-inserter-payload.schema.json`**: Standalone JSON Schema (draft-07) with `$id`, field-level descriptions pulled from the Kotlin doc comments, and `additionalProperties: false` on every definition. Consumable by cross-language tooling (`quicktype`, `kotlinx-serialization` codegen) alongside the hand-written Kotlin `BlockInserterPayload`.
4. **`src/utils/blocks.test.js`**: Imports the schema and validates payloads via `ajv`. When either side drifts — wrong type, missing field, or unknown field — the tests fail loudly with a JSON-Pointer path instead of failing silently at bridge-deserialization time. Upstream Gutenberg ships no JSON Schema for these shapes (only JSDoc typedefs on `WPEditorInserterItem` / `WPBlockType`), so the schema is hand-written but tracks the Kotlin model as the source of truth.
5. **`package.json`**: Adds `ajv` as a direct devDependency (was transitive).

## Test plan

- [x] `npx vitest run src/utils/blocks.test.js` — 17 tests pass
- [x] `make lint-js-fix` — clean

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -80,6 +80,7 @@
 				"@wordpress/env": "^11.2.0",
 				"@wordpress/eslint-plugin": "^24.3.0",
 				"@wordpress/prettier-config": "^4.43.0",
+				"ajv": "^8.18.0",
 				"eslint": "^8.57",
 				"eslint-import-resolver-typescript": "^4.4.4",
 				"eslint-plugin-react-refresh": "^0.4.26",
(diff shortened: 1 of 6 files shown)


jkmassel:
e2e test failures will be resolved by #473

oguzkocer (approved):

PR #470, opened 2026-04-23
feat(android): add HTML-to-Bitmap renderer

jkmassel:
## Summary

Groundwork for rendering block patterns as thumbnails in the native Android inserter. Introduces `HtmlToBitmapRenderer`: an off-screen `WebView` that loads an HTML string and snapshots the laid-out content to a `Bitmap` via `Canvas.draw()`. Lays out at the viewport width in device pixels, awaits `onPageFinished`, polls `document.images.every(i => i.complete)` so async-loaded images aren't missing from the snapshot, queries `document.documentElement.scrollHeight`, re-lays out to content height, then draws into a scaled bitmap. Never upscales.

Mirrors the iOS `HTMLWebViewRenderer` MVP under `ios/Sources/GutenbergKit/Sources/Views/HTMLPreview/`. No consumer yet — the `BlockInserterDialog` wiring that turns patterns into previews lands separately.

## Design notes

- **Single-use WebView.** Each call creates and destroys its own `WebView` in a `finally`. Pooling is a deliberate follow-up — iOS pools because `WKWebView` allocation is expensive, but we want to see the simpler model exercised before adding a pool.
- **Software layer + post-layout settle.** `setLayerType(LAYER_TYPE_SOFTWARE)` forces the WebView to raster onto a software canvas rather than a GPU texture (which `WebView.draw()` can't reach off-screen). After re-layout we `delay(POST_LAYOUT_SETTLE_MS)` because there's no parent-window invalidation cycle to drive a repaint.
- **No caching here.** Two-layer (memory + disk) caching, WebView pooling, and consumer wiring all land in follow-ups so this PR stays focused on the renderer primitive.

## Known gaps

- No caching. Renders are recomputed every call. (Follow-up.)
- No WebView pool. (Follow-up.)
- No unit tests. Robolectric's shadow `WebView` doesn't actually render. Coverage comes from the `androidTest/` instrumented tests, which run in CI via the `:android: Test Android Library Instrumented` step (#472).

## Test plan

- [x] `./gradlew :Gutenberg:detekt :Gutenberg:lintDebug :Gutenberg:compileDebugKotlin :Gutenberg:test` — `BUILD SUCCESSFUL`
- [x] Instrumented tests pass locally on a Pixel 9 Pro XL (`./gradlew :Gutenberg:connectedAndroidTest`)
- [ ] CI's `:android: Test Android Library Instrumented` step passes on this PR

## Follow-ups

- Add a caching layer (`CachingHtmlToBitmapRenderer` + `BlobCache`/`FileBlobCache`) on top of the renderer.
- Wire pattern thumbnails into `BlockInserterDialog` so the renderer is exercised end-to-end.

### DIFF
--- a/android/Gutenberg/src/androidTest/java/org/wordpress/gutenberg/preview/HtmlToBitmapRendererInstrumentedTest.kt
+++ b/android/Gutenberg/src/androidTest/java/org/wordpress/gutenberg/preview/HtmlToBitmapRendererInstrumentedTest.kt
@@ -1,0 +1,125 @@
+package org.wordpress.gutenberg.preview
+
+import android.graphics.Bitmap
+import android.graphics.Color
+import android.util.Log
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import kotlinx.coroutines.runBlocking
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.io.File
+
+/**
+ * Instrumented tests that exercise [HtmlToBitmapRenderer] against a real Android
+ * [android.webkit.WebView]. Runs in CI via the `:android: Test Android Library
+ * Instrumented` step (`make test-android-library-e2e`); locally:
+ *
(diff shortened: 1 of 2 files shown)


adalpari (approved):
LGTM! 🚢 it!

PR #471, opened 2026-04-23
ci: run Android instrumented tests on the mac-metal queue

jkmassel:
## What?

Adds a new Buildkite step that runs `:app:connectedDebugAndroidTest` on the `mac-metal` queue, booting an ephemeral AVD via the `start-android-emulator` helper.

## Why?

The Android instrumented tests in `android/app/src/androidTest/` haven't had a CI job — they've only run locally. The new `mac-metal` agents can boot real Android emulators (something the virtualized `mac` queue can't do), so we can finally wire these tests into the pipeline.
## How?

- Downloads the `dist.tar.gz` artifact produced by `Build React App` so the editor assets are available.
- Calls `start-android-emulator` with `system-images;android-34;google_apis;arm64-v8a` (pre-installed on the agent, matches our `targetSdk = 34`) and a `pixel_5` device profile.
- Runs `make test-android-e2e`. `ENSURE_ANDROID_DEVICE` detects the already-booted emulator via `adb` and skips its fallback boot path, so the existing Makefile target works unchanged.
- AVD cleanup is handled by the agent's `pre-exit` hook — no extra teardown needed here.

## Test plan

- [ ] CI runs the new `:android: Test Android E2E` step on this PR against the `mac-metal` queue and the step succeeds.
- [ ] `EditorInteractionTest` tests (`testEditorWebViewBecomesVisible`, `testUndoRedoAfterTyping`) pass on CI.
- [ ] Other pipeline steps (iOS E2E, Test Android Library, etc.) are unaffected.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -89,3 +89,7 @@ steps:
       agents:
           queue: mac-metal
       plugins: *plugins
+      artifact_paths:
+          - 'android/app/build/reports/androidTests/connected/**/*'
+          - 'android/app/build/outputs/androidTest-results/connected/**/*'
+          - 'android/app/build/outputs/connected_android_test_additional_output/**/*'
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -74,3 +74,18 @@ steps:
           tar -xzf dist.tar.gz
           make test-ios-e2e
       plugins: *plugins
+
+    - label: ':android: Test Android E2E'
+      depends_on: build-react
+      command: |
+          buildkite-agent artifact download dist.tar.gz .
+          tar -xzf dist.tar.gz
+          echo "--- :robot_face: Starting Android emulator"
+          start-android-emulator \
+              --system-image "system-images;android-34;google_apis;arm64-v8a" \
+              --device "pixel_5"
+          echo "--- :android: Running Android instrumented tests"
+          make test-android-e2e
+      agents:
+          queue: mac-metal
+      plugins: *plugins
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -51,6 +51,14 @@ android {
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
 
+        // Buildkite Test Engine: upload per-test results when a token is present in the env.
+        // Absent locally → no listener is registered and the collector is fully inert.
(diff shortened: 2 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -74,3 +74,18 @@ steps:
           tar -xzf dist.tar.gz
           make test-ios-e2e
       plugins: *plugins
+
+    - label: ':android: Test Android E2E'
```
iangmaia, line 78: Not a blocker as this is just the initial setup to get it going, but maybe for the long run it might be worth it to add `artifact_paths` to debug test results (stuff from logcat, test reports, etc).

iangmaia (approved):
👍 :shipit: 

Image

android/app/build.gradle.kts
```diff
@@ -51,6 +51,14 @@ android {
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
 
+        // Buildkite Test Engine: upload per-test results when a token is present in the env.
+        // Absent locally → no listener is registered and the collector is fully inert.
+        System.getenv("BUILDKITE_ANALYTICS_TOKEN_ANDROID_E2E")?.takeIf { it.isNotBlank() }?.let { token ->
+            testInstrumentationRunnerArguments["listener"] =
+                "com.buildkite.test.collector.android.InstrumentedTestCollector"
+            testInstrumentationRunnerArguments["BUILDKITE_ANALYTICS_TOKEN"] = token
```
iangmaia, line 61: By the way, it's probably also interesting to set these for Buildkite's test collector https://github.com/buildkite/test-collector-android/blob/main/CI_CONFIGURATION.md#buildkite

jkmassel: Nice – addressed by [9b4809f](https://github.com/wordpress-mobile/GutenbergKit/pull/471/commits/9b4809fd4bfff87af0f23371861177322136b836)

.buildkite/pipeline.yml
jkmassel: Sure – added in [d9329d2](https://github.com/wordpress-mobile/GutenbergKit/pull/471/commits/d9329d24d757fd535e8b700677b503bfaf33bf6a), thanks!

android/app/build.gradle.kts
iangmaia: 🎉

PR #473, opened 2026-04-24
fix(android): stabilize editor E2E tests after native inserter regression

jkmassel:
## Summary

`testUndoRedoAfterTyping` was failing deterministically on CI (build [2043](https://buildkite.com/automattic/gutenbergkit/builds/2043)) with `Atom evaluation returned null!`. Two independent issues, both fixed here.

### 1. `navigateToEditor` races editor startup

The existing `waitForWebViewElement(TITLE_SELECTOR)` call never actually polled anything — `withElement(findElement(...))` is lazy and requires a terminal op (`.perform()` / `.check()` / `script()`) to evaluate. The fall-through meant tests proceeded before `window.editor` was callable and before the React tree had mounted.

Confirmed locally: reproduced the failure, then hardened the gate, then re-ran green.

### 2. Demo defaulted to the native inserter

#464 flipped `SitePreparationUiState.enableNativeInserter` to `true`, which swaps the Gutenberg default `Inserter` (`aria-label="Add block"`) for `NativeInserter` (`title="Add block"`, different className). #463's `showBlockInserter` dispatches to the native bridge, but the Android `@JavascriptInterface` receiver hasn't landed yet — so tapping does nothing and the toolbar test failed. Reverting the demo default to `false` until the native handler ships.

## Changes

- **`EditorTestHelpers.kt`**: new `waitForEditorReady` gate that polls for both `window.editor.setContent` and `[aria-label="Editor toolbar"]` via a `script()` (terminal) atom. `navigateToEditor` now calls this instead of the broken element-wait.
- **`EditorTestHelpers.kt`**: rewrote `waitForWebViewElement` to use the same `script()`-based probe so it actually evaluates. Also added a defensive wait at the top of `insertBlock`.
- **`SitePreparationViewModel.kt`**: `enableNativeInserter` default back to `false`.

## Test plan

- [x] `EditorInteractionTest#testUndoRedoAfterTyping` passes locally (`gbkit_api34` AVD, API 34)
- [x] Full `EditorInteractionTest` class (2 tests) passes locally
- [x] CI goes green

## CI stability runs

Four independent runs of this commit (`e58ef3e`) all show `:android: Test Android E2E` green — the test is no longer flaky.

| Build | Android E2E | Notes |
| --- | --- | --- |
| [2045](https://buildkite.com/automattic/gutenbergkit/builds/2045) | ✅ | Initial PR build — all 11 jobs passed |
| [2046](https://buildkite.com/automattic/gutenbergkit/builds/2046) | ✅ | Rebuild 1/3 |
| [2047](https://buildkite.com/automattic/gutenbergkit/builds/2047) | ✅ | Rebuild 2/3 |
| [2048](https://buildkite.com/automattic/gutenbergkit/builds/2048) | ✅ | Rebuild 3/3 |

The rebuilds each show `:android: Publish Android Library` red — this is the pipeline's expected idempotency check rejecting the already-published artifact for commit `e58ef3e` (`'org.wordpress.gutenbergkit.android:473-e58ef3e...' is already published to S3!`). Unrelated to this PR and only fires on same-commit re-runs.

### DIFF
--- a/android/app/src/androidTest/java/com/example/gutenbergkit/EditorTestHelpers.kt
+++ b/android/app/src/androidTest/java/com/example/gutenbergkit/EditorTestHelpers.kt
@@ -48,8 +48,8 @@ object EditorTestHelpers {
 
     /**
      * Navigates from the main list through the configuration screen
-     * and into the full-screen editor. Waits for the "Add title" element
-     * in the WebView to confirm the editor has loaded.
+     * and into the full-screen editor, then blocks until the Gutenberg
+     * JS has finished booting and `window.editor` is callable.
      */
     fun navigateToEditor(
         rule: EditorTestRule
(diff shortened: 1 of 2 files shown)


oguzkocer (approved):

jkmassel:
admin-merging to get around the android publishing failure (the library was already published for this PR previously)

PR #476, opened 2026-04-25
test(ios): exercise the Swift package on macOS via `swift test`

jkmassel:
## Summary

- The only Swift test job in CI runs `xcodebuild -sdk iphonesimulator`, where `canImport(UIKit)` is always true. iOS-only `#if canImport(UIKit)` gates are never exercised against macOS, so it's possible to break the host-platform build without CI noticing — and `LockdownModeMonitorTests` was in fact already broken on macOS (it imported UIKit unconditionally).
- Add a "Library Tests" Buildkite step that runs `swift test` against the host platform (macOS, where `canImport(UIKit)` is false), so this class of regression gets caught going forward.
- Refactor `LockdownModeMonitor` so the detection state machine is host-buildable and only the UIKit-dependent presentation methods live in a gated extension. Consolidate test gating in the same spirit. The state machine now runs on both the iOS Simulator and macOS — 13 more tests via `swift test` than before.

## Why

The production `LockdownModeMonitor.swift` was gated on `#if canImport(UIKit)` because it depends on UIKit + WKWebView APIs. The companion test file was *not* gated, so it imported UIKit unconditionally and referenced types that don't exist on macOS. The breakage was invisible because:

- `.buildkite/pipeline.yml`'s sole Swift test job calls `make test-swift-package` → `xcodebuild ... -sdk iphonesimulator` (UIKit available).
- `Package.swift` advertises `.macOS(.v14)` support, but no CI job builds against macOS.

So the pipeline was happy while `swift test` was broken.

### CI

- `.buildkite/pipeline.yml`: rename `:swift: Test Swift Package` → `:swift: iOS Simulator Tests`, with `key: swift-test-swift-package` pinned so the GitHub commit-status name stays the same for branch protection. Add `:swift: Library Tests` running `make test-swift-library`.
- `Makefile`: add `test-swift-library` target running `swift test`. Tighten the existing `test-swift-package` doc comment to note it's the simulator one.

### Production split

- `ios/Sources/GutenbergKit/Sources/Services/LockdownModeMonitor.swift`: split so the state machine — `LockdownModeDetectable` protocol, `WKWebView: LockdownModeDetectable` extension, `detectLockdownMode`, `resetForForegroundCheck` — is host-buildable. Only the UIKit-dependent surface (`setup(presentingViewController:)`, `handleWillEnterForeground`, `presentSheetIfNeeded`, `dismissSheetIfPresented`, plus the `presentingViewController` stored property) is gated behind `#if canImport(UIKit)` and lives in a separate extension at the bottom.
- Class doc gains a one-sentence note explaining the split, so the next reader doesn't have to spelunk to find the gated extension.
- `deinit` calls `removeObserver(self)` unconditionally; added a comment noting that's a safe no-op when `setup()` never ran (macOS or pre-`setup()` iOS).

### Tests

- `ios/Tests/GutenbergKitTests/Services/LockdownModeMonitorTests.swift`: only the import of `UIKit` and the six tests that actually need it (`setup` / `presentSheetIfNeeded` / `dismissSheetIfPresented`) are gated. The mock-based tests and state-transition tests run on both platforms. The six gated tests are consolidated into a single `// MARK: - UIKit-Only Tests` block at the bottom of the suite under one `#if canImport(UIKit)`, instead of being scattered through six interleaved `#if`/`#endif` pairs.
- The "State Transition Tests" section comment was rewritten — its previous "would deadlock in CI" rationale was about the old shape where presentation lived as siblings; now it just notes that the state machine is host-buildable and presentation tests live in the section below.

## Test plan

- [x] `swift test` (macOS host) — 861 tests across 45 suites pass.
- [x] `make test-swift-package` (iPhone Simulator) — passes, including the consolidated `UIKit-Only Tests` block.
- [x] `make test-swift-library` (new target, macOS host) — passes.
- [x] Confirm Buildkite shows both `:swift: iOS Simulator Tests` and `:swift: Library Tests` and both pass.
- [ ] Confirm the existing `buildkite/gutenbergkit/swift-test-swift-package` commit-status name still appears (preserved via `key:`), so branch protection stays green.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -59,12 +59,22 @@ steps:
           queue: android
       plugins: *plugins
 
-    - label: ':swift: Test Swift Package'
+    - label: ':swift: iOS Simulator Tests'
+      key: swift-test-swift-package
       depends_on: build-react
       command: |
           buildkite-agent artifact download dist.tar.gz .
           tar -xzf dist.tar.gz
           make test-swift-package
+      plugins: *plugins
+
+    - label: ':swift: Library Tests'
+      key: swift-test-library
+      depends_on: build-react
+      command: |
+          buildkite-agent artifact download dist.tar.gz .
+          tar -xzf dist.tar.gz
+          make test-swift-library
       plugins: *plugins
 
     - label: ':ios: Test iOS E2E'
--- a/Makefile
+++ b/Makefile
@@ -231,8 +231,12 @@ test-js-watch: npm-dependencies ## Run JavaScript tests in watch mode
 	npm run test:unit:watch
 
 .PHONY: test-swift-package
-test-swift-package: build ## Run Swift package tests
+test-swift-package: build ## Run Swift package tests in the iOS Simulator
 	$(call XCODEBUILD_CMD, test, GutenbergKit-Package)
+
+.PHONY: test-swift-library
+test-swift-library: build ## Run Swift package tests against the host platform via `swift test`
+	swift test
 
 .PHONY: test-ios-e2e
 test-ios-e2e: ## Run iOS E2E tests against the production build
--- a/ios/Sources/GutenbergKit/Sources/Services/LockdownModeMonitor.swift
+++ b/ios/Sources/GutenbergKit/Sources/Services/LockdownModeMonitor.swift
@@ -5,6 +5,7 @@ import OSLog
 
 #if canImport(UIKit)
 import UIKit
+#endif
 
 /// Protocol for objects that can be checked for Lockdown Mode status.
 ///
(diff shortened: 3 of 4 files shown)


oguzkocer (approved):
>  [ ] Confirm the existing buildkite/gutenbergkit/swift-test-swift-package commit-status name still appears (preserved via key:), so branch protection stays green.

If I am understanding this test instruction correctly, I don't think this is currently the case. The `buildkite/gutenbergkit/swift-test-swift-package` job shows up as `Required` and `Waiting for status`, so the old branch protection rule will no longer work. However, we'd usually update the GitHub branch protection settings for these cases instead of trying to preserve the old key, so I am still going to approve.

PR #477, opened 2026-04-27
feat(ios): introduce SQLiteKVCache — a persistent KV blob cache

jkmassel:
Introduces `SQLiteKVCache`, a small self-contained SQLite-backed key-value blob cache. Standalone — not yet wired into any consumer. Migration of `EditorURLCache` (and follow-up of `HTMLPreviewManager`) onto this cache is a separate PR.

## What it is

```swift
let cache = SQLiteKVCache(handle: "MyCache", diskCapacity: 100 * 1024 * 1024)
try cache.put(key: "foo", storageDate: .now, metadata: Data(), value: Data("hello"))
let entry = try cache.get(key: "foo")    // (storageDate, metadata, value)?
try cache.delete(key: "foo")
try cache.clear()
```

Each entry is `(key, storageDate, metadata, value)`. Values and metadata are stored as raw bytes — the caller owns any serialization on top.

## API

```swift
init(handle: StaticString, directory: URL = URL.cachesDirectory, diskCapacity: Int)
func get(key: String) throws -> Entry?
func put(key: String, storageDate: Date, metadata: Data, value: Data) throws
func delete(key: String) throws
func clear() throws

enum Error: Swift.Error, CustomStringConvertible, LocalizedError {
    case databaseUnavailable
    case readFailed(sqliteCode: Int32)
    case writeFailed(sqliteCode: Int32)
}
```

`init` is **non-throwing**. The SQLite connection is opened lazily on the first operation and the result (success or cached failure) is reused for the lifetime of the cache — callers using `try?` get a uniform "cache unavailable" fall-through without a separate error path at the construction site.

`get`, `put`, `delete`, and `clear` all throw. `get` returns `nil` only for a genuine miss (`SQLITE_DONE`); any other non-row step result or a prepare/bind failure throws `Error.readFailed`. Callers wanting "treat read failures as misses" semantics wrap `get` in `try?`. Writes throw on IO failure (disk full, blob oversize, corruption); the SQLite result code is preserved in the associated value for callers that want to pattern-match.

The `Error` enum renders through `sqlite3_errstr`, so `print(error)` and `error.localizedDescription` both produce sentences like `"SQLite write failed: disk I/O error (code 10)"` instead of `"writeFailed(sqliteCode: 10)"`.

## Notable behavior

- **Cache, not store.** `synchronous = NORMAL` trades the last few committed writes on power loss; schema-version mismatch drops the table; eviction is destructive and oldest-first by `storage_date`. `Cache` in the type name is load-bearing — primary-store data does not belong here.
- **Single autocommit statements, no Swift-level lock.** Every operation is one SQL statement: `put` is `INSERT … ON CONFLICT DO UPDATE` with the eviction sweep happening inside `AFTER INSERT`/`AFTER UPDATE` triggers; `get`/`delete`/`clear` are single `SELECT`/`DELETE` statements. `SQLITE_OPEN_FULLMUTEX` serializes the C-level API calls and SQLite's per-statement atomicity (including trigger bodies) handles the rest — there's no Swift-level lock because there's no multi-statement transaction to fence. Single-thread `put` then `get` always observes the put.
- **Oldest-first eviction via triggers.** Two triggers (`AFTER INSERT` and `AFTER UPDATE`, because SQLite fires UPDATE triggers — not INSERT — on the upsert's `ON CONFLICT DO UPDATE` branch) with identical bodies: a window-function `DELETE` that removes oldest-by-`storage_date` rows until total stored size is at or below `diskCapacity`. Each is guarded by a `WHEN (SELECT SUM…) > diskCapacity` aggregate so puts that don't push the table past cap pay only the cheap aggregate scan, not the full window-function query. An entry whose own size exceeds `diskCapacity` is silently dropped (the trigger would insert it then evict it immediately, so the round trip is skipped). Not strict LRU: `get` doesn't bump recency. Callers wanting LRU semantics re-`put` on access.
- **Triggers recreated on every open.** `installEvictionTriggers(on:diskCapacity:)` runs unconditionally during the lazy open, with the current instance's `diskCapacity` baked into the trigger SQL. `diskCapacity == 0` drops the triggers and doesn't recreate them, disabling eviction entirely.
- **WAL is required, not best-effort.** `journal_mode = WAL` is verified by reading the resulting mode from the PRAGMA's row (`sqlite3_exec` can't tell "set" from "stayed at default"), and open fails if anything else comes back. WAL is a correctness dependency for `synchronous = NORMAL` — NORMAL is documented safe under WAL but only "probably safe" under rollback-journal — so the unsafe pairing is impossible by construction. Acceptable durability tradeoff for a cache: losing the last few writes on power loss costs an extra refetch, not data loss against a primary store.
- **Disk reclamation is one-shot at open.** `diskCapacity` caps content size (sum of `length(value) + length(metadata)` across rows), not on-disk file size — eviction and DELETE leave free pages on the freelist, which `auto_vacuum = NONE` (the default) doesn't reclaim. Rather than `auto_vacuum = FULL` (which writes freelist pages on every transaction), the open path runs a single `VACUUM` if the freelist exceeds 25% of total pages. Bounds SSD churn to once per process; tight launches pay only the freelist pragma reads.
- **Permissive blob input.** Any `String` key is SHA-256-hashed (raw 32-byte digest) before binding, so length, encoding, embedded null bytes, and SQL-escaping concerns are all moot. Any `Data` value/metadata uses `sqlite3_bind_blob64` so the byte-count cast can't trap on hypothetical >2 GiB inputs.
- **Tight handle input.** `handle` is `StaticString`, so callers must pass a string literal — runtime-constructed handles are rejected at compile time. The handle is lowercased before becoming the filename (so `"EditorURLCache"` and `"editorurlcache"` share a database file, matching case-insensitive filesystem defaults). A debug-build `assert` catches handles that aren't `[a-zA-Z0-9._-]`; release builds use the lowercased handle as-given.
- **One instance per backing file.** Two `SQLiteKVCache` instances against the same `<directory>/<handle>.sqlite` is **undefined behavior** — the eviction triggers are recreated unconditionally on open, so two instances with different caps would clobber each other's triggers. Documented but not enforced at runtime.
- **Schema versioning.** `PRAGMA user_version` is checked on open; anything other than the constant baked into code triggers a drop-and-recreate. That includes the fresh-DB case (`user_version` defaults to 0 on a new SQLite file). Currently at v1 with no shipped history; future schema changes that hit user devices bump the constant and accept the cache reset as the migration model.
- **Bind failures throw at the bind site.** `sqlite3_bind_blob64` / `sqlite3_bind_double` are wrapped in throwing helpers that surface the bind result code (`SQLITE_TOOBIG`, `SQLITE_RANGE`, `SQLITE_NOMEM`) under the operation's own `Error` category — no laundering into `SQLITE_MISUSE` at step time.

## Test coverage

`SQLiteKVCacheTests` — focused tests at the KV-cache API:

- **Round-trip and basic CRUD:** `put`/`get` round-trip, missing key, overwrite, clear, distinct keys, empty value/metadata round-trip, delete one, delete missing key
- **Persistence:** entries persist across instances (same directory + handle), handle independence in the same directory
- **Eviction (oldest-first by `storage_date`):** under capacity nothing evicts, oldest evicted over capacity, eviction tie-breaker deterministic on equal `storage_date`, `diskCapacity: 0` disables, oversized-entry dropped on store
- **Adversarial keys:** SQL-shaped (`'; DROP TABLE entries; --`), unicode/emoji, embedded null bytes — all round-trip independently because SHA-256 distinguishes them
- **Handle validation:** `isValidHandle` accepts ordinary identifiers, rejects empty / `.` / `..` / unsafe characters / uppercase / non-ASCII; `init` lowercases mixed-case handles so they share a database file
- **Error descriptions:** `writeFailed`, `readFailed`, and `databaseUnavailable` all render through `sqlite3_errstr`; `localizedDescription` matches `description`
- **Concurrency:** 200-task concurrent `put`/`get` against 10 overlapping keys (no eviction); a sibling test runs the same shape with a small `diskCapacity` so eviction fires on most puts, exercising the eviction-trigger path under contention
- **Schema setup:** fresh DB ends up at the current `schemaVersion`; legacy schema-mismatch recovery (uses `Int32.max` as a sentinel `user_version` so the test stays collision-proof if `schemaVersion` is bumped)
- **WAL:** opens with `journal_mode = wal`, verified via a separate connection
- **VACUUM on open:** reclaims the freelist after heavy churn (populate, delete most rows, reopen, assert freelist is zero); skipped when freelist is below threshold (write-only workload, reopen, assert page count unchanged)

## What's not in this PR

- No changes to `EditorURLCache` — it's still `URLCache`-backed, exactly as on `trunk`
- No changes to `RESTAPIRepository` or `HTMLPreviewManager` (the prospective consumers)

The follow-up PR migrates `EditorURLCache` onto `SQLiteKVCache`. A subsequent one does the same for `HTMLPreviewManager`.

## Test plan

- [x] `make test-swift-package` — all suites pass
- [x] `EditorURLCacheTests` (unchanged from `trunk`) still passes — the new file doesn't introduce any unintended coupling
- [ ] Confirm Buildkite swift-test passes

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVCache.swift
+++ b/ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVCache.swift
@@ -1,0 +1,674 @@
+import CryptoKit
+import Foundation
+import OSLog
+import SQLite3
+
+/// A SQLite-backed persistent key-value cache for opaque blobs.
+///
+/// **Cache only — not a primary store.** Although entries persist across
+/// process restarts, the on-disk contract is intentionally lossy: a schema
+/// version mismatch drops the table on open; eviction discards oldest-first
+/// when total content exceeds `diskCapacity`; and `synchronous = NORMAL`
+/// trades the last few committed writes on power loss for substantially
+/// faster small writes. Use this for data that's cheap to refetch from a
+/// canonical source — never as the only copy.
+///
+/// Each entry is `(key, storageDate, metadata, value)`. Values and metadata are
+/// stored as raw bytes; the caller owns any serialization. After every `put`,
+/// oldest entries (by storage date) are evicted until total stored size is at
+/// or below `diskCapacity`.
+///
+/// **Keys.** The cache accepts any `String` as a key and internally hashes it
+/// (SHA-256, raw 32-byte digest) before binding into SQLite. This means callers
+/// don't have to think about key length, SQL escaping, embedded null bytes, or
+/// character encoding — anything the caller can hand us round-trips. The trade
+/// is debuggability: a `sqlite3 ... SELECT hex(key) FROM entries` shows hash
+/// digests as hex, not the original input strings.
+///
+/// **Concurrency.** Thread-safe. Every operation is a single autocommit SQL
+/// statement: `put` is `INSERT … ON CONFLICT DO UPDATE` with the eviction
+/// sweep happening inside `AFTER INSERT` / `AFTER UPDATE` triggers; `get`,
+/// `delete`, and `clear` are also single statements. `SQLITE_OPEN_FULLMUTEX`
+/// serializes the C-level API calls on the connection, and SQLite's
+/// per-statement atomicity (including trigger bodies) handles the rest —
+/// there's no Swift-level lock because there's no multi-statement transaction
+/// to fence. Single-thread `put` then `get` always observes the put.
+///
+/// **One instance per backing file.** Opening two `SQLiteKVCache` instances
+/// against the same `<directory>/<handle>.sqlite` (whether in the same process
+/// or across processes — main app vs. share extension) is **undefined
+/// behavior**. The eviction triggers are recreated unconditionally on every
+/// open with the current instance's `diskCapacity` baked into them, so two
+/// instances with different caps would clobber each other's triggers; in the
+/// best case you get the wrong cap, in the worst case `SQLITE_BUSY` while the
+/// recreations race. Each backing file must have exactly one owning
+/// `SQLiteKVCache` for the lifetime of the process. Not currently enforced at
+/// runtime — this is a usage contract.
+///
+/// **Schema migrations.** A `schemaVersion` constant baked into the build is
+/// compared against `PRAGMA user_version` on open; mismatches drop and recreate
+/// the table. The check trusts the version: if a buggy past build wrote a
+/// matching `user_version` but the wrong column shape, this won't detect it.
+///
+/// **Disk reclamation.** `diskCapacity` caps the *content* size — the sum of
+/// `length(value) + length(metadata)` across rows. The on-disk file can drift
+/// past that cap because evicted/deleted rows leave free pages on the SQLite
+/// freelist, which `auto_vacuum = NONE` (the default) doesn't reclaim. Rather
+/// than enable `auto_vacuum = FULL` (which would rewrite freelist pages on
+/// every transaction), the open path runs a single `VACUUM` if the freelist
+/// has grown past `vacuumFreelistThreshold` of total pages — bounding the
+/// SSD-churn cost to once per process and skipping launches where there's
+/// nothing meaningful to reclaim.
+final class SQLiteKVCache: @unchecked Sendable {
+
+    enum Error: Swift.Error {
+        /// The cache couldn't be opened or set up. Typically the caches
+        /// directory is sandboxed-out, the disk is full, an existing file
+        /// at the path is corrupt, or a setup pragma / DDL statement
+        /// failed during open. The first operation on the cache throws
+        /// this and so does every subsequent operation (the failure is
+        /// cached) — callers using `try?` get a uniform "cache
+        /// unavailable" fall-through. Silently degrading to a
+        /// process-lifetime in-memory store would violate the persistence
+        /// contract `put` implies. The originating SQLite error code is
+        /// not preserved on this case (setup-time failures collapse to
+        /// the same surface) but is logged at the failure site.
+        case databaseUnavailable
+        /// A read (`get`) failed inside SQLite. The associated value is the
+        /// SQLite error code. Cache callers that want the prior "treat read
+        /// failures as misses" behavior can wrap the call in `try?`.
+        case readFailed(sqliteCode: Int32)
+        /// A write (`put`, `delete`, or `clear`) failed inside SQLite. The
+        /// associated value is the SQLite error code — common ones include
+        /// `SQLITE_FULL` (disk full, code 13), `SQLITE_TOOBIG` (blob over
+        /// `SQLITE_MAX_LENGTH`, code 18), and `SQLITE_CORRUPT` (code 11).
+        case writeFailed(sqliteCode: Int32)
+    }
+
+    // MARK: - Debugging
+    //
+    // Backed by a single SQLite database at `<directory>/<handle>.sqlite`.
+    // Useful queries from a shell:
+    //
+    //     # List entries by recency, with size and date
+    //     sqlite3 Store.sqlite \
+    //         "SELECT hex(key), length(value), datetime(storage_date + 978307200, 'unixepoch') \
+    //          FROM entries ORDER BY storage_date DESC"
+    //
+    //     # Export a specific value to a file (key is the SHA-256 digest as hex)
+    //     sqlite3 Store.sqlite \
+    //         "SELECT writefile('/tmp/value.bin', value) FROM entries \
+    //          WHERE key = x'<digest-hex>'"
+    //
+    // `storage_date` is `Date.timeIntervalSinceReferenceDate` (seconds since
+    // 2001-01-01); +978307200 shifts to unix epoch for `datetime(..., 'unixepoch')`.
+
+    struct Entry {
+        let storageDate: Date
+        let metadata: Data
+        let value: Data
+    }
+
+    private let diskCapacity: Int
+    private let directory: URL
+    private let filename: String
+
+    /// Cached open result. `nil` means "open hasn't been attempted yet"; once
+    /// set, either holds the live connection or the error we'll keep handing
+    /// back for the lifetime of the cache. Reading and writing both happen
+    /// under `openLock` — the cached value is read/written via the same code
+    /// path, and the open work runs at most once per cache.
+    private var dbResult: Result<OpaquePointer, Swift.Error>?
+    private let openLock = NSLock()
+
+    /// Schema version baked into this build. The on-disk `PRAGMA user_version`
+    /// is checked on open; anything other than this exact value triggers a
+    /// drop-and-recreate. That includes the fresh-database case (`user_version`
+    /// defaults to 0 on a brand-new SQLite file), so the first open of a new
+    /// database also runs the drop path — harmless because there's nothing to
+    /// drop. Bump only on changes that hit user devices — pre-ship iteration
+    /// shouldn't burn version numbers. Never decrement (a downgrade would
+    /// silently wipe users' caches a second time on the next upgrade).
+    private static let schemaVersion: Int32 = 1
+
+    private static let logger = Logger(subsystem: "GutenbergKit", category: "sqlite-kv-cache")
+
+    /// SQLite C API: signals that bound data should be copied. Reinvented here
+    /// because the `SQLITE_TRANSIENT` C macro doesn't import to Swift.
+    private static let SQLITE_TRANSIENT = unsafeBitCast(
+        OpaquePointer(bitPattern: -1), to: sqlite3_destructor_type.self
+    )
+
+    /// Creates a new cache. **Non-throwing**: the SQLite connection is opened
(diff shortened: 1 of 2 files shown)


ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVCache.swift
```diff
@@ -0,0 +1,485 @@
...
+    ///   - diskCapacity: Soft cap (in bytes) on the combined size of stored values
+    ///     and metadata. After every `put`, oldest entries (by storage date) are
+    ///     evicted until total size is at or below this cap. A `put` for an entry
+    ///     whose own size exceeds the cap is silently dropped — choose a cap
+    ///     comfortably above the largest expected entry. Pass `0` to disable
+    ///     eviction entirely.
+    init(
+        handle: StaticString,
+        directory: URL = URL.cachesDirectory,
+        diskCapacity: Int
```
crazytonyli, line 168: Using https://developer.apple.com/documentation/Foundation/Measurement, so that the unit is part of the type?

ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVCache.swift
```diff
@@ -0,0 +1,485 @@
...
+    var description: String {
+        switch self {
+        case .databaseUnavailable:
+            return "SQLite database is unavailable (file open and :memory: fallback both failed)"
+        case .writeFailed(let sqliteCode):
+            return "SQLite write failed: \(SQLiteKVStore.describe(sqliteCode))"
+        }
+    }
+
+    var errorDescription: String? { description }
```
crazytonyli, line 673: The `description` is not localized.

ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVStore.swift
```diff
@@ -0,0 +1,485 @@
...
+/// `<directory>/<handle>.sqlite` (e.g. main app vs. share extension) gets two
+/// separate connections that fall back to SQLite's file-level locking, and the
+/// synchronous reads-after-writes guarantee no longer holds across them. Treat
+/// each file as having a single owning instance.
+///
+/// **Schema migrations.** A `schemaVersion` constant baked into the build is
+/// compared against `PRAGMA user_version` on open; mismatches drop and recreate
+/// the table. The check trusts the version: if a buggy past build wrote a
+/// matching `user_version` but the wrong column shape, this won't detect it.
+final class SQLiteKVStore: @unchecked Sendable {
```
crazytonyli: I think we should explicitly add "cache" to the type name, to avoid it being misused.

ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVStore.swift
```diff
@@ -0,0 +1,485 @@
...
+            )
+        case SQLITE_DONE:
+            return nil
+        default:
+            Self.logger.error("get failed at sqlite3_step: \(Self.describe(stepResult))")
+            return nil
+        }
+    }
+
+    /// Inserts or overwrites the entry at `key`. Runs the LRU eviction sweep
```
crazytonyli: The `evictPastCapacity` deletes oldest entries, which is not "LRU eviction", I think?

ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVCache.swift
```diff
@@ -0,0 +1,485 @@
...
+    /// Used everywhere DDL and transaction control runs so SQLite failures
+    /// reach the caller. The one place callers may safely silence the throw
+    /// with `try?` is a `ROLLBACK;` on the recovery path of a prior failure —
+    /// the primary error is what the caller wants, and `exec`'s log line still
+    /// fires for the rollback failure itself.
+    private static func exec(_ sql: String, on db: OpaquePointer) throws {
+        let result = sqlite3_exec(db, sql, nil, nil, nil)
+        if result != SQLITE_OK {
+            Self.logger.error("exec failed: \(Self.describe(result)) — \(sql)")
+            throw Error.writeFailed(sqliteCode: result)
```
crazytonyli, line 640: Nitpick: The `exec` function is not just used for writing, so I think `writeFailed` may not be the right semantic for the SQLite error.

ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVStore.swift
jkmassel: Solid point!

ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVCache.swift
jkmassel: Heh I was planning to do this in the next PR as a convenience init 😁 

I think I always want the ability to specify the size in bytes, but it's nice to have some syntactic sugar atop that.

ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVCache.swift
jkmassel: This is actually deliberate – sqlite will emit an English description for any error code, so that's what goes here. Cache errors should never be shown to a user, they're just for debugging. But this way I can `print(error)` and it prints the nice description.

ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVStore.swift
jkmassel: Yeah, the approach is more FIFO and I've noted that now.

ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVCache.swift
jkmassel: Adjusted this

ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVCache.swift
```diff
@@ -0,0 +1,674 @@
...
+    /// is checked on open; anything other than this exact value triggers a
+    /// drop-and-recreate. That includes the fresh-database case (`user_version`
+    /// defaults to 0 on a brand-new SQLite file), so the first open of a new
+    /// database also runs the drop path — harmless because there's nothing to
+    /// drop. Bump only on changes that hit user devices — pre-ship iteration
+    /// shouldn't burn version numbers. Never decrement (a downgrade would
+    /// silently wipe users' caches a second time on the next upgrade).
+    private static let schemaVersion: Int32 = 1
+
+    private static let logger = Logger(subsystem: "GutenbergKit", category: "sqlite-kv-cache")
```
crazytonyli, line 134: Nitpick: use swift-log, which would work better in the app?

crazytonyli (approved):
I don't know much about sqlite3's WAL and VACUUM. If they are for improving file writing performance and accurate capacity control, I wonder if the implementation can be dumbed down a little bit:
1. Should we improve performance after we have a baseline benchmark?
2. If we can't keep capacity accurately, the worst thing is probably that we are a few megabytes (or less) over the limit, which can probably be fixed next time the "evict capacity" is invoked. That does not feel like a big deal to me?

jkmassel:
> WAL

WAL helps with performance, but more importantly correctness – see https://github.com/wordpress-mobile/GutenbergKit/pull/477/changes#diff-2dca77c03b10dc2fa896ac59deb88c66f46b98aec3c541f445eba596ba5b33ceR402-R405. It's the default for Core Data, so it's not unusual to use it this way.

> the worst thing is probably that we are a few megabytes (or less) over the limit

The issue is that without VACUUM that space is never given back to the system. We might only ever cache 100MB of data, but if we hit that number in the DB then evict every entry the DB still has ~100MB reserved on-disk. VACUUM is how we give that space back to the system (I mean...the OS can also just delete our file since it's in the cache directory, but this is slightly nicer).

ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVCache.swift
jkmassel: We don't actually have `swift-log` in this project yet. I'd be ok to adopt it, but I think that's a separate PR.

ios/Sources/GutenbergKit/Sources/Stores/SQLiteKVCache.swift
```diff
@@ -0,0 +1,674 @@
...
+            Self.logger.error("delete failed at sqlite3_step: \(Self.describe(stepResult))")
+            throw Error.writeFailed(sqliteCode: stepResult)
+        }
+    }
+
+    /// Removes all entries. Throws `Error.writeFailed` if SQLite rejects the
+    /// delete.
+    func clear() throws {
+        let db = try connection()
+        try Self.exec("DELETE FROM entries;", on: db)
```
crazytonyli, line 576: > The issue is that without VACUUM that space is never given back to the system.

This statement would need a `VACUUM`, too?

PR #478, opened 2026-04-28
feat(android): wire up block inserter search and tab filtering

jkmassel:
## Summary

Activates the tab chips and search field shipped as a non-functional shell in #469 — keeps the visual layer untouched, only adds behavior.

- **Tab filter** — selecting a category narrows the grid to a fixed id allowlist per `InserterTab`. The Recent tab surfaces the payload's `gbk-most-used` section when present, otherwise all blocks.
- **Search** — debounced 150ms, runs over the active tab's slice. Fuzzy scorer in `BlockSearch.kt` weights name/title/keyword/category matches.
- **Empty state** — renders `No results` / `No blocks match "X"` when the filter yields nothing.

Stacked on #461; rebase onto `trunk` once that lands.

## Test plan

- [x] Open the inserter on the sample app
- [x] Verify each tab chip shows appropriate blocks
- [x] Type a multi-character query quickly — confirm the debounce holds the grid steady mid-type and it settles once after you stop
- [x] Type a query that matches nothing — confirm empty-state copy renders
- [x] Clear the search — confirm the active tab's blocks come back
- [x] `./gradlew :Gutenberg:detekt :Gutenberg:assembleDebug :Gutenberg:testDebugUnitTest` passes

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/inserter/BlockPickerDialog.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/inserter/BlockPickerDialog.kt
@@ -44,6 +44,7 @@ import androidx.compose.material3.dynamicDarkColorScheme
 import androidx.compose.material3.dynamicLightColorScheme
 import androidx.compose.material3.lightColorScheme
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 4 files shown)


adalpari (approved):
Looks pretty solid and works as expected!! 🚢 it!

PR #479, opened 2026-04-28
feat(android): add photos and camera media strip to block inserter

jkmassel:
## Summary

Adds the recent-photos / Photos / Camera strip between the inserter header and category tabs.

Three runtime states for the strip, resolved by `resolveMediaStripView` in `PhotoAccessState.kt`:

1. **Rationale card** — shown when `READ_MEDIA_IMAGES` isn't granted and the user hasn't dismissed it. The card swaps body copy and primary-button label across three sub-states (`Unasked` / `Denied` / `PermanentlyDenied`); the transition to `PermanentlyDenied` is detected even after a synchronous second deny via a permission-result tick that drives recomposition. SharedPreferences tracks the first-prompt flag because `shouldShowRequestPermissionRationale` alone can't tell "never asked" from "permanently denied".
2. **Compact tiles** — when the user has tapped Reject on the rationale (persisted in SharedPreferences, sticky across launches). Photos/Camera buttons render as a 64dp full-width row; the rationale card is hidden. A granted permission overrides this state — the strip flips back to thumbnails on the next resume.
3. **Full strip** — once granted, queries MediaStore for the 64 most recent images and renders a horizontally-scrolling 2-row thumbnail grid. Lifecycle-resume aware, so revoking via system Settings → returning to the app updates the state without restart.

Photos / Camera tiles are usable in every state. Photos launches the system photo picker (permissionless via `PickVisualMedia`); Camera launches `ACTION_IMAGE_CAPTURE` against a cache-scoped FileProvider URI. Hand-off of the resulting URI into the editor is a follow-up — the tap-to-insert plumbing needs a `WebViewAssetLoader` path handler so the JS side can `fetch()` the content URI.

The library declares its own FileProvider keyed off `${applicationId}` so it won't collide with one a host app already provides. Hosts that don't need photo access can `tools:node="remove"` the manifest permissions.

`GutenbergView.resetBlockPickerPhotoPreferences(context)` is exposed for host apps that want to clear the rationale-rejection / first-prompt flags from a settings screen — also wired into the demo app's `⋮` menu as **Reset Photo Permissions Prompts** so reviewers can replay the rationale flow.

## Test plan

- [ ] First launch: rationale card shows with **Allow** + **Reject** alongside the Photos/Camera column
- [ ] Tap Allow → system prompt → grant → strip shows recent photos, behind the same Photos/Camera column
- [ ] Deny system prompt once → rationale switches to **Try Again**
- [ ] Deny system prompt twice → rationale switches to **Open Settings** (system suppresses the prompt)
- [ ] Tap Reject in rationale → rationale hides, Photos/Camera reflow into a 64dp full-width row; preference persists across dialog dismiss/relaunch and app cold-start
- [ ] After Reject, grant permission via system Settings → return to the app → strip flips to thumbnails (lifecycle resume picks up the new grant)
- [ ] Tap Photos tile → system photo picker opens (works in all three states)
- [ ] Tap Camera tile → camera app opens with a cache-scoped output URI (works in all three states)
- [ ] Vertical drag on the photo strip still drags the bottom sheet (nested-scroll relay)
- [ ] Demo app `⋮` → **Reset Photo Permissions Prompts** clears SharedPreferences; reopening the inserter shows the rationale fresh
- [ ] `./gradlew :Gutenberg:detekt :Gutenberg:assembleDebug :Gutenberg:testDebugUnitTest` passes (includes `PhotoAccessStateTest`)

### DIFF
--- a/android/Gutenberg/build.gradle.kts
+++ b/android/Gutenberg/build.gradle.kts
@@ -171,6 +171,7 @@ dependencies {
     implementation(libs.androidx.compose.ui)
     implementation(libs.androidx.compose.material3)
     implementation(libs.androidx.compose.material.icons.extended)
+    implementation(libs.androidx.activity.compose)
 
     testImplementation(libs.junit)
     testImplementation(kotlin("test"))
--- a/android/Gutenberg/src/main/AndroidManifest.xml
+++ b/android/Gutenberg/src/main/AndroidManifest.xml
@@ -5,6 +5,10 @@
     <!-- Used by the block inserter media strip to preview recent device photos.
          Host apps that don't need this can opt out via tools:node="remove". -->
     <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
+    <!-- Android 14+ partial-access opt-in. Declaring this lets the system offer
+         "Select photos and videos" alongside the all/none choice and lets us
+         detect partial grants so we can surface a "Manage selection" affordance. -->
+    <uses-permission android:name="android.permission.READ_MEDIA_VISUAL_USER_SELECTED" />
     <uses-permission
         android:name="android.permission.READ_EXTERNAL_STORAGE"
         android:maxSdkVersion="32" />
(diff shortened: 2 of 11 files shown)


jkmassel:
Superseded by four smaller PRs that split the media-strip work into reviewable chunks:

- #509 — Photos and Camera action tiles (no permissions)
- #510 — Photo-permission rationale and recent-photos thumbnail strip
- #511 — Android 14 partial-access support and Manage tile
- #512 — Touch-target accessibility fix for the existing category chips (orthogonal, independent of the above)

Closing this in favour of the stack.

PR #481, opened 2026-04-29
feat(android): render icons and wrap labels in the block inserter

jkmassel:
Two changes layered on top of #461 — wire up the `SvgIconCache` that landed there but was never called from the dialog, and tighten the tile label layout so multi-word entries wrap at whitespace while single-word entries shrink instead of mid-word breaking.

Tab filtering / search / empty state are out of scope — #478 covers them.

## Screenshots

Pixel 9 Pro XL, Recent tab. Single-word labels stay on one line at 12sp; multi-word labels wrap at whitespace; tile heights stay aligned across rows.

<table>
<tr>
<td>inserter-light</td>
<td>inserter-dark</td>
</tr>
<tr>
<td align="center">Light theme</td>
<td align="center">Dark theme</td>
</tr>
</table>

## Render block icons

Replaces the tonal placeholder `Box` in `BlockPickerDialog.BlockTile` with the real block icon, sourced from `SvgIconCache` (added alongside the dialog in #461 but never referenced from it). Without this the inserter ships as rows of identical solid chips: every block looks the same apart from its label.

- `BLOCK_TILE_ICON_GLYPH_SIZE_DP = 32` — inner glyph size, distinct from the 56dp chip. The cache renders at glyph pixels, not chip pixels, so bitmaps stay sharp at the displayed size.
- `rememberSvgIconCache(chipColor)` instantiates the cache once per `BlockGridContent` composition, keyed on `(renderSizePx, chipColor)`. A font-scale or theme change recomputes both the bitmap resolution and the contrast surrogate that `SvgIconCache.shouldTint` measures the brand colour against.
- `tintable == true` (monochrome / low-contrast brand) → `ColorFilter.tint(onSurface)`. `tintable == false` (multi-colour / sufficient-contrast brand) → render as-is. `block.isDisabled` → applied via `Image.alpha`, not folded into the tint, so branded icons fade uniformly instead of going monochrome on disable.

## Wrap labels with tight line height

`AutoShrinkTileLabel` previously shrank labels down to 9sp until they fit on a single line. Three things to fix:

1. **Single-word and multi-word were treated identically.** "Featured Image" got shrunk when it could have wrapped at the space at full size.
2. **Mid-word breaks on long single words.** Naively switching to `maxLines = 2` made Compose character-break "Preformatted" into "Preform" / "atted". Compose's default soft-wrap will break inside a word to satisfy the line budget, so single-word labels now measure with `maxLines = 1` and render with `softWrap = false` to force the shrink path instead.
3. **Loose vertical rhythm on wrapped labels.** With no explicit `lineHeight`, the gap between two wrapped lines tracked the font's default leading and read too tall. M3 typography presets handle this via `LineHeightStyle.Trim.Both`, but custom `sp + lineHeight` doesn't pick it up — set it explicitly so labels sit centred against the reserved two-line height.

`minLines = 2` reserves two lines worth of vertical space at the resolved size so tile heights stay consistent across the grid even when neighbouring labels need different line counts.

## Test plan

- [x] `./gradlew :Gutenberg:test detekt :Gutenberg:lintDebug :app:compileDebugKotlin` — BUILD SUCCESSFUL
- [x] Manually verified on Pixel 9 Pro XL with `enableNativeBlockInserter` on: icons render correctly, single-word labels (e.g. "Preformatted") shrink rather than break, multi-word labels (e.g. "Featured Image") wrap at the space, two-line labels have tight line spacing, tile heights stay aligned across the grid.

## Related

- Builds on #461.
- Tab filtering / search / empty state: #478.
- Unblocks the visual half of #469 (cards + search).

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/inserter/BlockPickerDialog.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/inserter/BlockPickerDialog.kt
@@ -69,8 +69,10 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.rememberTextMeasurer
+import androidx.compose.ui.text.style.LineHeightStyle
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.unit.Constraints
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import com.google.android.material.bottomsheet.BottomSheetBehavior
@@ -143,6 +145,7 @@ private const val BLOCK_TILE_ICON_LABEL_GAP_DP = 8
 private const val BLOCK_TILE_LABEL_SP = 12
 private const val BLOCK_TILE_LABEL_MIN_SP = 9
 private const val BLOCK_TILE_LABEL_LETTER_SPACING_SP = 0.1
+private const val BLOCK_TILE_LABEL_LINE_HEIGHT_RATIO = 1.2f
 
 private const val DISABLED_ALPHA = 0.5f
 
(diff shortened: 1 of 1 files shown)


adalpari:
A quick bug/fix:

```
  Bug — stale icon after theme/font-scale change
                                                                                                                                                                                  
  BlockPickerDialog.kt:570                                                                                                                                                        
  val rendered = remember(block.id) { iconCache.renderIcon(block) }                                                                                                               
                                                                                                                                                                                  
  iconCache is correctly re-keyed on sizePx and surfaceArgb in rememberSvgIconCache, so a theme switch or font-scale change gives BlockTile a new cache instance. However,        
  rendered is only keyed on block.id, so the old bitmap (rendered from the previous cache) is returned from remember and the new iconCache is never consulted. Icons won't update 
  until the tile leaves and re-enters composition.                                                                                                                                
                                                                                                                                                                                  
  Fix: remember(block.id, iconCache) { iconCache.renderIcon(block) }.
```

adalpari (approved):
Left a comment, about a possible bug. But other that that LGTM!

jkmassel:
> A quick bug/fix:
> 
> ```
>   Bug — stale icon after theme/font-scale change
>                                                                                                                                                                                   
>   BlockPickerDialog.kt:570                                                                                                                                                        
>   val rendered = remember(block.id) { iconCache.renderIcon(block) }                                                                                                               
>                                                                                                                                                                                   
>   iconCache is correctly re-keyed on sizePx and surfaceArgb in rememberSvgIconCache, so a theme switch or font-scale change gives BlockTile a new cache instance. However,        
>   rendered is only keyed on block.id, so the old bitmap (rendered from the previous cache) is returned from remember and the new iconCache is never consulted. Icons won't update 
>   until the tile leaves and re-enters composition.                                                                                                                                
>                                                                                                                                                                                   
>   Fix: remember(block.id, iconCache) { iconCache.renderIcon(block) }.
> ```

Nice thanks!

PR #483, opened 2026-04-30
refactor(ios): back EditorURLCache with SQLiteKVCache

jkmassel:
## Summary

- Switches `EditorURLCache`'s backing store from `URLCache` to `SQLiteKVCache` (introduced in #477).
- `EditorURLCache.init` shape changed — `init(cacheRoot:cachePolicy:)` is now `init(siteId:parentDirectory:cachePolicy:)`. `store`/`hasData`/`response`/`clear` are unchanged, as is per-(URL, method) keying and the 100 MB cap. No callers outside this repo construct `EditorURLCache` directly (verified via `gh api search/code`).
- Drops the `Thread.sleep(...)` settling hacks after `store` and `clear`, and the `__fixEmptyDataBug{Encode,Decode}` empty-`Data` sentinel — both were workarounds for `URLCache` quirks that `SQLiteKVCache` doesn't have.

## What's stored

| field | source |
|---|---|
| key | `"<method> <url.absoluteString>"` — `SQLiteKVCache` SHA-256-hashes it before binding, so length/escaping/encoding aren't concerns |
| storageDate | passed through to `SQLiteKVCache.Entry.storageDate` instead of being stuffed into `URLCache`'s `userInfo` dict |
| metadata | `JSONEncoder().encode(EditorHTTPHeaders)` — small JSON blob |
| value | `response.data` — raw bytes, no base64 overhead |

## Notes

- Cache policy filtering (`.ignore` / `.maxAge(_:)` / `.always`) still happens on read, against `Entry.storageDate`. The internal testing overloads that take `currentDate: Date` are preserved.
- **Eviction policy changed** — `URLCache` evicts least-recently-used; `SQLiteKVCache` evicts oldest-by-`storage_date` (insert/refresh order). Under cap pressure, a frequently-read old entry is now evicted before a recently-stored unread one. The cap isn't approached in practice — typical entry count is small (editor settings, theme, post type list) — but the policy is different.
- `SQLiteKVCache.clear()` is synchronous, so the `waitForClearToTakeEffect` polling helper in the tests is gone.
- **Migration:** pre-existing `URLCache` files in `cacheRoot` (a `Cache.db` and `fsCachedData/`) are left in place — dead weight until the OS reclaims the cache directory. Existing entries cache-miss on first read after upgrade and refetch from the network. Acceptable for a cache; a one-shot cleanup on init is possible but not worth the complexity.
- **One instance per backing file** carries over from `SQLiteKVCache`: two `EditorURLCache` instances pointed at the same `(siteId, parentDirectory)` is undefined behavior. `EditorService` creates one cache per service instance, so the typical app flow is fine.

## Test plan

- [x] All 36 `EditorURLCacheTests` pass (covers round-trip, overwrite, empty data, file-store, header round-trip, query-parameter independence, all three cache policies, and clear).
- [x] Full `swift test` suite passes (891 tests / 46 suites).
- [ ] CI runs the package against iOS via `make test-swift-package`.

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/Stores/EditorURLCache.swift
+++ b/ios/Sources/GutenbergKit/Sources/Stores/EditorURLCache.swift
@@ -1,4 +1,3 @@
-import CryptoKit
 import Foundation
 import OSLog
 
@@ -6,8 +5,15 @@ import OSLog
 ///
 /// Responses are stored on disk and survive process termination. Responses are keyed by both URL and HTTP method, so GET and OPTIONS requests to the
 /// same URL are stored independently.
+///
+/// Backed by `SQLiteKVCache`. The "one instance per backing file" contract from
+/// `SQLiteKVCache` carries over: two `EditorURLCache` instances pointed at the
+/// same `cacheRoot` is undefined behavior.
 public struct EditorURLCache: Sendable {
-    private let cache: URLCache
+    /// About enough for 10 sites of cached responses.
+    private static let diskCapacity = 100 * 1024 * 1024
+
+    private let store: SQLiteKVCache
     private let cachePolicy: EditorCachePolicy
     private let performanceMonitor = SignpostMonitor(for: Logger.performance)
 
@@ -21,8 +27,11 @@ public struct EditorURLCache: Sendable {
     ///     Use `.ignore` to always fetch fresh data, `.maxAge(_:)` to expire entries after
     ///     a time interval, or `.always` (the default) to use cached data regardless of age.
     public init(cacheRoot: URL? = nil, cachePolicy: EditorCachePolicy = .always) {
-        // About enough for 10 sites
-        self.cache = URLCache(memoryCapacity: 0, diskCapacity: 100 * 1024 * 1024, directory: cacheRoot)
+        self.store = SQLiteKVCache(
+            handle: "editorurlcache",
+            directory: cacheRoot ?? URL.cachesDirectory,
+            diskCapacity: Self.diskCapacity
+        )
         self.cachePolicy = cachePolicy
     }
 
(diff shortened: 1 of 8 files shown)


ios/Sources/GutenbergKit/Sources/Stores/EditorURLCache.swift
```diff
@@ -139,49 +138,31 @@ public struct EditorURLCache: Sendable {
...
-    ///
-    /// `URLCache` doesn't properly store an empty `Data` object, so we use a sentinel
-    /// value to represent empty data.
-    ///
-    /// - Parameter data: The data to encode.
-    /// - Returns: The encoded data, or a sentinel value if the data is empty.
-    private func __fixEmptyDataBugEncode(_ data: Data) -> Data {
-        guard data.isEmpty else { return data }
-        return Data("__is_empty__".utf8)
+        try self.store.clear()
```
crazytonyli, line 159: I added https://github.com/wordpress-mobile/GutenbergKit/pull/477#discussion_r3170823253 about the clear function, just in case you missed it.

crazytonyli (approved):

ios/Sources/GutenbergKit/Sources/Stores/EditorURLCache.swift
jkmassel: I did miss it! This is a good question – we could do it at that time, but I think I prefer freeing space on startup each time? Every time we do it, the entire file is re-written – that means we want to do it _sorta_ frequently but not too often that we're using up SSD write cycles? 

I thought on-launch was ideal, but LMK if you think otherwise!

crazytonyli: Maybe it's too much to update the entire file on every single update or insert, but I don't imagine `clear` would be called frequently? I don't feel too strongly about it, though.

jkmassel: The main reason `clear` should exist (as a public method anyway) is pull-to-refresh which definitely isn't common. I don't think that's a particularly great time to perform this bookkeeping though.

I'm definitely open to changing it later if we discover it's preferable that way – I don't think it really violates the object contract to adjust that later.

PR #487, opened 2026-05-04
fix(js): skip redux toggle when opening native block inserter

jkmassel:
## Summary

Tapping the `+` button currently calls `setIsInserterOpened(true)` on the editor store, which momentarily renders Gutenberg's web inserter UI before the native dialog covers it — a visible flash on slower devices. Call `prepareAndShowInserter()` directly instead; the redux flag isn't useful when the inserter is rendered natively.

## Test plan

- [x] On Android (or iOS), with `enableNativeBlockInserter` on, tap the `+` toolbar button
- [x] Native inserter dialog appears immediately, no flash of the web inserter behind it
- [x] The dialog still closes correctly via the X button (still goes through `onToggle(false)` from `window.blockInserter.onClose`)

### DIFF
--- a/src/components/native-inserter/index.jsx
+++ b/src/components/native-inserter/index.jsx
@@ -424,7 +424,11 @@ export default function NativeBlockInserterButton( { open, onToggle } ) {
 			title={ __( 'Add block' ) }
 			icon={ plus }
 			onClick={ () => {
-				onToggle( true );
+				// Skip the redux toggle and present the native inserter
+				// directly. Flipping `isInserterOpened` in editorStore would
+				// briefly render Gutenberg's web inserter behind the native
+				// dialog before it covers, causing a visible flash.
+				prepareAndShowInserter();
 			} }
 			onMouseDown={ ( e ) => {
 				e.preventDefault();


oguzkocer (commented):
@jkmassel Do you have a recording of the flash this is fixing? Because I tried recording before and after and I don't see any change. I am working with Claude to see maybe it's just a build issue on my end because I tested the fix before I tested the base branch. Here are the recordings:

Recording from: `jkmassel/android-block-picker`:

https://github.com/user-attachments/assets/9be6fef0-db4f-4090-8620-a1f4774a4bc9

Recording from `jkmassel/native-inserter-no-redux-flash`:

https://github.com/user-attachments/assets/925e2ee8-587a-491d-bb35-de9912baa8d7

oguzkocer (approved):
I think maybe the reason I am not seeing a flash is this:

> a visible flash on slower devices.

So, I am going to approve the PR on the code change and the fact that I am not observing a flash after the change. 🤷

dcalhoun:
@jkmassel based on my understanding of the native inserter JavaScript code, I believe this change may not actually have the desired impact. 

Gutenberg's own `Inserter` component is what [renders](https://github.com/WordPress/gutenberg/blob/a60c8b9eaa3bf20c0965730348d7474f7bc6510a/packages/block-editor/src/components/inserter/index.js#L362-L376) the block types list web UI (dropdown menu content + toggle button). When the native inserter is enabled, the Gutenberg web `Inserter` is [not rendered](https://github.com/wordpress-mobile/GutenbergKit/blob/5718c8014877060462811188729c10d11996c187/src/components/editor-toolbar/index.jsx#L100-L114) in the React tree. 

The only visible flash I was able to reproduce appears related to animating the position of the native inserter while the virtual keyboard is open. This occurred with and without the `prepareAndShowInserter()` change. Does the following video resemble you've observed? If so, I suggest we revert these changes—WDYT? 


https://github.com/user-attachments/assets/a41c502d-c108-42dc-bb86-bdaa8da9bcb7

PR #491, opened 2026-05-04
feat: resolve consumer locales against shipped translation bundles

### LINKED ISSUE #490: Resolve consumer locales against shipped translation bundles
## Summary

Consumers currently pass `locale` to `EditorConfiguration` as an opaque string, and `src/utils/localization.js` does a single-level dynamic import. If the tag doesn't match a shipped `translations/*.json` exactly, the editor silently falls back to English. The supported-locales list lives in `bin/prep-translations.js`; consumers don't see it, and have no way to ask "what's the closest bundle you actually ship?". This means each client has to mirror that table to do the right thing — and historically hasn't.

Concretely on WP-Android (today): `PerAppLocaleManager.getCurrentLocaleLanguageCode()` returns `Locale.getLanguage()` — ISO 639-1 only — so a Brazilian user sends `pt` (matches), a Chinese user sends `zh` (no `zh-cn`/`zh-tw` match → English), and a user who picked `en_GB` / `nl_BE` / `pt_BR` in the language picker loses their region variant before we ever reach GutenbergKit.

## Proposed shape

Move locale resolution into the library so every client gets correct behaviour without re-implementing the table:

### 1. Platform-native `setLocale` overloads

```kotlin
// Android
setLocale(locale: Locale)      // new — resolves internally
setLocale(locale: String)      // existing — power-user / tests
```

```swift
// iOS
func setLocale(_ locale: Locale)        // new
func setLocale(_ identifier: String)    // existing
```

The new overloads call `toLanguageTag().lowercased()` (or `Locale.identifier` on iOS) and run the resolution chain before storing the string for serialization. The wire format stays unchanged — only the consumer-facing API gains type safety.

### 2. Resolution chain

For an input tag `xx-yy`:

1. Full tag (`xx-yy`) — match if shipped
2. Language-only tag (`xx`) — match if shipped
3. Fall back to `en`

Determined at runtime against the actually-bundled translation files, not a hard-coded list — that way the resolver stays correct when `SUPPORTED_LOCALES` in `bin/prep-translations.js` changes without anyone remembering to update a duplicate.

### 3. JS-side fallback becomes defensive

`localization.js`'s catch-and-warn stays as a defense-in-depth net but should rarely fire in practice once the Android/iOS resolver narrows to a known-shipped tag.

## Out of scope

- Changing the `SUPPORTED_LOCALES` list itself.
- Changing the wire format of `EditorConfiguration.locale`.
- Pluralization / RTL handling — separate concerns.

## Why this matters

Two clients today (WP-Android, WP-iOS) and a third on the way (the GutenbergKit preloader being added in WP-Android #22579, where every preloaded session has to remember to overlay the locale or it ships English regardless of device language). One resolver in the library closes all three at once and any future client by default.

## Related

- WP-Android PR that made the footgun visible: https://github.com/wordpress-mobile/WordPress-Android/pull/22836

jkmassel:
## Summary

- Adds a single resolver — full normalised tag → language-only tag → `en` — implemented identically on JS, iOS, and Android, fed by a build-emitted `supported-locales.json` manifest so the "what do we ship?" answer has exactly one source of truth.
- Adds `setLocale(_ locale: Locale)` (iOS) and `setLocale(context: Context, locale: Locale)` (Android) overloads on `EditorConfiguration` builders. The wire format stays a string; only the consumer-facing API gains type safety.
- JS-side `localization.js` now applies the same resolution chain instead of hoping the consumer's tag matches a shipped filename exactly.

Fixes #490.

## Root Cause

`EditorConfiguration.locale` is an opaque string. Consumers (WP-iOS, WP-Android, the upcoming WP-Android preloader in wordpress-mobile/WordPress-Android#22579) historically passed whatever the platform handed back — often `Locale.getLanguage()` on Android, ISO 639-1 only — and `src/utils/localization.js` did a single-level dynamic `import` that silently fell back to English on any miss. The supported list lived only in `bin/prep-translations.js`, so each client had to mirror that table to do the right thing — and historically hadn't.

### Manifest

**vite.config.js**: New `emitSupportedLocalesManifest` plugin scans `src/translations/` at build time and emits `dist/supported-locales.json` (sorted array of locale tags). The existing `copy-dist-{ios,android}` Make targets ship it into both bundles.

### Resolution chain

For an input `xx-yy`, normalised to lowercase with `_` → `-`:

1. Full tag (`xx-yy`) — match if shipped ✅
2. Language-only tag (`xx`) — match if shipped ✅
3. Fall back to `en` ✅

Implemented in:

- **src/utils/localization.js** — `resolveLocale(input, supported?)`. Default supported set comes from `import.meta.glob('../translations/*.json')`, so JS is statically analysable and gets the same single source of truth without a separate fetch.
- **ios/Sources/GutenbergKit/Sources/Model/LocaleResolver.swift** — `LocaleResolver` struct, fed by `GutenbergKitResources.loadSupportedLocales()` reading the manifest from `Bundle.module`.
- **android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/LocaleResolver.kt** — `LocaleResolver` class with a `fromAssets(context)` factory that reads the manifest from the library's `assets/`.

### New consumer APIs

```swift
// iOS — runs the resolution chain
public func setLocale(_ locale: Locale) -> EditorConfigurationBuilder
```

```kotlin
// Android — Context required to read the manifest from assets
fun setLocale(context: Context, locale: Locale)
```

The existing `setLocale(String)` overloads stay untouched as a power-user / test escape hatch.

## What We Explored

- **Hard-coding the supported set in native code** — explicitly called out as the footgun in the issue. Rejected.
- **Reading the locale list from filenames in `assets/`** — Vite hashes the chunk filenames (`pt-br-UCkBcRdR.js`), and the prefixes can collide (`nl-be-...` vs `nl-...`), so a parser would have to re-encode the SUPPORTED_LOCALES list anyway. A manifest is simpler and unambiguous.
- **`androidx.startup` / hidden static Context** — would let `setLocale(Locale)` skip the `context:` parameter on Android. Rejected for now: adds a dependency and lifecycle surprise for one parameter; consumers calling this from an editor activity already have a `Context` in hand.

## Behaviour change for the cases the issue called out

| Input | Before | After |
| --- | --- | --- |
| `pt_BR` device locale | `pt` → matches → ✅ Brazilian | `pt-br` → matches → ✅ Brazilian |
| `pt_BR` via `Locale.getLanguage()` | `pt` → matches → ✅ Brazilian | (consumer can pass full `Locale`) `pt-br` → matches → ✅ Brazilian |
| `nl_BE` | `nl` → matches → 🇳🇱 Dutch | `nl-be` → matches → 🇧🇪 Belgian Dutch |
| `fr_CA` (no regional bundle) | `fr-CA` → ❌ English | `fr` → ✅ French |
| `Locale("zh","CN")` | `zh-CN` → ❌ English | `zh-cn` → ✅ Simplified Chinese |
| `zh` (language-only, no zh bundle) | ❌ English | ❌ English (unchanged — language-only `zh` isn't shipped) |

## Test plan

- [x] `npx vitest run src/utils/localization.test.js` — 4/4 pass
- [x] `swift test --filter "EditorConfigurationBuilderTests|LocaleResolverTests"` — 37/37 pass
- [x] `./android/gradlew :Gutenberg:testDebugUnitTest --tests "org.wordpress.gutenberg.model.LocaleResolverTest"` — 5/5 pass
- [x] `./android/gradlew :Gutenberg:testDebugUnitTest --tests "org.wordpress.gutenberg.model.EditorConfigurationBuilderTest"` — existing suite still green
- [x] `npm run lint:js -- src/utils/localization.js src/utils/localization.test.js vite.config.js` — clean
- [x] `./android/gradlew detekt` — clean
- [ ] Manual smoke: build the demo apps, point them at `Locale("pt","BR")` / `Locale("nl","BE")` / `Locale("fr","CA")`, confirm the editor UI renders in the expected language.

## Out of scope

- Changing the `SUPPORTED_LOCALES` list itself.
- Changing the wire format of `EditorConfiguration.locale` — still an opaque string on the JS side.
- Pluralization / RTL handling — separate concerns.

## Related

- WP-Android PR that made the footgun visible: wordpress-mobile/WordPress-Android#22836
- WP-Android preloader work that triggered this: wordpress-mobile/WordPress-Android#22579

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorConfiguration.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/EditorConfiguration.kt
@@ -1,9 +1,11 @@
 package org.wordpress.gutenberg.model
 
+import android.content.Context
 import android.os.Parcelable
 import kotlinx.parcelize.IgnoredOnParcel
 import kotlinx.parcelize.Parcelize
 import java.net.URI
+import java.util.Locale
 import java.util.UUID
 
 @Parcelize
@@ -97,6 +99,26 @@ data class EditorConfiguration(
         fun setAuthHeader(authHeader: String) = apply { this.authHeader = authHeader }
         fun setEditorSettings(editorSettings: String?) = apply { this.editorSettings = editorSettings }
         fun setLocale(locale: String?) = apply { this.locale = locale }
+
+        /**
+         * Resolves [locale] against the bundled translations and stores the
+         * resulting tag for serialization.
+         *
+         * Prefer this overload when the locale comes from a platform API
+         * (e.g. [android.content.res.Configuration.getLocales] or the user's
+         * per-app locale). It runs the resolution chain — full tag →
+         * language-only tag → `en` — using the manifest shipped in this
+         * library's assets, so a device configured for `pt_BR` correctly
+         * lands on the `pt-br` bundle instead of silently falling through
+         * to English.
+         *
+         * @param context Used to read the shipped translations manifest
+         *                from this library's assets.
+         * @param locale  The platform locale to resolve.
+         */
+        fun setLocale(context: Context, locale: Locale) = apply {
+            this.locale = LocaleResolver.fromAssets(context).resolve(locale)
+        }
         fun setCookies(cookies: Map<String, String>) = apply { this.cookies = cookies }
         fun setEnableAssetCaching(enableAssetCaching: Boolean) = apply { this.enableAssetCaching = enableAssetCaching }
         fun setCachedAssetHosts(cachedAssetHosts: Set<String>) = apply { this.cachedAssetHosts = cachedAssetHosts }
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/LocaleResolver.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/model/LocaleResolver.kt
@@ -1,0 +1,76 @@
+package org.wordpress.gutenberg.model
+
+import android.content.Context
+import kotlinx.serialization.json.Json
+import java.util.Locale
+
+/**
+ * Resolves an arbitrary locale tag to one of the bundles GutenbergKit
+ * actually ships translations for.
+ *
+ * Consumers historically hand [EditorConfiguration] an opaque locale string
+ * — on Android, often the output of [Locale.getLanguage], which strips the
+ * region. The editor then silently falls back to English whenever the tag
+ * doesn't match a shipped `translations/<tag>.json` file exactly. The
+ * resolver moves that decision into the library, so a device configured for
+ * `pt_BR` ends up with the Brazilian Portuguese bundle — and a tag like
+ * `nl-BE`, for which we don't ship a regional bundle, falls back to `nl`
+ * instead of all the way to English.
+ *
+ * Resolution chain for an input `xx-yy`:
+ *   1. Full normalised tag (`xx-yy`)
+ *   2. Language-only tag (`xx`)
+ *   3. `en`
+ *
+ * The supported set is read from a manifest emitted by the JS build, so it
+ * stays in sync with what the bundle actually ships.
+ */
+class LocaleResolver internal constructor(supportedLocales: Collection<String>) {
+    private val supportedLocales: Set<String> =
+        supportedLocales.map { normalize(it) }.toSet()
+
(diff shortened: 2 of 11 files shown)


jkmassel:
Splitting into separate iOS and Android PRs for independent review/landing. Replacement PRs incoming — iOS carries the foundational build manifest + JS resolver, Android stacks on top.

PR #499, opened 2026-05-11
ci: have Dependabot watch the Bundler ecosystem too

jkmassel:
## Summary

- #498 has since landed and brought us up to `fastlane-plugin-wpmreleasetoolkit ~> 14.4` manually. Had this dependabot entry already existed, that bump would have surfaced as an automatic PR and we would not have needed the manual catch-up — which is exactly the gap this PR closes for next time.

## Changes

- **`.github/dependabot.yml`**: add a second `updates:` entry for `package-ecosystem: 'bundler'`, weekly cadence, no ignores. The existing `npm` block is unchanged.

## Test plan

- [ ] After merge, Dependabot's next scan posts a PR for any outstanding `bundler` updates (release-toolkit is now current at 14.4.1 thanks to #498; `fastlane` is still at 2.232.2 vs latest 2.234.0).
- [ ] Weekly schedule kicks in on the standard Dependabot day; no further intervention needed.


mokagio (approved):
>  — which is exactly the gap this PR closes for next time.

Love it. Thanks!

PR #509, opened 2026-05-13
feat(android): add Photos and Camera tiles to block inserter

jkmassel:
## Summary

Adds a Photos / Camera quick-launch row between the inserter's header and category tabs. First slice of a four-part split of #479; covers just the permissionless tiles.

- **Photos tile** uses `ActivityResultContracts.PickVisualMedia` — the system photo picker, no manifest permission required.
- **Camera tile** uses `ACTION_IMAGE_CAPTURE` against a cache-scoped FileProvider URI. The library does not declare or request `CAMERA` — but per the [`MediaStore.ACTION_IMAGE_CAPTURE`](https://developer.android.com/reference/android/provider/MediaStore#ACTION_IMAGE_CAPTURE) docs, if a host app *does* declare `CAMERA` in its merged manifest, Android requires the host to hold it at runtime before the intent dispatches. That's a host-app concern, not a library one — flagging it here so the follow-up that wires up the hand-off doesn't get surprised.

The library declares its own `FileProvider` keyed off `${applicationId}.gutenberg.fileprovider` so it can't collide with one a host app already declares.

**No `uses-permission` lines are added in this PR.** The recent-photos thumbnail strip that needs `READ_MEDIA_IMAGES` lands in the follow-up.

| Inserter, light mode | Inserter, dark mode |
|:---:|:---:|
| Light mode | Dark mode |

## Design notes

- The icon and label inside each tile are deliberately tight — no `Spacer` between them. The block tiles below use an 8dp gap; the media tiles are intentionally punchier to read as quick actions, not as members of the block grid.

## What does not land here (follow-ups)

- Recent-photos thumbnail strip, photo-permission rationale, and the `READ_MEDIA_IMAGES` / `READ_EXTERNAL_STORAGE` declarations — next PR in the stack.
- Android 14 `READ_MEDIA_VISUAL_USER_SELECTED` partial-access support and the "Manage" tile — PR after that.
- Touch-target accessibility fix for the existing category chips — independent PR.

## Result-callback inertness

Both launcher result callbacks are intentionally empty:

```kotlin
ActivityResultContracts.PickVisualMedia()
) { /* picked uri — hand-off to editor insertion is a follow-up */ }
```

The picked URI / camera capture needs to round-trip through `WebViewAssetLoader` so the JS editor can `fetch()` it — that hand-off is a follow-up. Until then, this whole sheet is gated behind the demo's "Enable Native Inserter" toggle, so users outside that opt-in don't see the no-op buttons.

A `TODO` on `createCameraOutputUri` flags that each Camera tap creates an orphan file in the cache today; cleanup lands when the hand-off does.

## Test plan

- [x] Open the inserter in the demo app. Photos and Camera tiles render above the category tabs.
- [x] Tap **Photos** → system photo picker opens. Pick or cancel → sheet returns to the inserter intact.
- [x] Tap **Camera** → camera app opens with a cache-scoped output URI. Capture or cancel → sheet returns to the inserter intact.
- [ ] Camera tile is hidden on devices that don't report `FEATURE_CAMERA_ANY` (tablet/emulator without a camera).
- [ ] `./gradlew :Gutenberg:detekt :Gutenberg:assembleDebug :Gutenberg:testDebugUnitTest` passes.

### DIFF
--- a/android/Gutenberg/build.gradle.kts
+++ b/android/Gutenberg/build.gradle.kts
@@ -171,6 +171,7 @@ dependencies {
     implementation(libs.androidx.compose.ui)
     implementation(libs.androidx.compose.material3)
     implementation(libs.androidx.compose.material.icons.extended)
+    implementation(libs.androidx.activity.compose)
 
     testImplementation(libs.junit)
     testImplementation(kotlin("test"))
--- a/android/Gutenberg/src/main/AndroidManifest.xml
+++ b/android/Gutenberg/src/main/AndroidManifest.xml
@@ -3,4 +3,20 @@
 
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 
+    <application>
+        <!-- Exposes a writeable temp-file URI to the camera app so captures can
+             come back to us without the host app having to configure its own
+             FileProvider. Authority is keyed off the host app's applicationId
+             so it won't collide with one a host app already declares. -->
+        <provider
+            android:name="androidx.core.content.FileProvider"
+            android:authorities="${applicationId}.gutenberg.fileprovider"
+            android:exported="false"
+            android:grantUriPermissions="true">
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/gbk_file_paths" />
+        </provider>
+    </application>
+
 </manifest>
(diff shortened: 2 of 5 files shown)


adalpari (approved):
Just made a small change related to recomposition.
But overall, the PR looks good to me! 🚢 it!!

PR #513, opened 2026-05-13
ci: stop refetching translations and rebuilding JS in downstream jobs

jkmassel:
## Summary

- Cut PR build wall-time from **13:15 → 9:00** (~32% faster) and per-build agent-time by ~10 minutes.
- Single source of truth for translation fetching and JS compilation: only `Build React App` does either, and every other job consumes the `dist.tar.gz` artifact.
- Drop `node-fetch` in favour of Node 20's built-in `fetch` so the translation fetch can run in parallel with `npm ci`.

## Speedup

Compared against trunk #2251 (last clean run before this PR):

| Step | Trunk #2251 | This PR (#2307) | Δ |
|---|---|---|---|
| Build React App | 2:43 | 1:50 | **−53s** |
| Test Android Library | 4:19 | 2:43 | **−1:36** |
| iOS Simulator Tests | 4:46 | 2:57 | **−1:49** |
| Library Tests | 3:22 | 1:18 | **−2:04** |
| Build XCFramework | 4:55 | 1:29 | **−3:26** |
| Test Android Library Instrumented | 2:11 | 0:57 | **−1:14** |
| Test iOS E2E | 5:16 | 5:22 | similar |
| Test Web E2E | 6:00 | 6:07 | similar |
| Other steps | (no change) | (no change) | noise |
| **Total wall-time** | **13:15** | **9:00** | **−4:15** |

`Build XCFramework` was on the wall-time critical path; trimming `npm ci` + JS build + translation fetch from it accounts for most of the headline win.

## Root Cause

Three compounding inefficiencies:

### 1. `build-react` re-fetched translations every build

`make build REFRESH_L10N=1 STRICT_L10N=1` defeated the Makefile's existence-based cache (`Makefile:48`), forcing all ~50 locales to be re-fetched from translate.wordpress.org on every CI build. Buildkite agents don't share state, so every build paid the full GlotPress round-trip cost.

### 2. Every downstream \`: build\` consumer re-fetched translations *and* re-ran \`npm ci\`

`test-swift-library`, `test-swift-package`, `test-android`, `test-android-library-e2e`, and `build-resources-xcframework` all declare `build` as a Make prereq. `build: npm-dependencies prep-translations`, so both ran first. On agents that just extracted an upstream `dist.tar.gz`:

- `dist/` is populated → `build`'s recipe short-circuited the JS build
- `src/translations/` is empty (not in `dist.tar.gz`) → `prep-translations` ran the full fetcher
- `node_modules` is empty → `npm-dependencies` ran `npm ci`
- All of that work was then thrown away because `dist/` already had everything baked in

### 3. Two Android jobs rebuilt `dist/` from scratch

`Test Android Library` and `Test Android Library Instrumented` had no upstream dependency, so they walked the full `: build` chain — fetch all locales, `npm ci`, `npm run build`, copy assets — before finally running gradle.

### `.buildkite/pipeline.yml`

- **`Build React App`**: backgrounds `make prep-translations STRICT_L10N=1` while running `make npm-dependencies` in the foreground, then `wait`s on the fetch before `make build REFRESH_JS_BUILD=1`. Shell vars escaped with `$$` for Buildkite interpolation.
- **`Test Android Library`** and **`Test Android Library Instrumented`**: gain `depends_on: build-react` and download `dist.tar.gz`.

### `Makefile`

- **`prep-translations`**: gains a "skip if `dist/` exists" clause. Translations are baked into the bundle at JS build time, so downstream consumers don't need a fresh fetch. `REFRESH_L10N=1` and direct invocation still force the fetch.
- **`build`**: drops `npm-dependencies` from its prereq list and invokes it via `$(MAKE)` inside the rebuild branch. Downstream `: build` consumers (which run gradle / xcodebuild / swift, not anything from `node_modules/.bin`) no longer pay for an `npm ci` they don't need.
- **`npm-dependencies`**: drops the "direct invocation forces install" clause — `build`'s recipe now invokes it via `$(MAKE)` which would otherwise trigger the force. `REFRESH_DEPS=1` remains as the explicit force.
- **`test-android`** and **`test-android-library-e2e`**: copy `dist/` into Android assets unconditionally. `build`'s recipe short-circuits `copy-dist-android` when `dist/` already exists, so CI agents that download a dist artifact would otherwise have stale assets. Matches what `test-android-e2e:329-331` and `build-resources-xcframework:131` already do.

### `bin/prep-translations.js`

Drop `import fetch from 'node-fetch'` in favour of Node 20's built-in global `fetch`. The script's surface (`response.ok`, `response.status`, `response.headers.get`, `response.json()`) is API-compatible with WHATWG Fetch in both implementations. Also drops `node-fetch` from `package.json` and `package-lock.json` (98 lines of lockfile).

## Decision matrix for `prep-translations`

| `dist/` | `src/translations/` | MAKECMDGOALS | `REFRESH_L10N` | Outcome |
|---|---|---|---|---|
| absent | empty | `build` (or anything) | unset | fetch |
| absent | populated | `build` (or anything) | unset | skip (bundles present) |
| present | either | not `prep-translations` | unset | **skip (dist built) — new** |
| either | either | `prep-translations` | unset | fetch (direct invoke) |
| either | either | anything | `1` | fetch |

### 1. CI-level cache with a weekly key ❌

Cache `src/translations/` with a key that rolls weekly. Faster on most builds but accepts up to a week of staleness in shipped strings. Rejected — fresh-per-build is cheap once we stop fetching per-job.

### 2. Run `prep-translations` as a separate Buildkite step ❌

The first iteration of this PR did exactly that. It bought the agent-time win but added 2:39 of serial wall-time on the critical path because the fetch waited for its own agent startup + npm install. Replaced by the in-step parallel approach (see commit `9b4e35c1`).

### 3. Swap hand-rolled fetcher for `Fastlane::Helper::GlotpressDownloader`

Separate concern — doesn't change CI speed. Tracked as a follow-up; see the [release-toolkit helper](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/helper/glotpress_downloader.rb).

## Test plan

- [x] CI green on this PR (build #2307, 9:00 total)
- [x] Per-job wall-time savings on `Library Tests`, `iOS Simulator Tests`, `Test Android Library`, `Build XCFramework`, and `Test Android Library Instrumented` — confirms the `dist/`-based skip is firing
- [x] Local: `make prep-translations` still fetches when invoked directly (explicit user intent)
- [x] Local: `make test-swift-library` after a successful `make build` logs both "Skipping translations fetch (dist/ already built…)" and "Skipping NPM dependencies installation (node_modules already exists)" — and does not invoke `npm ci`
- [ ] Spot-check that `STRICT_L10N=1` still fails the prep step on a real GlotPress error (not exercised by this PR's builds — verified previously)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,24 +47,24 @@ npm-dependencies: ## Install npm dependencies
 
 .PHONY: prep-translations
 prep-translations: ## Fetch and cache locale string files
-# Skip when `dist/` already exists *and* `src/translations/` is
-# populated — translations are baked into the bundle at JS build time,
-# so there is nothing for a downstream consumer to refresh until the
-# bundle itself is rebuilt. This matters on CI agents that download
-# `dist.tar.gz` from an upstream job: without it, every downstream
-# `make` target that depends on `build` would re-fetch all ~50 locales
-# from translate.wordpress.org.
+# Skip when `dist/` already exists — translations are baked into the
+# bundle at JS build time, so there is nothing for a downstream
+# consumer to refresh until the bundle itself is rebuilt. This matters
+# on CI agents that download `dist.tar.gz` from an upstream job:
+# without it, every downstream `make` target that depends on `build`
+# would re-fetch all ~50 locales from translate.wordpress.org only to
+# discard the result when `build`'s recipe short-circuits.
 #
-# We also require `src/translations/` to be populated so that a stale
-# `dist/` paired with an empty `src/translations/` doesn't silently
-# produce a no-translations rebuild when `REFRESH_JS_BUILD=1` or a
-# direct `make build` triggers JS rebuilding against empty inputs.
+# Use `REFRESH_L10N=1` (which still forces the fetch) for the explicit
+# "I want fresh translations on disk" workflow — `rm -rf
+# src/translations && make build` alone will not refetch, since `dist/`
+# being present is read as "translations already shipped".
 #
 # Otherwise, skip unless...
 # - src/translations doesn't contain any fetched bundles (only `.gitkeep` is committed)
 # - REFRESH_L10N is set to true or 1
 # - prep-translations was invoked directly
-	@if [ -d "dist" ] && [ -n "$$(find src/translations -maxdepth 1 -name '*.json' -print -quit 2>/dev/null)" ] && [ "$(REFRESH_L10N)" != "true" ] && [ "$(REFRESH_L10N)" != "1" ] && ! echo "$(MAKECMDGOALS)" | grep -q "^prep-translations$$"; then \
+	@if [ -d "dist" ] && [ "$(REFRESH_L10N)" != "true" ] && [ "$(REFRESH_L10N)" != "1" ] && ! echo "$(MAKECMDGOALS)" | grep -q "^prep-translations$$"; then \
 		echo "--- :white_check_mark: Skipping translations fetch (dist/ already built, translations baked in). Use REFRESH_L10N=1 to force refresh."; \
 	elif [ -z "$$(find src/translations -maxdepth 1 -name '*.json' -print -quit 2>/dev/null)" ] || [ "$(REFRESH_L10N)" = "true" ] || [ "$(REFRESH_L10N)" = "1" ] || echo "$(MAKECMDGOALS)" | grep -q "^prep-translations$$"; then \
 		echo "--- :npm: Preparing Translations"; \
--- a/Makefile
+++ b/Makefile
@@ -31,8 +31,14 @@ npm-dependencies: ## Install npm dependencies
 # Skip unless...
 # - node_modules doesn't exist
 # - REFRESH_DEPS is set to true or 1
-# - npm-dependencies was invoked directly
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -41,11 +47,26 @@ npm-dependencies: ## Install npm dependencies
 
 .PHONY: prep-translations
 prep-translations: ## Fetch and cache locale string files
-# Skip unless...
+# Skip when `dist/` already exists *and* `src/translations/` is
+# populated — translations are baked into the bundle at JS build time,
```
oguzkocer: I'm not familiar with this pipeline, but if I am understanding it correctly, the translations are baked into the bundle we get from `dist/` which means if it exists, we don't need translations at all. However, the check `[ -n "$$(find src/translations -maxdepth 1 -name '*.json' -print -quit 2>/dev/null)" ]` makes it so that it'll still get downloaded - and not used?

If that understanding is correct, it'd mean that we are downloading translations in downstream CI builds where we have `dist/` but not the translations folder, and then simply discard the newly downloaded translations because there is no build step.

I think we can skip that check and save some time in CI steps such as `Test Android Library`.

oguzkocer (approved):
Looks good. I raised an issue, but that's not a regression, a possibly missed opportunity instead. So, I am approving it as is.

PR #518, opened 2026-05-19
feat(android): gate block inserter media strip behind config flag

jkmassel:
## What?

Adds `EditorConfiguration.enableInserterMediaStrip` (defaulting to `false`) and threads it into `BlockPickerDialog`. When the flag is off — the new default — the Photos + Camera media strip is omitted from the native block inserter sheet, so the sheet renders the drag handle / header / category tabs / search / block grid without the strip in between.

## Why?

The media strip landed in #509 with intentionally inert tap callbacks: the picked URI / camera capture still needs to round-trip through `WebViewAssetLoader` before the JS editor can `fetch()` it. Until that hand-off lands, the no-op buttons shouldn't be visible to production consumers. Gating the strip behind a config flag (off by default) keeps the surface available for testing while hiding it from anyone who opts in to the native inserter today.

Landing the remaining photo picker functionality will take some time, and isn't on the critical path to shipping the editor. It can land in a future release once it's had more time to bake.

## How?

- `EditorConfiguration`: new `enableInserterMediaStrip: Boolean = false` field with matching `Builder.setEnableInserterMediaStrip(...)`, `toBuilder()` round-trip, equals/hashCode coverage.
- `BlockPickerDialog`: takes a `showMediaStrip: Boolean` constructor arg and threads it through `BlockPickerSheet` → `SheetContent`, where the `MediaStrip()` call is wrapped in `if (showMediaStrip)`.
- `GutenbergView.presentBlockInserter`: passes `configuration.enableInserterMediaStrip` into the dialog.
- Demo app: adds a "Enable Inserter Media Strip" toggle in `SitePreparationActivity`'s feature card (disabled unless "Enable Native Inserter" is on) and wires it through `SitePreparationViewModel.buildConfiguration`.
- Tests: extends `EditorConfigurationBuilderTest` with default, setter, `toBuilder` round-trip, equality, and full-config assertions for the new field.

The flag intentionally stays Kotlin-side — the MediaStrip is a native Compose composable, so there's no need to propagate it through `GBKitGlobal` to the JS editor.

## Testing Instructions

1. Build the demo app and pick a configured site (or the bundled offline editor).
2. On the site preparation screen, ensure **Enable Native Inserter** is on. **Enable Inserter Media Strip** should default to off.
3. Tap **Prepare Editor**, open the block inserter — the Photos + Camera strip should not appear above the category tabs.
4. Return to the preparation screen, flip **Enable Inserter Media Strip** on, re-enter the editor, open the block inserter — the strip should reappear.
5. Run unit tests: `make test-android`.

No new UI surfaces. When the strip is hidden, TalkBack focus moves from the header's Close button straight to the first category tab; when it is shown, focus order is unchanged from the existing implementation.


adalpari (approved):
LGTM and works as expected!

PR #524, opened 2026-05-28
fix(android): derive inserter tabs from payload sections

### LINKED ISSUE #523: Native inserter tabs hide plugin and several core blocks via a hardcoded ID allowlist
## Summary

`BlockPickerTab` in `android/Gutenberg/src/main/java/org/wordpress/gutenberg/inserter/BlockPickerDialog.kt` carries a hardcoded set of block IDs per tab (paragraph/heading/list/… for Text, table/separator/code/… for Design, etc.) and uses it to filter what each tab shows. Anything not in the curated sets — every plugin block, plus a handful of core blocks that didn't make the design handoff — disappears from the sheet entirely. The comment above the enum acknowledges the taxonomy is a stub:

> Hardcoded block-id groupings from the design handoff. Production will likely derive these from the payload, but the design freezes the taxonomy shown in the tabs so we pin it here to match Variation B exactly.

Now that consumers are starting to enable plugins in production (we just turned on `setPlugins(true)` in wordpress-mobile/WordPress-Android), this gate is the visible reason Jetpack blocks never appear in the inserter despite being correctly registered, returned by `getInserterItems()`, and serialized into the payload sent to native.

## Repro

1. Build any consumer with `setPlugins(true)` and `setEnableNativeBlockInserter(true)` on a site that has Jetpack AI provisioned.
2. Open the editor, tap **+**.
3. In DevTools attached to the WebView, confirm the block is registered and routes through to the payload:

   ```js
   wp.data.select('core/block-editor').getInserterItems()
     .filter(i => i.name === 'jetpack/ai-assistant').length === 1
   ```

4. Observe: `jetpack/ai-assistant` is absent from every tab (Text, Media, Design, Widgets, Theme, Embeds). Recent doesn't contain it either since it's never been used. The block is functionally unreachable from the inserter.

The same gate also hides core blocks not in the curated sets — `core/code` is in Design, `core/columns` is in Design and Theme, but `core/group`, `core/details`, `core/footnotes`, `core/social-links`, etc. appear nowhere.

Confusingly, `core/table` is listed for ordering in `BLOCK_ORDER_BY_CATEGORY.text` on the JS side (`src/utils/blocks.js`) but core moved Table to `category: "design"` a while back, so the JS ordering hint is dead code and Table only surfaces via the native Design tab's allowlist. (Minor staleness, not the issue here.)

## Root Cause

`BlockPickerDialog.kt:920-927`:

```kotlin
private fun filterByTab(
    tab: BlockPickerTab,
    allBlocks: List<BlockType>,
    recentBlocks: List<BlockType>,
): List<BlockType> {
    if (tab == BlockPickerTab.Recent) return recentBlocks
    val ids = tab.blockIds ?: return allBlocks
    return allBlocks.filter { it.name.substringAfterLast('/') in ids }
}
```

`it.name.substringAfterLast('/')` strips the namespace, so `jetpack/ai-assistant` becomes `ai-assistant`. Each tab's `blockIds` is a hand-curated `Set<String>` of these stripped IDs. Anything outside the union of those sets is unreachable.

## Suggested Fix

Two options, not mutually exclusive:

1. **Drive tabs from the payload sections.** `BlockInserterPayload` already carries `sections: List<BlockInserterSection>` with category slugs and localized names. The Kotlin side could render one tab per non-synthetic section, falling back to a generic display for unknown categories. This matches what the comment above the enum already says the production behavior should be.

2. **Add a catch-all "More" tab.** Whichever direction tab taxonomy goes long-term, an "Everything else" tab guarantees the long tail of plugin and core blocks is at least reachable. Cheap to add and prevents this class of regression for future block additions.

A long-term cleanup: drop `BLOCK_ORDER_BY_CATEGORY.text`'s stale `core/table` reference in `src/utils/blocks.js` to keep the JS-side ordering hints honest. Doesn't affect anything functionally — `orderBlocksInCategory` silently no-ops when the listed block isn't in the bucket — but it's misleading to read.

## Affected

- `org.wordpress.gutenbergkit:android:v0.17.0` and later (introduced when the native inserter UI shell landed).

jkmassel:
## What?

- Replaced `BlockPickerTab`'s hardcoded block-ID allowlist with tabs derived from `BlockInserterPayload.sections`.
- Tab labels come from each section's `name` (already localized by the JS bridge via `getBlockCategories()`); `Recent` keeps its hardcoded string resource.

Fixes #523.

## Why?

`BlockPickerTab` carried a hand-curated `Set<String>` of block IDs per tab and filtered the inserter to that allowlist:

```kotlin
val ids = tab.blockIds ?: return allBlocks
return allBlocks.filter { it.name.substringAfterLast('/') in ids }
```

Anything outside the union of those sets disappeared — every plugin block, plus core blocks (`core/group`, `core/details`, `core/footnotes`, `core/social-links`, …) that didn't make the design handoff. Now that consumers are starting to enable plugins in production (e.g. `setPlugins(true)` in wordpress-mobile/WordPress-Android), this gate is the visible reason Jetpack blocks never appeared in the inserter despite being correctly registered, returned by `getInserterItems()`, and serialized into the payload. The comment above the enum already acknowledged the allowlist was a stub:

> Hardcoded block-id groupings from the design handoff. Production will likely derive these from the payload, but the design freezes the taxonomy shown in the tabs so we pin it here to match Variation B exactly.

This PR makes production match what the comment described.

## How?

The JS bridge already ships exactly what we need: `payload.sections` carries one section per WordPress block category, each with a localized `name` and the blocks in that category (core categories first, plugin categories appended). The Kotlin side just had to use it.

### 1. New `BlockPickerTabs.kt`

Pure helpers, no Compose, fully unit-testable:

- `BlockPickerTab` is now a sealed class — `Recent` (special-cased) and `Category(section)`.
- **`browsableSections(payload)`**: strips every section whose category starts with `gbk-` (the prefix the JS bridge uses for synthetic sections — `gbk-most-used`, `gbk-contextual`, `gbk-search-only`). Forward-compatible with any new synthetic section.
- **`buildTabs(sections)`**: `Recent` followed by one `Category` per browsable section. Payload order is preserved, so the JS side controls the tab taxonomy.
- **`blocksForTab(tab, recentBlocks)`**: recent for `Recent`, the section's blocks for `Category`. No allowlist, no `substringAfterLast('/')` namespace stripping.
- `recentBlocks` and `dedupeById` moved here unchanged so they're testable alongside the rest.

### 2. `BlockPickerDialog.kt`

- Removed the `BlockPickerTab` enum, `filterByTab`, `flattenBlocks`, `dedupeById`, and the `SEARCH_ONLY_CATEGORY` / `MOST_USED_CATEGORY` private constants.
- `BlockPickerSheet` derives the tab list from the payload and threads it through `SheetContent` → `CategoryTabs`.
- `tabLabel` renders `Recent` from a string resource and `Category` labels from `section.name` (defensive fallback to `section.category`, though the JS bridge always populates `name` for non-synthetic sections).

### 3. `strings.xml`

Removed unused per-tab labels (`gbk_block_inserter_tab_text`, `…_media`, `…_design`, `…_widgets`, `…_theme`, `…_embeds`). Tab labels come from the payload now. `gbk_block_inserter_tab_recent` stays.

## What We Explored

- **A catch-all "More" tab** (option 2 in #523) — option 1 makes it unnecessary. The JS bridge buckets every block into a category section (blocks with no `category` fall back to `'common'`, unknown categories get appended with `name = capitalize(slug)`), so every browsable block lives in some non-synthetic section. No overflow to catch.
- **Dropping the stale `core/table` entry** from `BLOCK_ORDER_BY_CATEGORY.text` in `src/utils/blocks.js` — was bundled in earlier, then split out. Tracked as a separate cleanup; the Android fix stands alone without it.
- **Surfacing `gbk-search-only` blocks in search** — pre-existing behavior outside the scope of this fix.

## Testing Instructions

Run the Android demo app against `jetpack.wpmt.co`, open the inserter, and confirm the AI block is reachable.

## Related issues

- Fixes #523

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/inserter/BlockPickerDialog.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/inserter/BlockPickerDialog.kt
@@ -105,8 +105,6 @@ import androidx.compose.ui.graphics.Color as ComposeColor
 import org.wordpress.gutenberg.model.BlockInserterPayload
 import org.wordpress.gutenberg.model.BlockType
 
-private const val SEARCH_ONLY_CATEGORY = "gbk-search-only"
-private const val MOST_USED_CATEGORY = "gbk-most-used"
 private const val SEARCH_DEBOUNCE_MS = 150L
 
 /** WordPress brand navy — seed for the fallback scheme on pre-API-31 devices. */
@@ -237,10 +235,12 @@ private fun BlockPickerSheet(
 ) {
     val colorScheme = rememberColorScheme()
     MaterialTheme(colorScheme = colorScheme) {
-        val allBlocks = remember(payload) { flattenBlocks(payload) }
+        val sections = remember(payload) { browsableSections(payload) }
+        val tabs = remember(sections) { buildTabs(sections) }
+        val allBlocks = remember(sections) { allBrowsableBlocks(sections) }
         val recentBlocks = remember(payload, allBlocks) { recentBlocks(payload, allBlocks) }
 
-        var selectedTab by remember { mutableStateOf(BlockPickerTab.Recent) }
+        var selectedTab by remember { mutableStateOf<BlockPickerTab>(BlockPickerTab.Recent) }
         var query by remember { mutableStateOf("") }
         var debounced by remember { mutableStateOf("") }
         LaunchedEffect(query) {
@@ -248,12 +248,13 @@ private fun BlockPickerSheet(
             debounced = query.trim()
         }
 
-        val filtered = remember(selectedTab, debounced, allBlocks, recentBlocks) {
-            val tabBlocks = filterByTab(selectedTab, allBlocks, recentBlocks)
+        val filtered = remember(selectedTab, debounced, recentBlocks) {
+            val tabBlocks = blocksForTab(selectedTab, recentBlocks)
             if (debounced.isEmpty()) tabBlocks else searchBlocks(debounced, tabBlocks)
         }
 
         SheetContent(
+            tabs = tabs,
             selectedTab = selectedTab,
             onSelectTab = { selectedTab = it },
             query = query,
@@ -271,6 +272,7 @@ private fun BlockPickerSheet(
 @Composable
 @Suppress("LongParameterList")
 private fun SheetContent(
+    tabs: List<BlockPickerTab>,
     selectedTab: BlockPickerTab,
     onSelectTab: (BlockPickerTab) -> Unit,
     query: String,
@@ -299,7 +301,7 @@ private fun SheetContent(
         if (showMediaStrip) {
             MediaStrip()
         }
-        CategoryTabs(selected = selectedTab, onSelect = onSelectTab)
+        CategoryTabs(tabs = tabs, selected = selectedTab, onSelect = onSelectTab)
         SearchField(query = query, onQueryChange = onQueryChange)
         BlockGridContent(
             blocks = blocks,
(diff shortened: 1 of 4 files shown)


nbradbury:
@jkmassel I'll approve this, but wanted to share what Claude had to say about it in case there's anything you want to address in a separate PR.

[review-pr-524-inserter-tabs-2026-05-28.pdf](https://github.com/user-attachments/files/28361186/review-pr-524-inserter-tabs-2026-05-28.pdf)

nbradbury (approved):
:shipit:

PR #533, opened 2026-07-01
[Tooling] Secure Claude workflows

iangmaia:
## Summary
- Allow `@claude` review comments only from trusted commenters and skip Claude workflows on fork/external PRs before checkout or Claude execution.
- Pin mutable action references and preserve the Claude action's required OIDC permission for GitHub App token exchange.
- Keep review feedback inline-only where the workflow does not grant issue-comment permissions.

PR #540, opened 2026-07-06
fix(e2e): stop wp-env warm-up JSON parse flakes

mokagio:
I run into E2E failures from time to time in this repo over the past few weeks. The last time this happened, in [the build](https://buildkite.com/automattic/gutenbergkit/builds/2401/canvas) for PR #533 which I was reviewing, I decided to task an agent to investigate with `/goal E2E green consistently`. This is the result. 

The approach makes sense to me. While one should avoid polling in tests, I think it's acceptable in the context of E2E . But, I'm not familiar withe the Playground / `wp-env` system that's in use for the tests, so I can't really judge.

I'm sharing this in the hope it helps getting the E2E back to stable, even if just as a starting point.

AI-generated details below.

---

## What?

Makes the Web E2E job resilient to wp-env warm-up hiccups, which were failing builds intermittently (e.g. [build 2401](https://buildkite.com/automattic/gutenbergkit/builds/2401)).

## Why?

Every failure in the flaky runs traced to `wp-env-fixtures.js`, where `response.json()` threw on a non-JSON body returned by the editor-settings/assets REST endpoints during warm-up — either a PHP notice printed into the response (`<br /><b>Warning...`) or a transient `500`. Editor settings are cached in module scope, but only on success, so a single early hiccup poisoned the first fetch and failed every early spec while later ones passed — the classic flaky signature (7 failed / 41 passed with all failures in `editor-load`, `editor-title`, `editor-error`).

## How?

Two complementary layers:

- `.wp-env.json`: set `WP_DEBUG_DISPLAY: false` so PHP notices never leak into REST bodies. `WP_DEBUG_LOG` stays on, so they still land in `debug.log`.
- `e2e/wp-env-fixtures.js`: route both fetches through a `fetchJson` helper that reads the body as text, retries transient `5xx`/parse failures, and includes a body snippet in the final error instead of an opaque `Unexpected token '<'`.

## Testing Instructions

This is a draft to exercise CI — the goal is a green (and repeatably green) Web E2E job.

Locally: `make wp-env-start` then `make test-e2e`. Before this change the settings-fetch failures reproduce on a cold instance; after, the suite passes (verified: 47/49 passed, the 2 remaining being local-only media-upload timeouts under `fullyParallel` workers, which pass at CI's `workers: 1`).

### DIFF
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -9,6 +9,7 @@
 	"config": {
 		"WP_DEBUG": true,
 		"WP_DEBUG_LOG": true,
+		"WP_DEBUG_DISPLAY": false,
 		"WP_ENVIRONMENT_TYPE": "local"
 	}
 }
--- a/bin/wp-env-setup.sh
+++ b/bin/wp-env-setup.sh
@@ -169,3 +169,40 @@ cat > "$CREDENTIALS_FILE" <<EOF
 EOF
 
 echo "Credentials written to $CREDENTIALS_FILE"
+
+# ---------------------------------------------------------------------------
+# Wait until WordPress is fully provisioned
+#
+# The Playground runtime keeps installing plugins (Blueprint steps) after it
+# starts accepting requests. During those installs WordPress drops a
+# `.maintenance` file and serves a 503, and there is a brief race where the
+# file is removed mid-request and PHP fatals. Both corrupt the first editor
+# REST call, so we gate here until the settings endpoint returns valid JSON.
+# ---------------------------------------------------------------------------
+
+echo "Waiting for the editor settings endpoint to be ready..."
+
+SETTINGS_URL="$SITE_URL/?rest_route=/wp-block-editor/v1/settings"
+READY_MAX_RETRIES=45
+READY_RETRY_INTERVAL=2
+
+for attempt in $(seq 1 $READY_MAX_RETRIES); do
+    if curl -s -H "Authorization: $AUTH_HEADER" "$SETTINGS_URL" | node -e "
+        let data = '';
+        process.stdin.on('data', chunk => data += chunk);
+        process.stdin.on('end', () => {
+            try { JSON.parse(data); }
+            catch { process.exit(1); }
+        });
+    "; then
+        echo "Editor settings endpoint is ready."
+        break
+    fi
+
+    if [ "$attempt" -eq "$READY_MAX_RETRIES" ]; then
+        echo "Error: editor settings endpoint did not return valid JSON after $((READY_MAX_RETRIES * READY_RETRY_INTERVAL)) seconds."
+        exit 1
+    fi
+
+    sleep $READY_RETRY_INTERVAL
+done
--- a/e2e/fetch-json.js
+++ b/e2e/fetch-json.js
@@ -1,0 +1,77 @@
+const MAX_FETCH_ATTEMPTS = 15;
+const FETCH_RETRY_DELAY_MS = 2000;
+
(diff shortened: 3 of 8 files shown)


iangmaia:
@claude review this PR.

iangmaia:
> **Claude finished @iangmaia's task in 0s** —— [View job](https://github.com/wordpress-mobile/GutenbergKit/actions/runs/28950483524)
> 
> I'll analyze this and get back to you.

FYI https://github.com/wordpress-mobile/GutenbergKit/pull/533 isn't merged so ofc this won't trigger the new Claude workflow.

mokagio:
@claude review this PR, now that #533 landed and the workflow is secured.

mokagio:
@iangmaia looks like the Claude review run but never went past reporting that it started? I had a quick look at the logs, https://github.com/wordpress-mobile/GutenbergKit/actions/runs/28994702797/job/86041799638, but couldn't make much of it. Haven't passed it to my AI for further investigation, but  can do that later if you don't have the bandwidth.

iangmaia:
> @iangmaia looks like the Claude review run but never went past reporting that it started? I had a quick look at the logs, https://github.com/wordpress-mobile/GutenbergKit/actions/runs/28994702797/job/86041799638, but couldn't make much of it. Haven't passed it to my AI for further investigation, but can do that later if you don't have the bandwidth.

It seems to be a problem with the token, which doesn't surface properly. See https://github.com/anthropics/claude-code-action/pull/1249.
I'll create a new token and update it in the action secrets.

iangmaia:
@claude review this PR.

mokagio:
Thanks for the review — addressed in `eeb6fc8` and `ec128f8`.

**🟡 `fetchJson` retrying every error** — fixed. A `NonRetryableError` marker now short-circuits the loop: a 4xx other than 408/429 fails immediately with the labelled message, while 5xx, network-level rejections, and JSON parse failures still retry. That matches what the JSDoc claimed and drops a bad `authHeader` or wrong route from a 30s hang to instant.

**The three 🟢 items** — left as-is, in line with your own "no change needed" framing:

- The readiness-gate overlap is deliberate, and your point stands that `bin/wp-env-setup.sh` only gates `wp-block-editor/v1/settings`, so the in-test retry still carries the weight for `wpcom/v2/editor-assets`.
- The duplicated Node JSON-parse snippet is still only two occurrences.
- The `lastError.message` invariant is now moot: the non-retryable path throws directly rather than falling through to that line.

On the body-snippet-in-logs note — keeping it. These fixtures point at a local wp-env, and that snippet is the entire point of the original change.

**Beyond the review:** retry/fast-fail branching is exactly the logic that regresses silently, so it now has 9 Vitest cases (success, 5xx retry, PHP-notice-in-body retry, network rejection, 401/404 fast-fail, 408/429 retry, budget exhaustion). Two things fell out of that:

- `e2e/**` was excluded from Vitest wholesale, and Playwright's default `testMatch` also claims `*.test.js`, so both runners would have fought over the file. Each is now narrowed to `*.spec.js`. Playwright still collects the same 49 tests in 20 files.
- `wp-env-fixtures.js` runs `readCredentials()` at module scope, so importing it anywhere without a provisioned wp-env throws — which broke the JS test job on the first push. `fetchJson` now lives in `e2e/fetch-json.js`, free of that side effect. Verified by running the suite with `.wp-env.credentials.json` removed, reproducing CI's environment rather than trusting mine.

Build is green, Web E2E included.

---

*Posted by Claude Code (Opus 4.8) on behalf of @mokagio with approval.*

iangmaia (approved):

PR #546, opened 2026-07-10
fix: native media upload review follow-ups (relay, delegate contract, CORS)

jkmassel:
Follow-ups from reviewing the native media upload pipeline in #357, plus a code-review hardening pass on those follow-ups. Stacked on #357; no new feature surface.

The load-bearing fix: **the editor now gets WordPress's upload response verbatim.** #357 reshaped that response into a hand-built subset, which dropped `media_details.sizes` — so every native-uploaded image fell back to full resolution in the editor. The proxy now changes only the bytes that get uploaded, not the response contract, so every consumer — image sub-size selection, gallery links, error notices — behaves exactly as a non-proxied upload.

## Response fidelity

-   Relay WordPress's raw status and body verbatim rather than a synthesized subset. A failed upload now carries WordPress's own error message and code instead of a local 500, so the editor's failure notice matches a direct upload.
-   Forward the whole original request through the local server — the file plus `post`, any additional form fields, and the `?_embed` query — rather than the file alone. Non-file fields are relayed as raw bytes through the re-encode, so a non-UTF-8 value can't be silently coerced.

## Delegate contract

-   `processFile` returns a `ProcessedProxyFile` (`.original`, or `.processed(url, mimeType:, filename:)`) instead of a bare path. Passthrough vs. processed was previously inferred by path equality, which conflated "unchanged" with "edited in place" and left the delegate no way to report a corrected type or name — so a format change (HEIC→JPEG, MOV→MP4) uploaded under the original extension. The transformed file's type and name are now explicit.

## Server lifecycle & reachability

-   **iOS:** stop the server in `deinit` rather than `viewDidDisappear`, which was tearing it down whenever the editor was merely backgrounded or covered; hold the delegate weakly to break a retain cycle.
-   **Android:** re-advertise the port/token to the WebView on restart, and **clear them on stop** so the JS middleware routes to the default upload path instead of fetching a now-dead port. Start only when a REST uploader can be built, and **skip start when cleartext to `localhost` is blocked** (`NetworkSecurityPolicy`) so a host that hasn't allow-listed it degrades gracefully instead of failing every upload. Sweep crash-orphaned temp files off the caller's thread, and post the port/token sync to the WebView's UI thread.
-   **JS:** the middleware guard is the single reachability decision — with no advertised port it passes straight through to the default upload. The previous connection-error **fallback is removed**: it re-ran a non-idempotent `POST /wp/v2/media` (risking a duplicate attachment) and pointed the wrong way under Lockdown Mode, where the loopback proxy is the path that works. A user-initiated cancel is still propagated — keyed off `signal.aborted` (so `AbortSignal.timeout()` is honored, not just `AbortError`), rethrowing the signal's canonical reason.

## Timeouts

-   The upload client uses per-operation **inactivity** timeouts matching URLSession's defaults — 15s connect, 60s read/write — rather than a total `callTimeout`. A large upload that keeps making progress is never capped by total duration; only a genuine 60s stall in the active direction fails.

## CORS

-   Move CORS into the HTTP library as an opt-in `permissive` policy. The library answers `OPTIONS` and stamps CORS headers on every response — including the ones it generates itself (read timeout, parse error) that never reach the upload handler and previously surfaced in the editor as an opaque `Failed to fetch`. Each server opts in instead of hand-maintaining CORS headers per response.

## Parse-error classification

-   Classify each HTTP parse error as `fatal` or `recoverable` through an explicit `Disposition`, pinned by a test: only an over-limit payload (413) is recoverable, so a smuggling-relevant malformed request can't silently be reclassified to reach the handler ahead of auth.

## Cleanup

-   Extract the site-root/namespace URL normalization into a shared, tested helper (`WordPressRESTURL` on iOS, `RestUrlBuilder` on Android) so the media endpoint stops being the one path that bypasses `RESTAPIRepository`'s canonical builder and can't drift from it. On iOS the request query is now merged via `URLComponents.percentEncodedQuery`, so a value that isn't URL-safe can't drop `?_embed`.
-   Remove upload temp files on failure, not only on success, and sweep crash-orphaned temps on startup.
-   Report the demo delegate's **actual** re-encoded image type on Android — a resized WebP/HEIC becomes `image/jpeg`/`.jpg` rather than JPEG bytes mislabeled `image/webp`.
-   Bake EXIF orientation into the demo apps' resized images so a portrait photo doesn't upload rotated.
-   Delete dead iOS `MediaUploadError` cases and an unused `Data.append` extension; regenerate `Package.resolved` against the committed manifest.

## Test plan

Automated (all green):

-   [x] **JS** — `api-fetch-upload-middleware.test.js`: interception, verbatim relay, WordPress-error surfacing, `invalid_json` normalization on a non-JSON body (both 2xx and non-2xx), abort/timeout propagation without fallback, and body + query forwarding.
-   [x] **iOS** — `swift test`: `GutenbergKitHTTP` (request-parsing fixtures + the fatal/recoverable disposition lock), `WordPressRESTURL`, and `MediaUploadServer` (verbatim relay incl. a real non-2xx path, processed-metadata forwarding, processed-file cleanup, the orphan-sweep age threshold, the weak-delegate lifetime that keeps `deinit` teardown working, and the 413 + CORS path).
-   [x] **Android** — `./gradlew :Gutenberg:testDebugUnitTest` and `detekt` green: `RestUrlBuilder`, `MediaUploadServer` (relay / metadata / URL normalization / non-UTF-8 field bytes / orphan-sweep), the shared HTTP fixtures, and the disposition lock.

Manual:

-   [ ] With the demo app's "Enable Native Media Upload" on, insert an Image block and upload a >2000px photo into an **existing** post: it displays at the `large` size (not full resolution), attaches to the post, and logs the resize.
-   [ ] Upload a portrait photo shot in landscape-with-orientation — it appears upright, not rotated.
-   [ ] Force a WordPress rejection (e.g. a disallowed file type) — the editor shows WordPress's own error message, not a generic failure.
-   [ ] Background the app while the editor is open, return to it, and upload — the upload still succeeds (the iOS `deinit` change keeps the server alive across backgrounding).

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/Media/MediaUploadDelegate.swift
+++ b/ios/Sources/GutenbergKit/Sources/Media/MediaUploadDelegate.swift
@@ -21,6 +21,19 @@ public struct MediaUploadResponse: Sendable {
     }
 }
 
+/// The result of a delegate's ``MediaUploadDelegate/processFile(at:mimeType:filename:)``.
+public enum ProcessedProxyFile: Sendable {
+    /// The delegate did not modify the file; the original upload is forwarded
+    /// to WordPress unchanged.
+    case original
+
+    /// The delegate produced a file to upload, along with its MIME type and
+    /// filename. Both are used verbatim, so a format change (e.g. transcoding
+    /// MOV to MP4, or an in-place EXIF strip) must report the resulting type and
+    /// filename for WordPress to store the file correctly.
+    case processed(URL, mimeType: String, filename: String)
+}
+
 /// Protocol for customizing media upload behavior.
 ///
 /// The native host app can provide an implementation to resize images,
@@ -28,8 +41,12 @@ public struct MediaUploadResponse: Sendable {
 /// pass files through unchanged and upload via the WordPress REST API.
 public protocol MediaUploadDelegate: AnyObject, Sendable {
     /// Process a file before upload (e.g., resize image, transcode video).
-    /// Return the URL of the processed file, or the original URL for passthrough.
-    func processFile(at url: URL, mimeType: String) async throws -> URL
+    ///
+    /// Return ``ProcessedProxyFile/original`` to upload the file unchanged, or
+    /// ``ProcessedProxyFile/processed(_:mimeType:filename:)`` with the processed
+    /// file and its metadata. When the format changes, report the new mimeType
+    /// and filename so WordPress stores it with the correct extension and type.
+    func processFile(at url: URL, mimeType: String, filename: String) async throws -> ProcessedProxyFile
 
     /// Upload a processed file to the remote WordPress site.
     ///
@@ -42,8 +59,8 @@ public protocol MediaUploadDelegate: AnyObject, Sendable {
 
 /// Default implementations.
 extension MediaUploadDelegate {
-    public func processFile(at url: URL, mimeType: String) async throws -> URL {
-        url
+    public func processFile(at url: URL, mimeType: String, filename: String) async throws -> ProcessedProxyFile {
+        .original
     }
 
     public func uploadFile(at url: URL, mimeType: String, filename: String) async throws -> MediaUploadResponse? {
--- a/ios/Sources/GutenbergKit/Sources/Media/MediaUploadDelegate.swift
+++ b/ios/Sources/GutenbergKit/Sources/Media/MediaUploadDelegate.swift
@@ -1,29 +1,23 @@
 import Foundation
 
-/// Result of a successful media upload to the remote WordPress server.
+/// A raw response from the WordPress REST API media endpoint.
 ///
-/// This structure matches the format expected by Gutenberg's `onFileChange` callback.
-public struct MediaUploadResult: Codable, Sendable {
-    public let id: Int
-    public let url: String
-    public let alt: String
-    public let caption: String
-    public let title: String
-    public let mime: String
-    public let type: String
-    public let width: Int?
-    public let height: Int?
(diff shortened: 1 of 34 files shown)


ios/Sources/GutenbergKit/Sources/Media/MediaUploadDelegate.swift
```diff
@@ -1,54 +1,69 @@
...
+public enum ProcessedProxyFile: Sendable {
+    /// The delegate did not modify the file; the original upload is forwarded
+    /// to WordPress unchanged.
+    case original
+
+    /// The delegate produced a file to upload, along with its MIME type and
+    /// filename. Both are used verbatim, so a format change (e.g. transcoding
+    /// MOV to MP4, or an in-place EXIF strip) must report the resulting type and
+    /// filename for WordPress to store the file correctly.
+    case processed(URL, mimeType: String, filename: String)
```
dcalhoun, line 34: From review with Claude, my perception is that the server takes ownership over these processed files and deletes them after uploading. Given that isn't documented, it may be confusing for a host that intends to retain its processed file that it stores in a chosen location for some other purpose. 

Is it worth documenting in the iOS and Android (`MediaUploadServer.kt:53`) comments that this ownership occurs?

dcalhoun (approved):
These changes make sense to me. I appreciate the improvements and increased robustness. Thanks! 

This also tested well for me. I'm going to merge this into the target branch so that we integrate the latest `trunk` branch and perform final testing. 

My one inline suggestion is relatively low priority that we could address later.

PR #555, opened 2026-07-18
chore: add app icon to iOS and Android demo apps

dcalhoun:
## What?

Adds a proper app icon to the iOS and Android demo apps, replacing the default placeholders (blank/generated on iOS, the Android Studio robot on Android). Both platforms now use the Gutenberg "G" glyph on a white background. The iOS home-screen label is also updated to "GutenbergKit" to match Android.

## Why?

The demo apps shipped with placeholder icons, making them hard to distinguish on the home screen/launcher and inconsistent between platforms. A shared "G" mark gives them a recognizable, consistent identity.

## How?

- **iOS**: Added a 1024×1024 `AppIcon.png` (single-size universal format already configured in the asset catalog) rendered from the Gutenberg "G" vector on an opaque white background. Set `INFOPLIST_KEY_CFBundleDisplayName` to `GutenbergKit` in both Debug/Release configs so the home-screen name matches Android (the product/binary name stays `Gutenberg`).
- **Android**: Replaced the adaptive icon — the foreground is a vector drawable of the "G", the background is solid white — and regenerated the legacy `ic_launcher`/`ic_launcher_round` WebP mipmaps for all densities (mdpi–xxxhdpi).
- **Visual centering**: The "G" has a curved tail extending right, which pulls its geometric bounding-box center rightward. Both icons center the glyph on its _visual_ mass (the bowl) instead, so it reads as centered. The Android glyph is additionally sized to leave padding comparable to neighboring launcher icons.

## Testing Instructions

1. Build and run the iOS demo app; confirm the "G" icon and "GutenbergKit" label appear on the home screen.
2. Build and run the Android demo app; confirm the "G" launcher icon appears (adaptive shape adapts to the launcher mask) with padding comparable to other icons, and the label reads "GutenbergKit".

## Screenshots or screencast

| iOS | Android |
| --- | ------- |
|  ios-icon  | android-icon |


adalpari (approved):
🚢 it!!

PR #557, opened 2026-07-21
fix: route the upload server on the path, not the full target

dcalhoun:
## What?

Media uploads through the native upload server fail with a 404.

## Why?

Uploads that route through the native server never reach WordPress — the editor surfaces an upload failure instead of inserting the image.

`@wordpress/media-utils` uploads to `/wp/v2/media?_embed=wp:featuredmedia`, and [#546](https://github.com/wordpress-mobile/GutenbergKit/pull/546) started forwarding that query on to the native server so it could be relayed to WordPress. The route guard still compared the _full_ request target against `"/upload"`, so the appended query string made the match fail and the server returned 404 before the upload handler ever ran:

```
POST /upload?_embed=wp:featuredmedia → 404
```

Both platforms had the bug. Notably, each handler already expected a query — both sliced it off the target to relay upstream — so only the guards were out of step.

## How?

Adds `path` and `query` accessors to `ParsedHTTPRequest` (iOS) and `HttpRequest` (Android), then routes on `path` and takes the relayed query from `query` instead of re-deriving it in each handler.

A bare trailing `?` carries no parameters, so all three sides normalize it to an empty query and the value can be appended to an upstream URL unconditionally. The middleware that builds the target derived its query by hand and kept the `?`, sending `POST /upload?` — a target the native accessors treat as impossible. It now shares the same rule via `requestQuery`.

## Testing Instructions

1. Run the demo app against a site, so uploads route through the native upload server.
2. Insert an Image block and pick a photo from the library.
3. Confirm the image uploads and renders in the editor.
4. In the Xcode/Android console, confirm the request is `POST /upload?_embed=wp:featuredmedia` and returns `201`, not `404`.

Automated coverage:

```
make test-swift-library
make test-android
npx vitest run src/utils/api-fetch-upload-middleware.test.js
```

Both native suites include a regression test that uploads through `/upload?_embed=wp:featuredmedia` and asserts the query reaches the uploader unchanged. Reverting either route guard to match on `target` makes them fail with the original 404. Each also pins which upload branch ran, since the mock records the query on both — without that, a regression collapsing routing onto one branch would still pass.

The middleware suite covers the bare-`?` normalization; restoring the hand-rolled `indexOf`/`slice` makes it fail.

Not applicable — no user interface changes.

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/HttpServer.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/HttpServer.kt
@@ -46,6 +46,23 @@ data class HttpRequest(
     val serverError: org.wordpress.gutenberg.http.HTTPRequestParseError? = null
 ) {
     /**
+     * The path portion of [target], without the query component
+     * (e.g., "/wp/v2/posts" for "/wp/v2/posts?per_page=10").
+     *
+     * Use this for routing — matching against [target] fails as soon as a
+     * client appends a query string.
+     */
+    val path: String
+        get() = target.substringBefore('?')
+
+    /**
+     * The query component of [target], including the leading "?"
+     * (e.g., "?per_page=10"), or an empty string when there is no query.
+     */
+    val query: String
+        get() = target.substringAfter('?', "").let { if (it.isEmpty()) "" else "?$it" }
+
+    /**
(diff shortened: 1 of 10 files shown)


dcalhoun:
Requested review for awareness, but also merging now given the target branch is broken, in-progress work currently under review.

PR #559, opened 2026-07-21
fix(demo-ios): reference GutenbergKit as a local Swift package

dcalhoun:
## What?

Stops the iOS demo app from rewriting the library's root `Package.resolved` every time the project is opened in Xcode.

## Why?

Opening `ios/Demo-iOS/Gutenberg.xcodeproj` added a `wordpress-rs` pin to the repo-root `Package.resolved`; resolving the Swift package on its own removed it again. The result was a recurring diff that had to be discarded by hand.

The root cause is in the demo project: it referenced the repo root as a plain folder (`PBXFileReference` with `path = ../..`) rather than as a package. Xcode treats a folder-referenced local package as an *editable root package*, so it wrote the demo app's merged dependency graph — including `wordpress-rs`, which is a demo-app dependency and not a `GutenbergKit` library dependency — into the library's own lockfile.

## How?

Declare the root as an `XCLocalSwiftPackageReference` and drop the folder reference. The demo app's graph now resolves into a demo-scoped lockfile under `project.xcworkspace/xcshareddata/swiftpm/`, leaving the library's `Package.resolved` untouched.

That demo lockfile is committed so the demo app and its E2E tests resolve reproducibly. It pins SwiftSoup 2.13.6 while the root pins 2.8.8 — expected, since these are now independent graphs and both satisfy `from: "2.7.5"`.

## Testing Instructions

1. Open `ios/Demo-iOS/Gutenberg.xcodeproj` in Xcode and let packages resolve.
2. Run `git status` — the root `Package.resolved` should be unmodified.
3. Confirm the demo app builds and runs.

To see the previous behavior, repeat step 1 on `trunk`: the root `Package.resolved` gains a `wordpress-rs` pin.

Verified locally with DerivedData and the demo lockfile deleted beforehand, so the result is not a warm-cache artifact:

- Pre-fix, resolving added 9 lines to the root `Package.resolved`.
- Post-fix, a cold resolve leaves it untouched.
- `xcodebuild build -scheme Gutenberg` succeeds, and the built app contains `GutenbergKit_GutenbergKitResources.bundle`, confirming the local package links.

### DIFF
--- a/ios/Demo-iOS/Gutenberg.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ios/Demo-iOS/Gutenberg.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,0 +1,33 @@
+{
+  "originHash" : "a8d9099ea14692b5a05469f935a234e28952c8e9499519d8e1f7800b52964342",
+  "pins" : [
+    {
+      "identity" : "svgview",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/exyte/SVGView.git",
+      "state" : {
+        "revision" : "6465962facdd25cb96eaebc35603afa2f15d2c0d",
+        "version" : "1.0.6"
+      }
+    },
+    {
+      "identity" : "swiftsoup",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/scinfu/SwiftSoup.git",
+      "state" : {
+        "revision" : "ead56133a693d0184d8c2db1a6d6394410cacfd6",
+        "version" : "2.13.6"
+      }
+    },
+    {
+      "identity" : "wordpress-rs",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/Automattic/wordpress-rs",
+      "state" : {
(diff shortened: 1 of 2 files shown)


ios/Demo-iOS/Gutenberg.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,33 @@
...
+        "version" : "1.0.6"
+      }
+    },
+    {
+      "identity" : "swiftsoup",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/scinfu/SwiftSoup.git",
+      "state" : {
+        "revision" : "ead56133a693d0184d8c2db1a6d6394410cacfd6",
+        "version" : "2.13.6"
```
dcalhoun, line 19: The below is referenced in the PR description. I don't believe this is an issue, but let me know if you do. 

> That demo lockfile is committed so the demo app and its E2E tests resolve reproducibly. It pins SwiftSoup 2.13.6 while the root pins 2.8.8 — expected, since these are now independent graphs and both satisfy `from: "2.7.5"`.

crazytonyli (approved):

PR #561, opened 2026-07-22
fix: harden the native media upload server (review of #357)

jkmassel:
Adversarial review of the native media upload server introduced in #357, with the findings fixed. Twelve changes across iOS, Android, and the JS middleware — timeout, lifecycle, concurrency, resource-leak, and injection bugs — each verified, and the behavior-changing ones mutation-tested (remove the guard → the new test fails → restore). The last two (`8d836f42`, `dfa7f855`) came from a second prosecutor/defender validation pass over the already-hardened code.

**Stacks on #357**: the base is `feat/leverage-host-media-processing`, so this merges into that PR's branch rather than trunk.

## Summary

**Correctness**

- **Large uploads were aborted mid-transfer** (`b0739f10`, both platforms) — the read timeout was a single 30s total-duration cap over the _whole_ request while the body limit is 4 GB, so a legitimate large upload that streamed steadily for more than 30s got a 408. Split the read: the pre-body phase (headers + oversized drain — the unauthenticated-reachable part) keeps the total cap; the accepted body is bounded by the per-read idle timeout plus a generous ceiling. Coupled fix: reject auth-exempt `OPTIONS` that carry a body, so an unauthenticated `OPTIONS` with an oversized `Content-Length` can't abuse the now idle-only body read.
- **iOS: `mediaUploadDelegate` set after load silently did nothing** (`cdb1683b`) — it's captured once into the page config at load. Kept the `weak` reference (making it strong reintroduces a deliberately-avoided retain cycle that would leak the server) and made both misuse paths trap: set-after-load, and a delegate deallocated before load.
- **iOS: concurrent editors wiped each other's in-flight upload buffers** (`8a08a084`) — the start-time temp-file sweep deleted _every_ file in a shared directory, and two editors (iPad multi-window, or one tearing down as another starts) share it. Added Android's `activeFiles` guard so the sweep skips live buffers while still reclaiming crash orphans.
- **Android: the upload server could be resurrected on a detached view** (`d7cddecb`) — a delegate assigned after `onDetachedFromWindow` started a server that nothing would ever stop. `@Volatile` the field (matching its sibling) plus a torn-down guard.
- **Android: a failed response-body read hung the upload coroutine forever** (`8d836f42`) — the `execute()`→`enqueue` refactor moved the body read inside OkHttp's `onResponse`, which OkHttp marks signalled _before_ invoking, so a throw there (a truncated/reset body after WordPress's `201`, or the read timeout firing mid-body) is swallowed — logged, never routed to `onFailure`. The continuation never resumed, so the coroutine hung holding a connection-semaphore permit; five exhaust `maxConnections` and the server stops accepting. Resume from a `catch (IOException)`, restoring the prompt failure the synchronous `execute().use{}` produced.
- **iOS: the upload client silently dropped the request-observing delegate** (`dfa7f855`) — `uploadClient()` built its sibling without the `EditorHTTPClientDelegate`, so a host wiring one to observe _all_ requests (as the protocol doc promises) saw zero `POST /wp/v2/media` uploads or passthroughs. Constrained the protocol to `Sendable` and carried the delegate over; only the REST timeout is still dropped. Source-compatible in-repo (the test spy is already `@unchecked Sendable`); external non-`Sendable` conformers would need the conformance.

**Resource leaks**

- **iOS: a stalled upload leaked the writer thread and its file descriptor** (`63d046a1`) — the bound-pair writer blocks forever if URLSession abandons the body stream; close the input on every exit so it unwinds.
- **Android: `stop()` didn't cancel an internally-created coroutine scope** (`70497ea3`).

**Security hardening**

- **iOS: multipart header injection** (`ebc9e802`) — client-supplied filename, field names, and MIME type were interpolated into `Content-Disposition`/`Content-Type` unescaped, so a `"` or CRLF could inject headers or an extra part into the request relayed to WordPress. Escape at the serialization boundary. Bounded by the trust model (the token holder already holds the WordPress credential), so this is defense-in-depth rather than privilege escalation.

**JS middleware**

- **A cancelled upload could `throw undefined`** (`b7b39a5e`) when the abort signal carried no `reason`; fall back to a canonical `AbortError`.
- **The cancel/timeout tests were tautological** (`4aa967ca`) — they rejected `fetch` with the same object they asserted on, so a regression that rethrew the network error would have passed. Rewritten to exercise the abort-vs-network race.

**Documented, not fixed**

- **iOS: the upload body is a one-shot stream** (`55867b64`) — a `307`/`308` redirect on the media `POST` (which WordPress core never emits) would resend an exhausted, empty body → a clean failure, not a corrupt attachment. Documented the limitation rather than adding `needNewBodyStream` plumbing for so rare a case.

## Determined non-issues

- **iOS `siteApiRoot` guard** — the finding assumed Android's empty-`String` check applies, but iOS's `siteApiRoot` is a non-optional `URL`, so there is no empty state to guard against.
- **iOS ATS / cleartext guard** — loopback isn't subject to iOS cleartext blocking (uploads work), and there's no runtime API to mirror Android's `NetworkSecurityPolicy` check. Android-specific by nature.

## Test plan

- [x] iOS `GutenbergKitHTTP` suite (386 tests) plus new timeout / OPTIONS-with-body / temp-file-cleanup / bound-stream-teardown tests
- [x] iOS `GutenbergKit` upload and multipart suites, including a new header-injection test
- [x] iOS **Simulator** build — `EditorViewController` is `#if canImport(UIKit)`, so host `swift build` compiles it empty; verified with `xcodebuild -destination 'generic/platform=iOS Simulator'`
- [x] Android `:gutenberg` unit tests and `detekt`, including new Robolectric detached-view-leak and owned-scope-cancel tests
- [x] JS suite (183 tests) and `eslint`
- [x] Mutation-tested the behavior-changing fixes (timeout split, header escaping, detached-view guard, `AbortError` fallback, scope cancel, cancel-test rewrite) — removed each guard, confirmed the new test fails, restored
- [x] Android: new `MediaUploadServer` truncated-response regression test — mutation-tested (`MockWebServer` `DISCONNECT_DURING_RESPONSE_BODY` + a `withTimeout` guard; without the fix the coroutine hangs and the test trips the timeout)
- [x] iOS: new `EditorHTTPClient` test asserting `uploadClient()` carries the delegate through to uploads

## Out of scope

- The WordPress `authHeader` persisted to WebView `localStorage` predates this feature (it's already in `origin/trunk`); this PR only adds a short-lived loopback token to that already-persisted blob, so it isn't addressed here.

### DIFF
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/HttpServer.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/HttpServer.kt
@@ -437,6 +437,8 @@ class HttpServer(
                 )
                 val response = try {
                     handler(request)
+                } catch (e: CancellationException) {
+                    throw e
                 } catch (e: Exception) {
                     Log.e(TAG, "Handler threw", e)
                     HttpResponse(
@@ -554,6 +556,11 @@ class HttpServer(
         }
         return try {
             handler(request)
+        } catch (e: CancellationException) {
+            // Never swallow cooperative cancellation (e.g. from stop()/detach):
+            // rethrow so handleConnection unwinds cleanly instead of writing a 500
+            // to a connection that's being torn down.
+            throw e
         } catch (e: Exception) {
             Log.e(TAG, "Handler threw", e)
             HttpResponse(status = 500, body = "Internal Server Error".toByteArray())
--- a/android/Gutenberg/src/main/java/org/wordpress/gutenberg/HttpServer.kt
+++ b/android/Gutenberg/src/main/java/org/wordpress/gutenberg/HttpServer.kt
@@ -23,10 +23,13 @@ import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.async
 import kotlinx.coroutines.cancel
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.selects.select
 
 /**
  * A received HTTP request.
@@ -506,9 +509,13 @@ class HttpServer(
                     body = parsed.body,
                     parseDurationMs = parseDurationMs
                 )
-                val response = resolveResponse(request)
-                sendResponse(socket, response)
-                Log.d(TAG, "${parsed.method} ${parsed.target} → ${response.status} (${"%.1f".format(parseDurationMs)}ms)")
+                val response = resolveResponseRacingClose(request, socket, input)
+                if (response != null) {
+                    sendResponse(socket, response)
+                    Log.d(TAG, "${parsed.method} ${parsed.target} → ${response.status} (${"%.1f".format(parseDurationMs)}ms)")
+                } else {
+                    Log.d(TAG, "${parsed.method} ${parsed.target} → client disconnected before response; cancelled in-flight handler")
+                }
             }
         }
     }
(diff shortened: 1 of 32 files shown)


ios/Sources/GutenbergKitHTTP/HTTPServer.swift
```diff
@@ -444,6 +504,28 @@ public final class HTTPServer: Sendable {
         connectionTasks.track(taskID, task)
```
dcalhoun, line 504: Capturing a note from Claude worth considering: 

> The task is created on line 348 but only registered here. If `stop()` runs in that window, `cancelAll()` iterates a map that doesn't yet contain this task and then clears it — `track()` then inserts an uncancelled task nobody will ever cancel.
> 
> This is an unstructured `Task`, so nothing else propagates the cancellation. Past the `try Task.checkCancellation()` on line 476, the handler runs the relay to completion and `POST /wp/v2/media` lands — the orphaned attachment the close-racing work exists to prevent, via a different door.
> 
> Android isn't exposed to this: `launch` makes the handler a child of `scope`, so `scope.cancel()` in `stop()` covers children that haven't dispatched yet. (`CoroutineStart.ATOMIC` there solves a different problem — a pre-dispatch cancel skipping the body and leaking the fd + permit.) The iOS registry has no parent/child relationship, so it needs the latch explicitly.
> 
> Suggest latching a stopped flag in `ConnectionTasks` (line 882, outside this diff):
> 
> ```swift
> final class ConnectionTasks: @unchecked Sendable {
>     private let lock = NSLock()
>     private var tasks: [UUID: Task<Void, Never>] = [:]
>     private var isStopped = false
> 
>     func track(_ id: UUID, _ task: Task<Void, Never>) {
>         let cancelImmediately: Bool = lock.withLock {
>             guard !isStopped else { return true }
>             tasks[id] = task
>             return false
>         }
>         // Registered after cancelAll() — the server is already stopping, so this
>         // task would otherwise never be cancelled.
>         if cancelImmediately { task.cancel() }
>     }
> 
>     func cancelAll() {
>         lock.withLock {
>             isStopped = true
>             for task in tasks.values { task.cancel() }
>             tasks.removeAll()
>         }
>     }
> }
> ```

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -417,7 +473,7 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         }
```
dcalhoun, line 473: Capturing a note from Claude worth considering: 

> (Re: the `authHeader` guard just above — line 471 isn't in the diff, so anchoring here.)
> 
> This guard omits `isOfflineModeEnabled`. `GBKitGlobal` nils `siteURL`/`siteApiRoot` in offline mode ("No site exists; skip all networking entirely"), but `DefaultMediaUploader` is built from `configuration.siteApiRoot` directly — so an offline host that still supplies an `authHeader` gets a live server that would relay to the real site.
> 
> Low impact in practice: `EditorService.prepare()` short-circuits to empty dependencies in offline mode, so there's little to upload against. Filing it for contract consistency rather than an observed failure.
> 
> ```swift
>         guard !configuration.isOfflineModeEnabled, !configuration.authHeader.isEmpty else {
>             return
>         }
> ```

android/Gutenberg/src/main/java/org/wordpress/gutenberg/HttpServer.kt
```diff
@@ -111,6 +111,17 @@ enum class CorsPolicy {
...
+                // deliberate, and safe here — not an oversight to tighten. The
+                // server is loopback-only, and every non-OPTIONS request is gated
+                // by a per-session random bearer token stored only in the editor
+                // origin's localStorage/window.GBKit, which is origin-scoped and
+                // unreadable by any other origin — so no cross-origin can obtain
+                // it. `*` only governs whether a *token-holding* origin may read
+                // the response, and the sole token-holder is the editor itself, the
+                // legitimate client. Echoing the origin isn't viable anyway: the
+                // editor loads from file:// (Origin null), which can't be cleanly
+                // allowlisted.
```
dcalhoun, line 124: The last clause is true on iOS but not here. Android sets a custom or default asset domain. Local files, but served from the site's own origin, not `file://`. So echoing the origin _is_ viable on Android; it's just unnecessary, since the token already gates every non-OPTIONS request.

The rest of the rationale holds unchanged. Worth correcting, though.

```suggestion
                // legitimate client. An origin allowlist would be viable here
                // (the editor document is served from the site's own origin —
                // see `assetDomain` in `GutenbergView`), but it would add
                // nothing the token gate doesn't already provide.
```

dcalhoun (approved):
Thanks for the effort put into this! 🙇🏻‍♂️ 

I stepped through each commit, focusing on review of the logic rather than tests. All the changes look good to me. I left a few inline suggestions to consider. None are blocking. 

I tested this branch in the Demo app on both iOS and Android. I tested media uploads for Simple, Atomic, and self-hosted sites. I did not encounter issues while testing. Images were correctly resized when necessary as configured by the Demo app, others were passthrough to the WordPress server.

PR #145, opened 2025-05-30
feat: Mark content outside of popovers inert

dcalhoun:
## What?

Mark content outside of the block inserter and block inspector [inert](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert).

## Why?

Close #86. Close CMM-192. Without this, screen readers unexpectedly navigate content that is visually
hidden by the modal popovers.

## How?

For the block inserter and block inspector:

- Enable modal ARIA attribute.
- Apply the dialog role.
- Mark content outside of the popover as inert.
- Enable closing via VoiceOver's two-finger scrub.

## Testing Instructions

> [!Tip]
> Test using the WP-iOS [prototype build](https://github.com/wordpress-mobile/WordPress-iOS/pull/24571#issuecomment-2923031781). 

**1: Content outside of popover is inert** [^1]
1. Enable [VoiceOver](https://support.apple.com/guide/iphone/turn-on-and-practice-voiceover-iph3e2e415f/ios).
1. Open the block inserter.
1. Activate the Media tab.
1. [Swipe up with two fingers](https://support.apple.com/guide/iphone/use-voiceover-gestures-iph3e2e2281/ios#:~:text=Two%2Dfinger%20swipe%20up).
1. Verify the screen reader does not announce HTML content behind the
   dialog—e.g., block canvas content.

[^1]: This also applies to Android TalkBack, but testing Android with these steps is not possible due to an existing bug: https://github.com/WordPress/gutenberg/issues/70277. 

**2: Dismiss dialogs with iOS' two-finger scrub**
1. Enable [VoiceOver](https://support.apple.com/guide/iphone/turn-on-and-practice-voiceover-iph3e2e415f/ios).
1. Open the block inserter.
1. Perform a [two-finger scrub](https://support.apple.com/guide/iphone/use-voiceover-gestures-iph3e2e2281/ios#:~:text=Two%2Dfinger%20scrub%20(move%20two%20fingers%20back%20and%20forth%20three%20times%20quickly%2C%20making%20a%20“z”)).
1. Verify the dialog is dismissed.

### Accessibility Testing Instructions

Listed in testing instructions.

## Screenshots or screencast 
N/A, no visual changes.

PR #562, opened 2026-07-23
fix: prevent dropdown menus from overflowing the editor root

dcalhoun:
## What?

Opening a block toolbar dropdown menu (e.g. the "Options" more menu) can add horizontal overflow to the editor's root element, letting the canvas be swiped sideways to reveal the empty background behind it.

## Why?

Ref CMM-997.

The editor is not iframed on mobile, so the whole editor lives in a single document. Block toolbar popovers render into a fallback container appended to `<body>` and are positioned relative to the viewport. When a popover is anchored near the right edge—such as after scrolling the block toolbar so the toggle sits at the far right—it extends past the viewport. Because nothing clips horizontal overflow on `<html>`/`<body>`/`#root`, this grows the scrollable width of the root element, and the canvas can be swiped horizontally to reveal the empty background behind the editor styles wrapper.

Measured with the menu open on a 375px-wide viewport: the popover's right edge lands at 419px (44px past the viewport), `html.scrollWidth` grows to 419 while `clientWidth` stays 375, and only `<html>` becomes horizontally scrollable.

## How?

Render popovers into slots within body-level containers, split by whether the popover needs clipping:

- **Dropdown popovers** use the default slot, placed in a `position: fixed; inset: 0; overflow: hidden` container. Being out of flow, it clips the overflow without contributing to the document's scrollable size, so the document remains the scroll container. The container spans the viewport, so it sets `pointer-events: none` and its slot content restores `pointer-events: auto`.
- **The full-screen block settings menu** opts into its own slot in an unclipped container via `__unstableSlotName`. It fills the viewport and never overflows, so it does not need clipping. It is also kept clear of a clipping ancestor as a precaution: with an earlier clipping approach the menu rendered semi-transparent in WebKit (editor content bleeding through). The exact trigger is narrower than "any `position: fixed` element under a clipped ancestor"—the block inserter is also fixed and full-screen and renders into the clipped container with no such artifact (verified on device)—but the menu gains nothing from clipping, so keeping it in an unclipped slot sidesteps the risk entirely.

Both containers are body children, so `modalize` can keep the active popover accessible while marking the rest of the document inert. The clip container is stacked one above the overlay container (`z-index`). Dropdowns opened from within the settings menu—color and typography controls, etc.—render into the default (clipped) slot; stacking the clip container above the opaque menu keeps them visible. They are anchored on-screen within the menu, so clipping them to the viewport is harmless. Routing them into the overlay slot instead is not possible: they arrive through the `InspectorControls` slot-fill, which renders them in the fill's own React context—outside any slot-name provider wrapped around `BlockInspector`—so a context override does not reach them.

A few supporting changes:

- **`modalize` now walks the ancestor path.** It previously hid only `document.body`'s children, which assumed the modal element was itself a body child—true for the fallback container, not for a popover nested in a slot. It now walks from each modal element up to the body, hiding the other children of every ancestor, collecting ancestors first so they are never hidden. It accepts multiple modal elements, and returns a handle so `unmodalize` reverses that specific batch by identity rather than assuming modals close in reverse-open order—the inserter and settings menu can be open at once, and React runs effect cleanups in hook-declaration order.
- **The layout is wrapped in a `SlotFillProvider`** so the slots and the editor's popovers share one registry. `BlockEditorProvider` renders `<SlotFillProvider passthrough>`, which creates its own registry when no provider exists above it, leaving the slots unreachable and silently sending popovers back to the fallback container.

Copying additional context from https://github.com/wordpress-mobile/GutenbergKit/pull/562#discussion_r3685373688 for posterity: 

> These styles [`src/components/popover-slots/style.scss` and the corresponding `Popover.Slot`s are the crux of the solution to the overflow problem. All other changes are adapting existing logic for compatibility with the new DOM structure. 
> 
> Essentially, we now have two separate "slots" in which popovers are rendered: one for popovers that need clipping to avoid horizontal body overflow receive (`#popover-clip-container`); another for all others that do not and operate the same as before (`#popover-overlay-container`). 
> 
> The problem is unique to GutenbergKit, in comparison to web editor, in that the former pursues relying upon `body` element scrolling to preserve iOS' tap-status-bar-to-scroll-to-top behavior. The web does not pursue this—instead relying upon children element scrolling—and can easily forcibly clip `body` overflow without concern for breaking the ability scrolling post content. Tapping the status bar when using the web editor does not scroll to the top. 

### Alternatives considered

`contain: paint` on the body (the previous approach in this PR) clips the overflow, but it also stops the document from scrolling content taller than the viewport. Restoring vertical scroll on `.gutenberg-kit-root` moves the scroll off the document, which **breaks iOS status-bar-tap scroll-to-top**—that gesture only drives the web view's main scroll view. Both rules are reverted here, so the document is the scroll container again.

`overflow-x: hidden`/`clip` on `html`/`body` does not clip out-of-flow popovers at all (the value propagates to the viewport), and clipping on `#root` misses them entirely since the fallback container is a sibling of `#root`, not a descendant.

## Testing Instructions

1. Run the iOS demo app and open an editor with theme styles (e.g. a site with a non-white editor background).
2. Focus the first empty placeholder Paragraph block.
3. Scroll the block toolbar to the end to reveal the more menu (three-dots) toggle, and tap it to open the menu.
4. Scroll the block toolbar back to its beginning so the expanded menu is at least partially off screen.
5. Swipe horizontally (right to left) on the editor canvas beneath the focused block.
6. Confirm the canvas stays put and no empty/background band is revealed on the right. (On `trunk`, the canvas shifts left and reveals the root element's background.)

Please also confirm the following, as they cover the constraints this approach had to satisfy:

7. Open the block settings menu (cog). Confirm it renders fully opaque, with no editor content bleeding through.
8. Within the open settings menu, tap the Text and Background color swatches, the Color and Typography options (three-dots) menus, and the font-size custom-size control. Confirm each dropdown appears on top of the menu.
9. Add enough content to overflow the viewport and confirm it scrolls, then **tap the status bar and confirm the editor scrolls to the top**.
10. Open the block inserter and the block settings menu and confirm both still dismiss via outside tap and close button.

Popovers now render into slots rather than Gutenberg's fallback container, and `modalize` was refactored to match, so the inert behaviour from #145 is worth re-checking with VoiceOver:

1. Open the block inserter (or the block settings menu) with VoiceOver enabled.
2. Confirm the popover's contents are reachable and announced.
3. Confirm swiping past the popover does not reach the editor canvas or toolbar behind it.
4. Dismiss the popover and confirm the editor content becomes reachable again.

Verified in the DOM: with a popover open, `#root` is `aria-hidden="true"` while both popover containers, the popover's ancestor chain, and the `aria-live` regions are untouched; dismissing removes every attribute that was added.

## Screenshots or screencast

Before | After |
-- | -- 
<video src="https://github.com/user-attachments/assets/a9c4af1f-a0a3-45ec-99bb-9e67557edb57" /> | <video src="https://github.com/user-attachments/assets/0199ec84-2738-4441-835d-c54c93e5e916" />

### DIFF
--- a/src/components/popover-slots/style.scss
+++ b/src/components/popover-slots/style.scss
@@ -1,0 +1,29 @@
+// Clips popovers to the viewport so they cannot grow the document's
+// scrollable width. `position: fixed` takes the container out of flow, which
+// both sizes it to the viewport and keeps it from contributing to the
+// document's scrollable size—leaving the document as the scroll container.
+//
+// `overflow-x` on the body or the root element does not clip popovers: the
+// value propagates to the viewport rather than clipping out-of-flow
+// descendants. Clipping on `#root` misses them entirely, as popovers render
+// into a sibling of `#root` rather than a descendant.
+#popover-clip-container {
+	inset: 0;
+	overflow: hidden;
+	position: fixed;
+	// The container spans the viewport, so it must not intercept touches
+	// intended for the editor. Popovers within it restore pointer events.
+	pointer-events: none;
+	z-index: 1000000; // Mirrors Gutenberg's `.components-popover` z-index
+}
+
+#popover-clip-container .components-popover {
+	pointer-events: auto;
+}
+
+// Popovers that must not be clipped—see `./index.jsx`. Left out of flow by its
+// popovers alone, which are positioned absolutely or fixed.
+#popover-overlay-container {
+	position: relative;
+	z-index: 1000000; // Mirrors Gutenberg's `.components-popover` z-index
+}
--- a/src/components/popover-slots/style.scss
+++ b/src/components/popover-slots/style.scss
@@ -18,20 +18,26 @@
 	overflow: hidden;
 	position: fixed;
 	// The container spans the viewport, so it must not intercept touches
-	// intended for the editor. Popovers within it restore pointer events.
+	// intended for the editor. Its slot content restores pointer events.
 	pointer-events: none;
 	z-index: 1000001; // One above the overlay container
 }
 
-#popover-clip-container .components-popover {
+// Restore pointer events on the slot content, not only on `.components-popover`
+// roots—a fill may portal in sibling markup that is not itself a popover.
+#popover-clip-container > * {
 	pointer-events: auto;
 }
 
 // Popovers that must not be clipped—namely the full-screen block settings
-// menu. WebKit renders a `position: fixed` element semi-transparent when an
-// ancestor clips its overflow, so this container does not clip. The menu fills
-// the viewport and never overflows, so it does not need clipping. See
-// `./index.jsx`.
+// menu. This container does not clip, so it does not stack a WebKit
+// compositing bug: an `overflow: hidden` ancestor rendered the fixed,
+// full-screen settings menu semi-transparent. (The block inserter is also a
+// fixed, full-screen popover but renders into the clipped container above
+// without that artifact—so the trigger is narrower than any clipped fixed
+// element; the menu is kept clear of clipping as a precaution.) The menu fills
+// the viewport and never overflows, so clipping would gain nothing regardless.
+// See `./index.jsx`.
 #popover-overlay-container {
 	position: relative;
 	z-index: 1000000; // Mirrors Gutenberg's `.components-popover` z-index
--- a/src/components/popover-slots/style.scss
+++ b/src/components/popover-slots/style.scss
@@ -7,6 +7,12 @@
 // value propagates to the viewport rather than clipping out-of-flow
 // descendants. Clipping on `#root` misses them entirely, as popovers render
 // into a sibling of `#root` rather than a descendant.
+//
+// Stacks above the overlay container (below). Dropdowns opened from within the
+// full-screen block settings menu—color and typography controls, etc.—render
+// here, and the settings menu is opaque, so they must paint above it to be
+// visible. They are anchored on-screen within the menu and so stay within the
+// viewport clip.
 #popover-clip-container {
 	inset: 0;
 	overflow: hidden;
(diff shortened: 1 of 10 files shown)


src/components/popover-slots/style.scss
dcalhoun, line 1: These styles and the corresponding `Popover.Slot`s are the crux of the solution to the overflow problem. All other changes are adapting existing logic for compatibility with the new DOM structure. 

Essentially, we now have two separate "slots" in which popovers are rendered: one for popovers that need clipping to avoid horizontal body overflow receive (`#popover-clip-container`); another for all others that do not and operate the same as before (`#popover-overlay-container`). 

The problem is unique to GutenbergKit, in comparison to web editor, in that the former pursues relying upon `body` element scrolling to preserve iOS' tap-status-bar-to-scroll-to-top behavior. The web does not pursue this—instead relying upon children element scrolling—and can easily forcibly clip `body` overflow without concern for breaking the ability scrolling post content. Tapping the status bar when using the web editor does not scroll to the top.

nbradbury:
@dcalhoun I asked Claude for a review and it found three potential issues.

[review-pr562-modalize-findings-2026-07-31.pdf](https://github.com/user-attachments/files/30586790/review-pr562-modalize-findings-2026-07-31.pdf)

nbradbury:
@dcalhoun I verified the fix works in the iOS simulator so I'll approve it. I'm not sure if any of Claude's findings are valid so feel free to merge after reviewing them.

nbradbury (approved):
:shipit:

dcalhoun:
Addressed the shared findings in https://github.com/wordpress-mobile/GutenbergKit/pull/562/commits/0c74e65c7cc7f94dacb2ef33beb7f5ade6ae42f7, then re-tested.

Issue #564, opened 2026-07-27
Add stylelint with logical-property rule to catch RTL styling bugs


dcalhoun:
## What problem does this address?

GutenbergKit's own SCSS is never processed by `rtlcss`. Gutenberg's stylesheets ship pre-generated `-rtl` variants, and the editor now loads whichever matches the text direction, but our own styles under `src/` have no such counterpart — a physical property written there stays physical in both directions.

That produced a real bug during RTL testing: `.gutenberg-kit-editor-toolbar .components-toolbar-group` used `border-right-color`, which put the toolbar group divider on the wrong side in right-to-left layouts. It was found by eye, and nothing prevents the next one.

There is currently no stylelint setup in the repo at all — no config, no dependency, no `lint:css` script, no Makefile target — so this class of bug has no automated guard.

## What is your proposed solution?

Add stylelint with `stylelint-plugin-logical-css`, matching how upstream Gutenberg catches the same problem (`tools/stylelint/config.js`).

**Setup**

- Dependencies: `stylelint`, `stylelint-plugin-logical-css`, `@wordpress/stylelint-config`, `postcss-scss`
- `.stylelintrc.mjs` extending `@wordpress/stylelint-config/scss`, enabling `plugin/use-logical-properties-and-values`
- Reuse upstream's `ignore` list, which exempts vertical and direction-neutral properties (`margin-top`, `width`, `overflow-y`, `border-top`, …)
- `lint:css` / `lint:css:fix` scripts in `package.json`, plus `lint-css` / `lint-css-fix` Makefile targets matching the existing `lint-js` pattern
- Run `lint-css` wherever CI runs `lint-js`

**Existing violations**

The asymmetric cases that actually misrendered have already been fixed. What remains across the 11 SCSS files under `src/`:

- **15 physical-property declarations**, all symmetric pairs (`padding-left: 8px; padding-right: 8px`, `margin-left: auto; margin-right: auto`, `border-left: none; border-right: none`) or `text-align: center`. None misrender in RTL, but the rule will flag them because it cannot tell a symmetric pair is direction-safe. Converting to the shorthand (`padding-inline: 8px`) is mechanical and arguably cleaner.
- **10 `left`/`right` positioning declarations** that are symmetric full-width pins (`left: 0; right: 0`, `left: 16px; right: 16px`). Same situation — safe today, flagged by the rule, trivially convertible to `inset-inline`.

[...]

PR #566, opened 2026-07-27
fix: render the editor in the locale's text direction

### LINKED ISSUE #565: Toolbar scroll-indicator gradients render on the wrong side in RTL layouts
## Description

The editor toolbar's scroll-indicator gradients are positioned and computed using physical left/right values, and render incorrectly in right-to-left layouts. Confirmed on device: the gradient does not display on the correct side for the current scroll position and overflow content.

Unlike the other physical-property usages in `src/`, this one cannot be fixed by swapping to logical CSS alone — the CSS and the JavaScript that drives it are coupled, and the JS half depends on how WebKit reports `scrollLeft` in an RTL container.

**CSS** — `src/components/editor-toolbar/style.scss:44-91`

Two different gradients pinned to opposite edges, each with a physical position _and_ a physical gradient direction:

- `&::before` — `left: 0`, `background: linear-gradient(to right, …)`
- `&::after` — `right: 0`, `background: linear-gradient(to left, …)`

**JS** — `src/components/editor-toolbar/use-scroll-indicators.js:32-34`

```js
const canScrollLeft = scrollLeft > threshold;
const canScrollRight = scrollLeft + clientWidth < scrollWidth - threshold;
```

These set the `has-scroll-left` / `has-scroll-right` classes (`src/components/editor-toolbar/index.jsx:95-96`) that toggle each gradient's opacity.

In an RTL container, browsers report `scrollLeft` as negative or inverted depending on the engine, scrolling right-to-left from zero. If WebKit does not normalise it, `canScrollLeft` is never true and `canScrollRight` misbehaves, so the gradients appear on the wrong edge or not at all.

## Step-by-step reproduction instructions

1. Set the demo app to a right-to-left language — Arabic or Hebrew. On iOS use the scheme's _App Language_; on Android use Settings > Apps > GutenbergKit > Language.
2. Open the editor and select a block whose toolbar overflows horizontally (add enough formatting controls that the toolbar scrolls).
3. Scroll the toolbar in both directions.
4. Observe whether the fade gradients appear on the correct edges — the gradient should mark the edge with more content beyond it.

**Expected:** gradients appear on whichever edge has off-screen content, mirroring correctly in RTL.

**Actual:** the gradient does not display on the correct side for the current scroll position and overflow content.

The fix spans both the CSS (logical positioning plus gradient direction) and the scroll-state calculation, which needs to account for how `scrollLeft` is reported in an RTL container.

## Environment info

- Affects the GutenbergKit editor in the iOS and Android demo apps, and any host, when running a right-to-left locale.
- Confirmed in the demo app with a right-to-left language selected.
- RTL rendering support was added in the branch that also fixed the toolbar border-side bug; before that the editor never rendered RTL at all, so this would not have been observable.

dcalhoun:
## What?

Renders the editor in the reading direction of the loaded translations, instead of always left-to-right, and declares the editor's language to assistive technology. Both demo apps gain a way to switch languages so it can be tested.

## Why?

Fix CMM-1099. Fix CMM-2188.

The editor loaded the correct translations for all 49 shipped locales but always rendered them left-to-right. `src/index.html` hardcodes `lang="en"` with no `dir`, and nothing set either at runtime.

For the four right-to-left locales we ship translations for — Arabic, Persian, Hebrew, Urdu — that meant:

- **Mirrored layout was absent.** CSS logical properties resolved as left-to-right, so toolbars, sidebars, and inspector panels laid out backwards.
- **Bidirectional text ran wrong.** Mixed-direction content resolved under a left-to-right base direction, so trailing punctuation landed on the wrong end — in the user's post content, not just the chrome.
- **Controls pointed the wrong way.** Components across `components`, `block-editor`, `block-library`, and `editor` call `isRTL()` at runtime to choose icons, accessibility labels, keyboard navigation, and drop-zone geometry. It returned `false`, so block movers pointed the wrong way _and_ announced the opposite of what they did.
- **Assistive technology mispronounced the UI.** `lang="en"` tells screen readers to apply English phonetics to every language — a WCAG 3.1.1 failure that affects all 49 locales, not only the right-to-left ones. VoiceOver honors the corrected `lang`; TalkBack's support is partial, see below.

In WordPress, core establishes all of this: it renders `<html lang dir>` and `<body class="rtl">`, populates the `text direction` string behind `isRTL()`, and swaps each enqueued stylesheet for its `-rtl` variant. GutenbergKit loads a static `index.html`, so nothing performed that role.

Neither demo app called `setLocale`, so the editor always ran in English regardless of the device or scheme language — which is why this went unnoticed.

## How?

**Text direction** — `configureLocale()` derives direction from the resolved locale and applies it two ways: `documentElement.lang` / `dir` and `body.rtl` for CSS, and a `setLocaleData()` entry for the `text direction` key that backs `isRTL()`. The second half is what fixes the icons, labels, and keyboard navigation that CSS cannot reach. The string is injected rather than read from the bundle because the `wp-plugins/gutenberg` GlotPress project does not carry it — it belongs to core.

**Stylesheets** — both the left-to-right and right-to-left variants are imported, and one is injected at runtime. The `-rtl` bundles are full rewrites rather than overrides: across the five editor stylesheets roughly 690 selectors appear in both files with conflicting declarations and almost no `[dir=rtl]` scoping, so loading both would let source order decide the direction every user gets. Both variants ship in the bundle rather than being fetched on demand — the assets are already bundled into the host app, so the added weight costs no network time, and selecting synchronously avoids introducing async work before first paint.

The block canvas iframe needs no change: it inherits `dir` from the parent document.

**Accessibility language** — the web content declares its own language via `documentElement.lang`, but the native UI presented alongside it does not. `EditorViewController` sets `accessibilityLanguage` from the configured locale in `viewDidLoad`, so VoiceOver selects a matching speech voice. The value is inherited, including across modal presentations, so the block inserter and the sheets it presents are covered by that single assignment.

**Toolbar scroll indicators** — the gradients marking overflow read `scrollLeft` directly, which is `0` at the right edge and grows negative moving left in a right-to-left container. The left gradient was therefore unreachable and the right one showed at rest. The offset is normalized to a distance from the start and mapped onto the physical edges the gradients are anchored to, which do not flip with the writing direction. Fixes #565.

**Demo apps** — iOS reads the scheme's _App Language_; Android declares a `localeConfig` so it appears in the system's per-app language picker, and links to it from the configuration screen. Both surface the resolved locale, since a language with no bundled translations falls back to English, which is otherwise indistinguishable from the selection being ignored.

`DemoAppLocale.swift` duplicates the resolution chain that #492 adds to the library as `LocaleResolver`, matching Android's already merged in #493. It is scoped to one file so reviving that PR removes it wholesale, leaving a one-line call change. Android needs no such helper — `setLocale(Locale)` already resolves.

## Testing Instructions

**iOS**

1. Edit the `Gutenberg` scheme → Run → Options → App Language, and select **Arabic** (or Hebrew).
2. Run the demo app and open a site's Editor Configuration.
3. Confirm the **Editor Locale** row reads `ar`.
4. Start the editor and confirm the UI is mirrored, the toolbar reads right-to-left, and translations are in Arabic.
5. Repeat with **Spanish** to confirm a left-to-right locale still translates without mirroring.
6. Set App Language back to **System Language** and confirm the editor returns to English, left-to-right.

Xcode's _Right-to-Left Pseudolanguage_ options are deliberately not supported — they are layout overrides rather than languages, so the app around the editor flips while the editor does not. Use a real right-to-left language.

**Android**

1. Run the demo app and open a site's Editor Configuration.
2. Tap **Change** on the Editor Locale row to open the system language picker (Android 13+).
3. Select **العربية**, then return to the app.
4. Confirm the Editor Locale row updates to `ar` without backing out of the screen.
5. Start the editor and confirm the UI is mirrored and translations are in Arabic.
6. Repeat with **Español** to confirm a left-to-right locale still translates without mirroring.
7. Set the language back to **System default** and confirm the editor returns to English.

The native block inserter stays English on Android in every locale — its strings are Android resources with no localized variants. Unchanged by this PR, and noted under Follow-ups.

**Speech voice — iOS**

1. Install a voice for the test language under Settings → Accessibility → VoiceOver → Speech, and set the app to that language.
2. Navigate the editor. Confirm strings are spoken by that language's voice rather than the default one.
3. Open the native block inserter and confirm the same, including the patterns sheet.

Applies to every locale, not only the right-to-left ones.

**Speech voice — Android: partially out of reach**

Nothing to test. TalkBack does not apply `lang` when it builds accessibility nodes for interactive controls.

## Screenshots or screencast 

| Android                                                                                                                                | iOS                                                                                                                                |
| -------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
| android | ios |

## Follow-ups

- #564 — add stylelint with a logical-property rule, so physical properties in our own SCSS are caught before they ship

Android's native strings are not localized. The library ships only `values/`, and unlike iOS there is no host-override hook, so the block inserter is English in every locale. Deferred — it needs either localized resources or an `EditorLocalization` equivalent, and is independent of text direction.

Preloaded REST data is not localized — GutenbergKit sends no `Accept-Language` or `_locale`, so post-type labels and theme strings stay English in every locale. Out of scope here; the response cache is keyed only by method and URL, which is correct today precisely because nothing locale-specific is ever fetched. Worth revisiting if that changes.


src/index.scss
```diff
@@ -67,12 +67,6 @@ $baseline-interactive-font-size: 17px;
 		min-height: 100vh;
 	}
 
-	.block-inspector-siderbar {
-		background: #f6f6fbff;
-		border-left: 0.5px solid #c8c7cc;
-		width: 320px;
-	}
-
```
dcalhoun, line 75: Discovered as dead code and removed.

ios/Sources/GutenbergKit/Sources/Views/EditorAccessibilityLanguage.swift
dcalhoun, line 1: This approach makes sense to me at a high level, but I welcome intentional review of and feedback on this.

ios/Demo-iOS/Sources/Views/SitePreparationView.swift
```diff
@@ -283,6 +314,7 @@ class SitePreparationViewModel {
     private static func applyDemoAppDefaults(to configuration: EditorConfiguration) -> EditorConfiguration {
         configuration.toBuilder()
             .setNativeInserterEnabled(true)
+            .setLocale(DemoAppLocale.current)
```
crazytonyli, line 317: Is it a requirement for the app to call this function to support text direction in GBK? If we need to call this function, I presume we'd need to duplicate the `resolve` in the app, too?

ios/Sources/GutenbergKit/Sources/Views/EditorAccessibilityLanguage.swift
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/// Applies the language once the view is in a hierarchy.
+///
+/// `updateUIView` runs before this view is necessarily attached to a superview,
+/// and with no superview the walk up the responder chain finds no hosting
+/// controller. Because the language does not change afterward, SwiftUI has no
+/// reason to call `updateUIView` again, so a miss there would be permanent.
+/// Re-applying on move to a window retries once the hierarchy exists.
+private final class AccessibilityLanguageView: UIView {
```
crazytonyli: Any particular reason creating this view, instead of updating `accessibilityLanguage` on the EditorViewController?

ios/Demo-iOS/Sources/Views/SitePreparationView.swift
dcalhoun: Yes, that is by design. We do currently invoke this function [within WordPress-iOS](https://github.com/wordpress-mobile/WordPress-iOS/blob/0e41da3424520cceabd13a98d728230bd340f6e1/WordPress/Classes/Utility/Editor/EditorConfiguration%2BBlog.swift#L58). 

Regarding `resolve`, that is pointedly addressed for Android by #493. The in-progress #492 does so iOS. So, it's an existing problem not addressed by this PR. 

The `DemoAppLocale` is intended as a stopgap for resolving locales on iOS to unblock addressing the issues stemming from the currently absent `lang` and `dir` support. When merging #492, we should remove `DemoAppLocale` altogether. I opted to keep the two matters (locale resolutions and `lang`/`dir`) separate to avoid entangling PRs.

ios/Sources/GutenbergKit/Sources/Views/EditorAccessibilityLanguage.swift
dcalhoun: I appreciate this feedback and guidance. I re-tested with your suggested approach in both GBK Demo and WP-iOS and it does work as it should. 

At some point, my earlier testing encountered unexpected VoiceOver Speech Voices that mismatched the selected locale when navigating native modals (e.g., native block inserter sheet, patterns sheet, camera sheet). With Claude, I came to believe the sibling nature of the modal presentations were the issue and required this more complex approach. From testing your proposed change, it that appears misguided. 

Addressed this in https://github.com/wordpress-mobile/GutenbergKit/pull/566/commits/9bc49b5ea915a72786e24b9cd6778019746333f2.

crazytonyli (approved):

PR #569, opened 2026-07-28
fix(ios)!: localize the patterns count and let hosts fall back to editor strings

dcalhoun:
## What?

Localizes the patterns count string, and lets host apps fall back to the editor's own strings for keys they do not translate.

> [!WARNING]
> **Source-breaking for host apps.** `EditorLocalization.localize` now returns `String?` instead of `String`, so every host that assigns it needs a signature update — independent of the new `patternsCount(Int)` case, which also stops an exhaustive switch from compiling. WordPress-iOS needs a companion PR before it can adopt this version: https://github.com/wordpress-mobile/WordPress-iOS/pull/25848

## Why?

The patterns count rendered as a hardcoded English string. Fixing that means adding a case to `EditorLocalizableString`, which exposed a design problem worth fixing in the same change.

Hosts override `EditorLocalization.localize` by **replacing** the closure, so they switch over the enum exhaustively with no `default:`. Swift requires that switch to be exhaustive, so every string the editor adds breaks the host build — and the break surfaces later, when the host bumps the dependency, not when the string is added. A host that does nothing gets a build failure rather than an untranslated string, which blocks them from adopting unrelated fixes until someone writes a translation.

That penalty is disproportionate for this localization gap. Android string resources fall back through `values-ar/` → `values/`, and gettext returns the source English when no translation exists; neither fails the build. The editor's own JS translations already work this way.

## How?

**Declining a key** — `localize` returns `String?`. A host returns `nil` for keys it does not translate, and the editor supplies its own string:

```swift
EditorLocalization.localize = { key in
    switch key {
    case .showMore: NSLocalizedString("editor.blockInserter.showMore", ...)
    // ...keys the host translates.
    @unknown default: nil
    }
}
```

Once a host writes that fallback case, future keys render untranslated instead of failing to compile.

An optional return rather than a public fallback function the host calls itself, so the fallback cannot be skipped or reimplemented: `nil` is the only way to decline, and it routes through the editor's defaults and reporting. Every read goes through the `EditorLocalization[...]` subscript, which unwraps, falls back, and reports in one place.

**Reporting** — falling back logs each untranslated string once per process, at `notice` so it persists to the log store without the host configuring `EditorLogger`. Reporting once per key matters because every call site sits in a SwiftUI `body` that re-evaluates on each render pass; logging per read would write an entry per row per frame while a list scrolls. Keys are deduplicated by case name, so `patternsCount(3)` and `patternsCount(7)` count as one missing string.

Only a `nil` return reports. The default closure answers every key, so strings the editor reads while building views — `loadingEditor` among them, which can run before the host assigns `localize` — are never reported as missing.

Hosts that render the editor's own strings deliberately can opt out with `EditorLocalization.reportsMissingTranslations = false`.

**Isolation** — `@MainActor` sits on `localize` and the subscript rather than the whole class. `localize` holds a non-`Sendable` closure and the subscript reads it; nothing else needs isolation, and no annotation is forced onto host code. Verified with `swiftc -typecheck -swift-version 6 -strict-concurrency=complete` against a file reproducing WordPress-iOS's call pattern, including that reads off the main actor are still rejected.

## Testing Instructions

This is not testable in the Demo app. See testing instructions in https://github.com/wordpress-mobile/WordPress-iOS/pull/25848.

No UI changes. Strings that previously rendered in English still do; the change is which code supplies them.

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorLocalization.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorLocalization.swift
@@ -74,14 +74,16 @@ public final class EditorLocalization {
     /// having to opt in. Set to `false` in apps that render the editor's own
     /// strings deliberately, where every fallback is expected and the reports
     /// are noise.
-    public static var reportsMissingTranslations: Bool {
-        get { reportingLock.withLock { _reportsMissingTranslations } }
-        set { reportingLock.withLock { _reportsMissingTranslations = newValue } }
-    }
+    ///
+    /// Unsynchronized: hosts set this once during setup, before any editor view
+    /// reads a string, so the worst a race could cost is a single stray report.
+    public nonisolated(unsafe) static var reportsMissingTranslations = true
 
-    private static let reportingLock = NSLock()
-    nonisolated(unsafe) private static var _reportsMissingTranslations = true
-    nonisolated(unsafe) private static var reportedKeys: Set<String> = []
+    /// Keys already reported, so each is logged once. Guarded rather than
+    /// `nonisolated(unsafe)` because `Set` is not safe to mutate concurrently.
+    private static let reportedKeys = OSAllocatedUnfairLock<Set<String>>(
+        initialState: []
+    )
 
     /// The editor's untranslated strings.
     private static func defaultString(
@@ -118,14 +120,13 @@ public final class EditorLocalization {
     private static func reportMissingTranslation(
         for key: EditorLocalizableString
     ) {
+        guard reportsMissingTranslations else { return }
+
         // Associated values distinguish cases that share a translation:
         // `patternsCount(3)` and `patternsCount(7)` are one missing string.
         let name = String(String(describing: key).prefix { $0 != "(" })
 
-        let shouldReport = reportingLock.withLock {
-            _reportsMissingTranslations && reportedKeys.insert(name).inserted
-        }
-        guard shouldReport else { return }
+        guard reportedKeys.withLock({ $0.insert(name).inserted }) else { return }
 
         // Logged through `OSLog` rather than `EditorLogger`, which reaches only
         // hosts that install a logger and raise the log level. This message is
@@ -161,9 +162,7 @@ public final class EditorLocalization {
     /// Clears the record of which keys have already been reported so tests do
     /// not leak state into each other.
     static func resetMissingTranslationReportingForTesting() {
-        reportingLock.withLock {
-            _reportsMissingTranslations = true
-            reportedKeys.removeAll()
-        }
+        reportsMissingTranslations = true
+        reportedKeys.withLock { $0.removeAll() }
     }
 }
--- a/ios/Sources/GutenbergKit/Sources/EditorLocalization.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorLocalization.swift
@@ -51,14 +51,16 @@ public final class EditorLocalization {
     ///     switch key {
     ///     case .showMore: NSLocalizedString("editor.blockInserter.showMore", ...)
     ///     // ...keys the host translates.
-    ///     default: nil
+    ///     @unknown default: nil
     ///     }
     /// }
     /// ```
     ///
-    /// Declining with `default: nil` rather than switching exhaustively keeps
-    /// the host compiling when the editor adds a string: the new key renders
-    /// untranslated instead of breaking the build.
+    /// Declining rather than switching exhaustively keeps the host compiling
+    /// when the editor adds a string: the new key renders untranslated instead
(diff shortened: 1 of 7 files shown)


ios/Sources/GutenbergKit/Sources/EditorLocalization.swift
```diff
@@ -39,7 +41,81 @@ public enum EditorLocalizableString {
...
+    /// Host apps are encouraged to delegate unhandled keys to
+    /// ``defaultLocalize`` rather than switching exhaustively:
+    ///
+    /// ```swift
+    /// EditorLocalization.localize = { key in
+    ///     switch key {
+    ///     case .showMore: NSLocalizedString("editor.blockInserter.showMore", ...)
+    ///     // ...keys the host translates.
+    ///     @unknown default: EditorLocalization.defaultLocalize(key)
+    ///     }
```
crazytonyli, line 54: Should `EditorLocalization.localize` allow returning a nil, which means GBK can call `defaultLocalize` internally, rather than requiring the app to call it?

ios/Sources/GutenbergKit/Sources/EditorLocalization.swift
```diff
@@ -39,7 +41,81 @@ public enum EditorLocalizableString {
 @MainActor
```
crazytonyli: Majority of this class are `nonisolated` functions. Should it be bound to the main actor at all?

ios/Sources/GutenbergKit/Sources/EditorLocalization.swift
```diff
@@ -39,7 +41,81 @@ public enum EditorLocalizableString {
...
+    ///
+    /// ``defaultLocalize`` reads this, and host apps call that from whatever
+    /// context their own localization runs in, so it cannot be isolated to the
+    /// main actor. Guarded by the same lock as the reported-key set below.
+    public nonisolated static var reportsMissingTranslations: Bool {
+        get { reportingLock.withLock { _reportsMissingTranslations } }
+        set { reportingLock.withLock { _reportsMissingTranslations = newValue } }
+    }
+
+    private nonisolated static let reportingLock = NSLock()
```
crazytonyli: I feel like adding this lock increases complexity. I don't think it's too much to log missing translations repeatedly on the same keys.

ios/Sources/GutenbergKit/Sources/EditorLocalization.swift
dcalhoun: This approach makes sense. Thanks for the suggestion. Addressed in https://github.com/wordpress-mobile/GutenbergKit/pull/569/commits/42b4bb4324aa8d0a2517c9448b0c04f8d9364d6c.

ios/Sources/GutenbergKit/Sources/EditorLocalization.swift
dcalhoun: I believe I refactored the class to align with your suggestion, but let me know if I did not. See https://github.com/wordpress-mobile/GutenbergKit/pull/569/commits/b08789c18e59c97972f87ba6c02b5d7b5a58f02a.

ios/Sources/GutenbergKit/Sources/EditorLocalization.swift
dcalhoun: I agree there is definitely increased complexity to achieve de-duping logs. I attempted simplifying it in https://github.com/wordpress-mobile/GutenbergKit/pull/569/commits/b681e106123c9b11b9bdd7923ea18f227dc5a85f. Let me know what you think of the state of it now. 

I agree losing de-duping is likely not a big issue. The one outlier is the Patterns sheet, where a single missing translation string would result in numerous logs as you scroll the sheet. However, if you feel it's still worth dropping the de-duping entirely, let me know, I'm open to that.

ios/Sources/GutenbergKit/Sources/EditorLocalization.swift
crazytonyli: I think you should be able to delete the `nonisolated` declarations now.

dcalhoun: I believe the only remaining `nonisolated` is with [`reportsMissingTranslations`](https://github.com/wordpress-mobile/GutenbergKit/blob/7b48e6db33eeb332b0d501f286cafb92deb92f58/ios/Sources/GutenbergKit/Sources/EditorLocalization.swift#L81). When removing it, I see an error: `Static property 'reportsMissingTranslations' is not concurrency-safe because it is nonisolated global shared mutable state`. 

Should this be removable or are you referencing other declarations? 

Screenshot 2026-07-29 at 21 05 18

crazytonyli (approved):

PR #572, opened 2026-07-31
build: bump wordpress-rs to 0.6.0 in the demo apps

dcalhoun:
## What?

Bumps `wordpress-rs` from pre-release pins to the `0.6.0` release on **both platforms**, and adapts to the API changes that release introduced.

## Why?

Neither demo app can load a self-hosted site's data when the site uses **plain permalinks**. Such a site has no path-based REST root — WordPress advertises the query form `https://site/?rest_route=/` instead. Both apps were pinned to March builds that constructed endpoint URLs by extending the root's *path*, leaving the `rest_route` query dangling, so requests returned a non-JSON page.

The symptom differed by platform, which is why this went unnoticed on Android:

- **iOS** surfaced it — Editor Configuration failed outright with "Response couldn't be parsed: expected value at line 1 column 1".
- **Android** hid it — `loadPostTypes` returns `null` on failure and the caller substitutes a hardcoded `post`, so the screen opened but the post-type picker listed only "Posts" instead of the site's real types.

`wordpress-rs` [fixed this in `2924e36b`](https://github.com/Automattic/wordpress-rs/commit/2924e36b) (July 6), included in [`0.6.0`](https://github.com/Automattic/wordpress-rs/releases/tag/0.6.0): `WpOrgSiteApiUrlResolver::resolve` now appends into the `rest_route` value rather than the path. Both pins predated it by ~4 months.

## How?

**iOS** — package requirement moves from `kind = revision` / `alpha-20260313` to `0.6.0`. Replaces the removed `apiRootUrl` and `apiUrlResolver` arguments on `WordPressAPI` with the new `SiteInfo` enum, matching [WordPress-iOS#25516](https://github.com/wordpress-mobile/WordPress-iOS/pull/25516).

**Android** — version catalog moves from `trunk-dc86c7a6` to `0.6.0`. This required two build changes **in the demo app only**:

- `compileSdk`/`targetSdk` 34 → 36, because 0.6.0 pulls okhttp 5.4.0, which requires API 36.
- `minSdk` 24 → 26, required by the 0.6.0 AAR manifest.

Only `:app` depends on wordpress-rs, so the published `:Gutenberg` library keeps `compileSdk` 34 and `minSdk` 24 — **this is not a breaking change for library consumers.** WordPress-Android already builds against 36 with `minSdk` 26 and consumes wordpress-rs `0.6.0`, so the demo app is not moving ahead of its host.

0.6.0 also removed `wordpressComSiteApiRoot()` from the `wp_mobile` account layer. It's replaced with a local helper using the identical URL format (`…/wp/v2/sites/{id}`), which matters because the demo app parses the site ID back out of the stored account in two places. WordPress-Android has no equivalent to update — it uses only the client and login APIs, not wordpress-rs account persistence.

Demo apps and their build configuration only — no changes to the published GutenbergKit library or its SDK levels.

## Testing Instructions

You need a self-hosted site set to **Settings ▸ Permalinks ▸ Plain**. Confirm it advertises the query-form REST root before starting:

```
curl -sI https://your-site.com/ | grep -i '^link'
```

The `rel="https://api.w.org/"` value should contain `?rest_route=/` rather than `/wp-json/`.

**iOS**

1. Open `ios/Demo-iOS/Gutenberg.xcodeproj`.
2. If the build reports a missing `libwordpressFFI.xcframework`, run **File ▸ Packages ▸ Resolve Package Versions** — the pin change invalidates the cached binary artifact.
3. Build, run, and add the plain-permalink site using an application password.
4. Open its Editor Configuration screen.

Before: fails with "Response couldn't be parsed: expected value at line 1 column 1". After: the screen loads and the API Root row shows the `?rest_route=` form.

**Android**

1. Build and run the demo app, then add the same plain-permalink site.
2. Open its Editor Configuration screen and check the **Post Type** picker.

Before: only "Posts". After: the site's real post types, including "Pages".

**Both** — confirm no regression for a pretty-permalink site and a WP.com site; each should behave as before. On Android, WP.com sign-in is worth exercising specifically, since it covers the `wordPressComSiteApiRoot` replacement.

Note that reaching a fully working editor on a plain-permalink site also needs the library fix in `fix/support-plain-permalinks`, which is stacked on this branch — this PR gets site preparation past the failures above.

N/A — no user interface changes.


ios/Demo-iOS/Sources/Views/SitePreparationView.swift
```diff
@@ -318,6 +323,20 @@ class SitePreparationViewModel {
             .build()
     }
 
+    /// Whether the site could not be reached at all — the host did not resolve
+    /// or refused the connection.
+    ///
+    /// wordpress-rs intercepts the underlying `URLError` and rewraps it as a
+    /// `WpApiError`, so matching `URLError` alone never fires for a site that
+    /// is simply not running.
+    private static func isUnreachableSiteError(_ error: Error) -> Bool {
```
jkmassel, line 332: We should probably lower this into wprs on either the Swift or Rust layers

jkmassel (approved):

ios/Demo-iOS/Sources/Views/SitePreparationView.swift
dcalhoun: Captured as a follow up task in https://github.com/wordpress-mobile/GutenbergKit/issues/578.

PR #575, opened 2026-08-01
build: lint swift code via a pinned SwiftLint plugin

dcalhoun:
## What?

Makes `make lint-swift` actually work, and runs it on CI.

## Why?

The target has been broken since it was introduced in #146. It ran `swift package plugin swiftlint`, but no SwiftLint plugin was ever declared as a dependency:

```
$ make lint-swift
error: Unknown subcommand or plugin name 'swiftlint'
```

There was also no `.swiftlint.yml`, no rule set, and no CI step — despite a `swiftlint:disable` comment already sitting in the Demo app, written for a linter that had never run. Swift was the only one of the three languages without lint coverage; JS has ESLint and Android has Detekt.

Enabling it surfaced 124 violations, including one genuine `error`-severity finding (details below).

## How?

Mirrors how WordPress-iOS manages SwiftLint:

- **`BuildTools/`** package depending on `SwiftLintPlugins`, which keeps the linter out of the dependency graph of anyone consuming GutenbergKit via SwiftPM.
- **`.swiftlint.yml`** porting the WordPress-iOS opt-in (`only_rules:`) set, so both codebases stay consistent for the shared team. `swiftlint_version` is the single source of truth — `BuildTools/Package.swift` parses it at manifest-evaluation time, so for local and hook runs the binary and the rules cannot drift.
- **Makefile**: fixed `lint-swift`, added `lint-swift-fix`, and added a `SWIFT_LINT_PATHS` variable for scoping a run to specific files (newline-separated, so paths containing spaces stay intact).
- **CI**: added a `:swift: SwiftLint` step alongside the JS and Android ones, and added it to the release step's `depends_on` — which already gated on `lint-js`, so a release could previously publish with Swift lint failing. Per review, it runs on the shared `linter` agent queue, matching Automattic-Tracks-iOS. That queue's image supplies its own SwiftLint, so CI is not covered by the `swiftlint_version` pin above.
- **Editor hook**: a `PostToolUse` hook runs after any Swift edit and reports violations back for self-correction, ported from WordPress-iOS. It shells out to `make lint-swift` so it shares one definition of the binary and config with the project command. Per review, it lints the whole project rather than the edited file: process startup dominates the runtime (2.2s vs 1.5s locally), so scoping saves little, and a full run also catches violations in files the edit did not name. Scoping falls out of `.swiftlint.yml`'s `included:`/`excluded:` keys, which keep sibling working directories and the generated sources out of the run.

Deviation from WordPress-iOS: dropped their `swiftui_localization` rule (bans `LocalizedStringKey`). It encodes a WP-iOS tooling constraint, and this codebase has zero `LocalizedStringKey` usages, so adopting it would assert a convention we haven't actually adopted.

### The violations

Commits are split so the tooling is reviewable apart from the churn: infrastructure, then 105 autocorrections, then the 19 that needed judgment.

Two of those are worth a look:

**`weak_delegate` in `HTMLWebViewRenderer` was a false positive** — and the only `error`-severity hit. `PooledWebView` is the sole owner of its `RenderDelegate`, the delegate holds no reference back, and `WKWebView.navigationDelegate` is already weak. Making the property `weak` would deallocate the delegate right after `init` and rendering would silently never complete. Silenced with a comment explaining why, rather than "fixed". The rule's fix is not even applicable here: `weak` forces the property optional, so the change does not compile without further edits, and the compiler flags it directly — `warning: weak reference will always be nil because the referenced object is deallocated here`.

**7 of the 17 `empty_string` violations compare an `Optional<String>` to `""`.** That asserts present-_and_-empty, which `isEmpty` cannot express — and it is the distinction those tests exist to make (`emptyQuotedValue` checks a quoted empty parameter parses to `""` rather than `nil`). Rewriting them would have quietly weakened the assertions, so those 7 are silenced individually; the other 10 became `isEmpty`.

## Testing Instructions

1. `make lint-swift` — passes clean (exit 0).
2. Introduce a violation, e.g. add trailing whitespace to any file under `ios/Sources/`, and re-run. It should be reported with a `file:line:col:` diagnostic and a non-zero exit.
3. `make lint-swift-fix` — auto-corrects it.
4. `make lint-swift SWIFT_LINT_PATHS="$(git diff --name-only trunk -- '*.swift')"` — scopes the run to the changed files.
5. `swift test` — 906 tests pass, confirming the assertion rewrites preserved test semantics.
6. Confirm the `:swift: SwiftLint` step runs green on this PR's build.

N/A — build tooling only. No UI or runtime behavior changes.


ios/Sources/GutenbergKit/Sources/Views/HTMLPreview/HTMLWebViewRenderer.swift
```diff
@@ -84,6 +84,12 @@ final class HTMLWebViewRenderer {
     @MainActor
     private class PooledWebView {
         let webView: WKWebView
+        // Deliberately strong: `PooledWebView` is the delegate's sole owner, and
+        // `RenderDelegate` holds no reference back, so there is no retain cycle.
+        // `WKWebView.navigationDelegate` is itself weak, so a weak reference here
+        // would deallocate the delegate immediately after `init` and rendering
+        // would never complete.
+        // swiftlint:disable:next weak_delegate
         let delegate: RenderDelegate
```
dcalhoun, line 93: I perceive this as sound, but welcome additional validation of this.

.swiftlint.yml
dcalhoun, line 1: Inspired by WordPress-iOS' [configuration](https://github.com/wordpress-mobile/WordPress-iOS/blob/9a66f0d5833d7cc17955f490c552f1861279cdb9/.swiftlint.yml).

bin/claude-hooks/swiftlint.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/sh
```
mokagio: We've been recommending a different shebang recently.


```suggestion
#!/usr/bin/env bash

```

bin/claude-hooks/swiftlint.sh
```diff
@@ -0,0 +1,62 @@
...
+# Claude Code PostToolUse hook: lint the just-edited Swift file with SwiftLint.
+#
+# Reads the hook payload (JSON) on stdin, extracts the file path, and runs it
+# through `make lint-swift` so the hook uses the exact same SwiftLint binary and
+# configuration as the project lint command. On violations, prints them to
+# stderr and exits 2 so Claude Code feeds them back to the model for
+# self-correction.
```
mokagio, line 9: Have you considered running SwiftLint on the whole codebase? It should be almost as fast and would simplify the script a bit. It might generate some out of scope changes from time to time, but the occurrence should be low if it runs on every change by Claude + in CI.

Just an idea. A strong case could be made for only running on modified files being conceptually cleaner.

.buildkite/pipeline.yml
```diff
@@ -90,6 +90,11 @@ steps:
           queue: android
       plugins: *plugins
 
+    - label: ':swift: Lint Swift'
+      key: lint-swift
+      command: make lint-swift
+      plugins: *plugins
```
mokagio: We have a dedicated linter agent that can run SwiftLint and is optimized to be fast and safe.

```suggestion
    - label: ":swift: SwiftLint"
      command: swiftlint
      agents:
        queue: linter
```

mokagio (approved):
Thanks for setting this up @dcalhoun ! Love a good linting.

bin/claude-hooks/swiftlint.sh
dcalhoun: Makes sense. Thanks for the suggestion! Addressed in https://github.com/wordpress-mobile/GutenbergKit/pull/575/commits/4d5fa6ffede6c201faf9ca89c1a6b2d739190ec0.

PR #577, opened 2026-08-02
feat: give the native inserter button platform-appropriate press feedback

dcalhoun:
## What?

Gives the native inserter button press feedback, and replaces its iOS-only styling on Android with Material 3's shape and motion.

## Why?

Tapping the button produced no visual confirmation that it registered. Touch devices have no hover state, so the pressed state is the only affordance available.

The button's circular container and its motion were authored for iOS and applied unconditionally, so on Android the shape read as out of place against the platform's shape language.

## How?

Split into reviewable commits:

- **Colocate styles** — move the button's rules from `editor-toolbar/style.scss` into a new `native-inserter/style.scss`. Pure move, no visual change.
- **Press feedback** — the icon's background darkens while pressed.
- **Android shape and motion** — a rounded square that morphs to full-round while pressed, per Material 3's icon button, using `EasingStandard` and `DurationShort3`.
- **iOS motion** — the button grows on press, with a `linear()` easing sampled from SwiftUI's `.snappy` spring. Small controls grow rather than shrink so their edges stay visible around the finger.
- **Demo default** — the Android demo app enables the native inserter by default, matching iOS. The `EditorConfiguration` default stays `false`, so host apps are unaffected.

Only static appearance is shared between platforms; each declares its own complete transition list. Non-Android rules are scoped `:not(.is-android)` rather than `is-ios`, since neither class is set in a browser during development.

`linear()` requires WKWebView 17.2 while `Package.swift` deploys to iOS 17.0, so an eased fallback precedes the `@supports` block.

## Testing Instructions

1. Run the Android demo app. The native inserter is now enabled by default.
2. Press and hold the add block (`+`) button in the editor toolbar.
3. Confirm the container darkens immediately, grows slightly, and its corners morph from a rounded square to a circle.
4. Release and confirm it settles back smoothly.
5. Repeat on the iOS demo app. The button should stay circular, growing and fading in on press, settling with a spring-like curve.

No change to the button's semantics, accessible name, or focus behavior — these styles are purely visual and apply on `:active`. The existing "Add block" label and 46px minimum touch target are unchanged.

## Screenshots or screencast

Android | iOS
-- | -- 
<video src="https://github.com/user-attachments/assets/a4dc63aa-f8ad-4d9a-b122-d9a103dd8db3" /> | <video src="https://github.com/user-attachments/assets/19f57bf8-acd0-4bda-b52b-5ad44ff57fad" />

### DIFF
--- a/android/app/src/androidTest/java/com/example/gutenbergkit/EditorInteractionTest.kt
+++ b/android/app/src/androidTest/java/com/example/gutenbergkit/EditorInteractionTest.kt
@@ -69,7 +69,7 @@ class EditorInteractionTest {
         EditorTestHelpers.waitForEnabled(composeTestRule, "Undo")
 
         // Insert a Paragraph block and type in the content area.
-        EditorTestHelpers.typeInContent("World")
+        EditorTestHelpers.typeInContent("World", composeTestRule)
 
         // Undo should still be enabled after typing content.
         EditorTestHelpers.waitForEnabled(composeTestRule, "Undo")
--- a/android/app/src/androidTest/java/com/example/gutenbergkit/EditorTestHelpers.kt
+++ b/android/app/src/androidTest/java/com/example/gutenbergkit/EditorTestHelpers.kt
@@ -1,7 +1,10 @@
 package com.example.gutenbergkit
 
+import androidx.compose.ui.test.SemanticsMatcher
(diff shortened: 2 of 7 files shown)


nbradbury (approved):
Looks good! :shipit:

PR #580, opened 2026-08-06
refactor(ios): use wordpress-rs connectivity error helpers

### LINKED ISSUE #578: Lower connectivity error classification into wordpress-rs

The iOS demo app hand-rolls two error classifiers in `ios/Demo-iOS/Sources/Views/SitePreparationView.swift`:

- `isUnreachableSiteError` (line 332) — matches `.nonExistentSiteError`, used to show wp-env guidance when the local WordPress site can't be reached
- `isNetworkError` (line 340) — matches `.deviceIsOfflineError`, used to fall back to an offline editor configuration

Both pattern-match `WpApiError.RequestExecutionFailed` with five positional associated values to read a single nested reason case. That's fragile — it breaks whenever `WpApiError` gains a case — and it's knowledge that belongs to wordpress-rs rather than to a consumer.

wordpress-rs _creates_ this distinction and then makes consumers rebuild it. `WpRequestExecutor.perform` maps `URLError.cannotFindHost` / `.cannotConnectToHost` / `.dnsLookupFailed` → `.nonExistentSiteError`, and `.notConnectedToInternet` / `.networkConnectionLost` → `.deviceIsOfflineError`. Consumers have to know that mapping exists and match it positionally.

Raised in [review feedback on #572](https://github.com/wordpress-mobile/GutenbergKit/pull/572#discussion_r3706841604): "We should probably lower this into wprs on either the Swift or Rust layers."

### The Android demo app has the same need, but can't express it

`SitePreparationViewModel.loadPostTypes` discards the failure reason entirely:

```kotlin
if (result !is WpRequestResult.Success) return null
```

The caller then substitutes a hardcoded `post` type (`loadPostTypes(config) ?: listOf(PostTypeDetails.post)`), so every failure — device offline, site unreachable, HTTP 403, response parsing error — collapses into the same silent fallback.

This is why the plain-permalink bug fixed in #572 went unnoticed on Android: iOS surfaced a hard error, while Android quietly showed a post-type picker containing only "Posts". Android also handles offline mode proactively via `networkAvailabilityProvider` rather than reacting to a failed request, so it has no equivalent of the iOS offline fallback path.

### 1. Add the classification to wordpress-rs (Rust layer)

Two additions to `WpApiError` in `wp_api/src/api_error.rs`:

- `isSiteUnreachable` → `.nonExistentSiteError`
- `isDeviceOffline` → `.deviceIsOfflineError`

Implemented in the **Rust layer** rather than as a Swift-only extension, so both platforms share one implementation. A Swift-only extension could not serve the Android work below, which would force a second hand-written copy of the same pattern match — the exact duplication this issue exists to remove.

Shape follows the existing `localize_wp_api_error` precedent (generated by `WpDeriveLocalizable` in `wp_localization_macro`): inherent methods on `WpApiError` for Rust callers, plus thin `#[uniffi::export]` free functions taking `&WpApiError`, wrapped in ergonomic Swift/Kotlin extensions. `WpApiError` is a `uniffi::Error` with rich associated values, so this route is preferred over an exported `impl` block. Swift wrapper mirrors the existing `WpApiError.isCancellationError` (`native/swift/Sources/wordpress-api/Extensions.swift`).

Additive; no breaking change.

### 2. Refactor the iOS demo app

Once a wordpress-rs release carrying the above is available, replace both helpers with the library properties.

The `error is URLError` fallback in both helpers should be dropped rather than moved into the library — it only catches errors that never passed through `WpRequestExecutor`, and inside wordpress-rs the rewrapping is guaranteed. Worth verifying no call path in site preparation bypasses the client before removing it.

### 3. Align the Android demo app

Done alongside the iOS refactor, not as separate work. Stop discarding the failure reason in `loadPostTypes` and use the new classification so Android can distinguish "site unreachable" from "device offline" from "request failed for another reason", rather than collapsing all of them into a hardcoded `post` type.

This closes the behavioural gap that hid the #572 bug on Android and brings the two demo apps to parity.

## Notes

Blocked on a wordpress-rs release. The `[Unreleased]` section there currently carries a breaking `ProductType` change, so release timing is coordinated with the wordpress-rs maintainers rather than driven by this issue.

dcalhoun:
## What?

Replaces the iOS demo app's hand-rolled connectivity error classifiers with the equivalents added to wordpress-rs.

Closes #578.

## Why?

`SitePreparationView` matched `WpApiError.RequestExecutionFailed` with five positional associated values to read a single nested reason — once for "site unreachable", once for "device offline". That is fragile: it breaks whenever `WpApiError` gains a case, and it duplicates knowledge that belongs to wordpress-rs, which creates the distinction in the first place.

Raised in [review feedback on #572](https://github.com/wordpress-mobile/GutenbergKit/pull/572#discussion_r3706841604): "We should probably lower this into wprs on either the Swift or Rust layers."

## How?

Both helpers are deleted and their call sites read the library property directly:

```swift
} catch let error as WpApiError where error.isSiteUnreachable {
```

A typed `catch` does the cast, so no wrapper is needed — and the demo app no longer defines its own vocabulary for a distinction the library already names.

The `error is URLError` fallback goes with them. Every throwing call in the enclosing `do` blocks goes through `WordPressAPI` — `apiRoot.get()` and `postTypes.listWithEditContext()` — so `WpRequestExecutor` has already rewrapped any `URLError` into a `WpApiError` by the time these run. The only remaining `URLSession` reference in the file is the `URLSessionConfiguration` passed _into_ `WordPressAPI`, so the fallback could never fire.

Depends on [Automattic/wordpress-rs#1488](https://github.com/Automattic/wordpress-rs/pull/1488), which adds `isSiteUnreachable` / `isDeviceOffline`.

### Follow-up: Android

#578 also covers bringing the Android demo app to parity — it currently discards the failure reason in `loadPostTypes` and substitutes a hardcoded `post` type, which is why the plain-permalink bug fixed in #572 went unnoticed there. That work is blocked on a wordpress-rs mapping inconsistency: the Kotlin executor maps a refused connection to `HttpError` while Swift maps it to `NonExistentSiteError`, so `isSiteUnreachable` currently answers differently per platform. Details in the "Known limitation" section of #1488.

## Testing Instructions

1. Open `ios/Demo-iOS/Gutenberg.xcodeproj` and let package resolution finish.
2. **Site unreachable** — with wp-env stopped and the network up, open the Local WordPress site. Expect the wp-env guidance: "Could not connect to Local WordPress at localhost:8888…".
3. **Device offline** — with a saved account, enable Airplane Mode and open its Editor Configuration. Expect the screen to load with the offline fallback applied: Network Fallback shows "Automatic", Post Type shows "Post", and Site URL shows the stored API root.
4. **The two are distinct** — with the network up, block the site's host in `/etc/hosts`, then open its Editor Configuration. Expect the error view with "A server with the specified hostname could not be found", _not_ the offline fallback. This confirms `isSiteUnreachable` and `isDeviceOffline` return opposite answers for the same error.

Verified on device: all three behave as described, and a full `xcodebuild` of the demo app succeeds.

N/A — no user interface changes. The error and fallback paths are unchanged; only how they are detected differs.

### DIFF
--- a/ios/Demo-iOS/Gutenberg.xcodeproj/project.pbxproj
+++ b/ios/Demo-iOS/Gutenberg.xcodeproj/project.pbxproj
@@ -573,8 +573,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/Automattic/wordpress-rs";
 			requirement = {
-				kind = revision;
-				revision = d70c99e1ff38772353bd8f6a81879320990a560e;
+				kind = exactVersion;
+				version = 0.7.0;
 			};
 		};
 /* End XCRemoteSwiftPackageReference section */
--- a/ios/Demo-iOS/Gutenberg.xcodeproj/project.pbxproj
+++ b/ios/Demo-iOS/Gutenberg.xcodeproj/project.pbxproj
@@ -573,8 +573,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/Automattic/wordpress-rs";
 			requirement = {
-				branch = "pr-build/1488";
-				kind = branch;
+				kind = revision;
+				revision = d70c99e1ff38772353bd8f6a81879320990a560e;
 			};
 		};
 /* End XCRemoteSwiftPackageReference section */
--- a/ios/Demo-iOS/Gutenberg.xcodeproj/project.pbxproj
+++ b/ios/Demo-iOS/Gutenberg.xcodeproj/project.pbxproj
@@ -573,8 +573,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/Automattic/wordpress-rs";
 			requirement = {
-				kind = exactVersion;
-				version = 0.6.0;
+				branch = "pr-build/1488";
+				kind = branch;
 			};
 		};
 /* End XCRemoteSwiftPackageReference section */
--- a/ios/Demo-iOS/Gutenberg.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ios/Demo-iOS/Gutenberg.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -24,8 +24,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/Automattic/wordpress-rs",
       "state" : {
-        "revision" : "2525759228b0b5ba0379671354399cf8ede8d0a3",
-        "version" : "0.6.0"
+        "branch" : "pr-build/1488",
+        "revision" : "290cfaca4171591b0d0579485e1ab8990536580f"
       }
     }
   ],
--- a/ios/Demo-iOS/Gutenberg.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ios/Demo-iOS/Gutenberg.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -24,8 +24,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/Automattic/wordpress-rs",
       "state" : {
-        "branch" : "pr-build/1488",
-        "revision" : "290cfaca4171591b0d0579485e1ab8990536580f"
+        "revision" : "d70c99e1ff38772353bd8f6a81879320990a560e"
       }
     }
   ],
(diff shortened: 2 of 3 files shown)


ios/Demo-iOS/Gutenberg.xcodeproj/project.pbxproj
```diff
@@ -573,8 +573,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/Automattic/wordpress-rs";
 			requirement = {
-				kind = exactVersion;
-				version = 0.6.0;
+				kind = revision;
+				revision = d70c99e1ff38772353bd8f6a81879320990a560e;
```
dcalhoun: Is it safe to use a wordpress-rs commit hash for the Demo app or do we need to await a tag?

jkmassel (approved):
We should wait for a wprs release, but this looks solid

PR #583, opened 2026-08-12
build(deps): clear 40 dev-dependency security advisories

jkmassel:
## What?

Bumps **15 dev-only dependencies** to clear **40 open Dependabot advisories** — 1 critical, most of the rest high — across the build, lint, test, and local-WP (`@wordpress/env`) tooling. **None of these packages ship in the iOS/Android app bundle.**

Dashboard: https://github.com/wordpress-mobile/GutenbergKit/security/dependabot

## Why?

The Security tab currently has 85 open alerts. This PR clears the 40 that are (a) in the **development** dependency tree and (b) resolvable with a dependency bump — including the lone **critical** ([`shell-quote`, alert 133](https://github.com/wordpress-mobile/GutenbergKit/security/dependabot/133)) and the `undici` SOCKS5 TLS-validation advisory that kicked this off ([alert 136](https://github.com/wordpress-mobile/GutenbergKit/security/dependabot/136)). The other 45 alerts need a major upgrade or are runtime-scoped — see **What's not addressed**.

## How?

Two mechanisms. Alert numbers below are Dependabot alert IDs (navigate via the dashboard link above).

### 1. In-range bumps
The declared semver ranges already permitted the patched versions, so `npm update` — plus a floor bump for the one direct dep (`vite`) — was enough:

| Package | Bump | Alerts |
|---|---|---|
| `undici` | 7.24.7 → 7.29.0 | 136, 137, 139, 140, 141, 143, 180–184 |
| `vite` (direct; floor → `^8.0.16`) | 8.0.5 → 8.2.1 | 134, 135 |
| `shell-quote` | → 1.10.0 | **133 (critical)**, 165 |
| `fast-uri` | → 3.1.5 | 163, 164, 179, 206, 207 |
| `js-yaml` | → 3.15.1 / 4.3.1 | 157, 158, 167, 168, 200, 201 |
| `brace-expansion` | → 1.1.18 / 2.1.4 | 160, 166 |
| `immutable` | → 5.1.9 | 161, 162 |
| `fast-xml-builder` | → 1.3.0 | 122, 123 |
| `tmp` | → 0.2.7 | 131 |
| `simple-git` | → 3.36.0 | 119 |
| `@babel/plugin-transform-modules-systemjs` | → 7.29.8 | 126 |

### 2. Scoped `overrides`
Four packages were exact-pinned by a parent (`ws@"8.18.3"`, `minimatch@"9.0.3"`, …), so a leaf `npm update` couldn't move them. Each override is **scoped** so it can't clobber a legitimately-different major elsewhere in the tree:

| Override | Scope | Preserves | Alerts |
|---|---|---|---|
| `ws` → 8.21.0 | global, with an exception | `ws@7.5.13` under `react-devtools-core` (needs `^7`) | 138, 153 |
| `qs` → 6.15.2 | under `@wordpress/env` | — | 130 |
| `minimatch` → 9.0.9 | under `@typescript-eslint/typescript-estree` | `minimatch@3.1.5` under ESLint | 65 |
| `@babel/core` → 7.29.7 | under `@wordpress/babel-preset-default` | — | 154 |

### What's *not* addressed
- **`electron` (33 alerts) + its bundled `got` (alert 33)** — both come only through `react-devtools@7.0.1`, which is already the latest published version and still bundles `electron@^23`. npm's only offered "fix" is a **breaking downgrade** to `react-devtools@1.0.5`. Clearing these needs a separate effort to replace or drop `react-devtools`.
- **Runtime-scope advisories** — `seroval` (critical), `nanoid`, `postcss`, `showdown`, `uuid` (11 alerts). These ship to users, so they're out of scope for a dev-only bump and deserve a focused follow-up.

## CI fixes

Getting this PR green surfaced two failures in the **Test Web E2E** job, unrelated to the dependency bumps. Both are fixed here as separate commits — the second is a repo-wide breakage that also blocks `trunk`.

### `ci: skip electron's binary download`
The E2E job installs the full dep tree, and `electron`'s postinstall downloads a ~100 MB prebuilt binary (GitHub releases) that **no CI step runs** — it only backs `react-devtools`' `dev:tools`. That download flaked (`RequestError: socket hang up`) and failed the job. Set `ELECTRON_SKIP_BINARY_DOWNLOAD=1` in the pipeline `env:` block so every `npm ci` skips it; local `dev:tools` still pulls the binary it needs.

### `ci: source wp-env WordPress core from the WordPress.org zip`
With no `core` in `.wp-env.json`, wp-env resolves core to `WordPress/WordPress#<latest>` — a **git** source — and `git fetch`es the newest version tag from `api.wordpress.org`. At a WordPress release boundary the version API returns the new version (**7.0.4**) before the `WordPress/WordPress` git mirror is tagged, so the fetch fails with `couldn't find remote ref 7.0.4` and `wp-env start` dies — **on every branch, `trunk` included** (it only surfaced here once the electron fix let the job reach `wp-env start`). Pointing `core` at `https://wordpress.org/latest.zip` — served immediately on release — keeps tracking latest while sidestepping the git-mirror lag. Verified locally: core resolves to a `zip` source (no git ref) and the playground runtime boots in ~12s.

## Testing Instructions

- [x] `npm run test:unit` — 202 passing
- [x] `npm run build` — clean production build on `vite@8.2.1`
- [x] `wp-env start --runtime=playground` boots locally with the zip core (~12s)
- [ ] CI green — the electron skip lets the E2E `npm ci` finish; the zip core lets `wp-env start` succeed
- [ ] After merge, confirm Dependabot re-scans `trunk` and auto-closes the 40 listed alerts

N/A — dependency and CI-config change; no UI or app-runtime behavior is affected.

## Screenshots or screencast

N/A

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -3,6 +3,10 @@ agents:
 
 env:
     IMAGE_ID: $IMAGE_ID
+    # electron ships a ~100 MB prebuilt binary via a postinstall download
+    # that no CI step ever runs (it only backs react-devtools' `dev:tools`).
+    # The download is flaky from CI, so skip it -- see the Test Web E2E job.
+    ELECTRON_SKIP_BINARY_DOWNLOAD: '1'
 
 steps:
     - label: ':react: Build React App'
--- a/package-lock.json
+++ b/package-lock.json
@@ -22073,9 +22073,9 @@
 			}
 		},
 		"node_modules/undici": {
-			"version": "7.24.7",
-			"resolved": "https://registry.npmjs.org/undici/-/undici-7.24.7.tgz",
-			"integrity": "sha512-H/nlJ/h0ggGC+uRL3ovD+G0i4bqhvsDOpbDv7At5eFLlj2b41L8QliGbnl2H7SnDiYhENphh1tQFJZf+MyfLsQ==",
+			"version": "7.29.0",
+			"resolved": "https://registry.npmjs.org/undici/-/undici-7.29.0.tgz",
+			"integrity": "sha512-IDxfleLmmbSskfWSUATiN1nfn2rDuvnMOqb5CWR92iIfojA0Ud+ulOAAEQ57LPr9rWmsreUyf5lwyao+7GNNVw==",
 			"dev": true,
 			"license": "MIT",
 			"engines": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -91,7 +91,7 @@
 				"react-devtools": "^7.0.1",
 				"react-devtools-core": "^7.0.1",
 				"sass-embedded": "^1.98.0",
-				"vite": "^8.0.5",
+				"vite": "^8.0.16",
 				"vitest": "^4.1.0"
 			}
 		},
@@ -101,20 +101,6 @@
 			"integrity": "sha512-baYZExFpsdkBNuvGKTKWCwKH57HRZLVtycZS05WTQNVOiXVSeAki3nU35zlRbToeMW8aHlJfyS+1C4BOv27q0A==",
 			"dev": true,
 			"license": "MIT"
-		},
-		"node_modules/@ampproject/remapping": {
-			"version": "2.3.0",
-			"resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
-			"integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
-			"dev": true,
-			"license": "Apache-2.0",
-			"dependencies": {
-				"@jridgewell/gen-mapping": "^0.3.5",
-				"@jridgewell/trace-mapping": "^0.3.24"
-			},
-			"engines": {
-				"node": ">=6.0.0"
-			}
 		},
 		"node_modules/@ariakit/core": {
 			"version": "0.4.18",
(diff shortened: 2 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -3,6 +3,10 @@ agents:
 
 env:
     IMAGE_ID: $IMAGE_ID
+    # electron ships a ~100 MB prebuilt binary via a postinstall download
+    # that no CI step ever runs (it only backs react-devtools' `dev:tools`).
+    # The download is flaky from CI, so skip it -- see the Test Web E2E job.
+    ELECTRON_SKIP_BINARY_DOWNLOAD: '1'
```
mokagio, line 9: TIL. Neat.

package-lock.json
```diff
@@ -17803,9 +17750,9 @@
 			"license": "MIT"
 		},
 		"node_modules/nanoid": {
-			"version": "3.3.11",
-			"resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.3.11.tgz",
-			"integrity": "sha512-N8SpfPUnUp1bK+PMYW8qSWdl9U+wwNWI4QKxOYDy9JAro3WMX7p2OeVRF9v+347pnakNevPmiHhNmZ2HbFA76w==",
+			"version": "3.3.18",
+			"resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.3.18.tgz",
+			"integrity": "sha512-DTg4MJbGMWkfi6VZFdNt2/caMbQy4Ou+Op/hJQvGEWcnVfoA1QA+xzRKAzw9jD6+GVOOeYr/mIcuDSdug6F6+w==",
```
mokagio, line 17755: > Runtime-scope advisories — `seroval` (critical), `nanoid`, `postcss`, `showdown`, `uuid` (11 alerts). These ship to users, so they're out of scope for a dev-only bump and deserve a focused follow-up.

Looks like `nanoid` and `postcss` (see below) ended up being touched. Does it matter? They're just patches so I think it should be fine without explicitly checking the behavior in the clients.

mokagio (approved):
LGTM. I'd never seen ovverrides in `npm`.

package-lock.json
jkmassel: Interesting that it's part of the shipping bundle – I don't think it'll be a problem, but we can keep an eye out for CSS weirdness?

PR #588, opened 2026-08-18
ci(release): generate release notes against an explicit previous tag

dcalhoun:
## What?

Generated release notes restated every PR back to `v0.16.0` on every release since `v0.17.1`. This makes the release lane pass an explicit previous tag instead of letting GitHub infer one.

The already-published notes have been regenerated separately (see below), so this PR is what stops the next release from reintroducing the problem.

## Why?

Every GitHub Release from `v0.17.1` through `v0.20.0-alpha.0` listed months of already-shipped work — `v0.20.0-alpha.0` showed 24 PRs where only 3 were new — and all of them reported `compare/v0.16.0...`.

GitHub picks the notes base by walking tags newest-to-oldest and taking the first whose commit is an **ancestor** of the one being released. Our release tags never satisfy that. `publish_release_to_github` tags a `Package.swift` rewrite committed on a local `release/vX.Y.Z` branch that is never pushed, so no release tag is reachable from any other:

```
v0.16.0           -> ON trunk       ← lightweight tag predating this flow; the only candidate left
v0.17.1 … v0.19.0 -> NOT on trunk
```

GitHub falls back to `v0.16.0` and re-lists everything since. Releases before `v0.17.1` were unaffected — they predate this tagging flow, so the inference worked.

The off-trunk tag placement is deliberate and documented — it's what lets the tag carry a `Package.swift` pointing at the prebuilt XCFramework while `trunk` keeps `.local`. This PR leaves that design alone and fixes the notes instead.

## How?

`set_github_release` can't express `previous_tag_name` (it only forwards a `generate_release_notes` boolean), so `generated_release_notes` calls the `releases/generate-notes` endpoint directly and passes the result through as `description`.

`previous_release_tag` resolves the base as the most recent stable release older than the version being published. Prereleases are skipped as candidates, matching GitHub's default — a stable release reports everything since the last stable release, including work already listed in its own alphas.

It reads the Releases API rather than local tags: CI checkouts may not have fetched every tag, the API reports `prerelease` authoritatively rather than us inferring it from the tag name, and it ignores stray tags never published as releases (`vtest-s3-xcframework-*`).

Resolution fails loudly if no base is found. A silently wrong base looks like a successful release and is only caught by someone reading the release page later.

`validate` also resolves and logs the base. That step runs before anything is published, so a wrong base surfaces while a re-run is still free. It recomputes rather than threading the value through, because Buildkite runs `validate` as a separate step — separate agent, separate process, no shared `lane_context`.

## Testing Instructions

It's not possible to test this without publishing a new release. We'll just need to verify the release notes of the next few releases when they occur. 

## Already-published notes

Regenerated ahead of this PR, oldest first, so the history reads correctly today rather than waiting on a merge:

| Release | Base was | Base now | PRs |
|---|---|---|---|
| `v0.17.1` | `v0.16.0` | unchanged | 5 — was already correct |
| `v0.17.2` | `v0.16.0` | `v0.17.1` | 7 → 2 |
| `v0.18.0` | `v0.16.0` | `v0.17.2` | 10 → 2 |
| `v0.18.1` | `v0.16.0` | `v0.18.0` | 10 → 0 |
| `v0.19.0-alpha.0` | `v0.16.0` | `v0.18.1` | 22 → 12 |
| `v0.19.0` | `v0.16.0` | `v0.18.1` | 22 → 12 |
| `v0.20.0-alpha.0` | `v0.16.0` | `v0.19.0` | 24 → 3 |

Only the release `body` changed — tags, assets, checksums, and prerelease flags are untouched, so nothing SPM consumers resolve is affected.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -317,10 +317,9 @@ end
 # previous tag. `set_github_release`'s `is_generate_release_notes` cannot express
 # one.
 #
-# Uses `GithubHelper` rather than the `get_prs_between_tags` action that wraps it:
-# the action rescues every `StandardError` and returns the message as the
-# changelog, publishing a release whose notes read "❌ Error computing the list of
-# PRs…". The tag is already pushed by this point, so this must fail the step.
+# Uses `GithubHelper` rather than the `get_prs_between_tags` action that wraps it
+# until https://github.com/wordpress-mobile/release-toolkit/pull/772 gets ships in the next release-toolkit
+# version (at which point we'll be able to use the action and its new `fail_on_error: true` parameter)
 def generated_release_notes(version:, token:)
   Fastlane::Helper::GithubHelper.new(github_token: token).generate_release_notes(
     repository: GITHUB_REPO,
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -290,11 +290,13 @@ def previous_release_tag(version:, token:)
   candidates = releases.reject { |release| release['draft'] || release['prerelease'] }
                        .map { |release| release['tag_name'] }
                        .reject { |tag| tag == version }
-                       .select { |tag| tag =~ /\Av\d+\.\d+\.\d+\z/ }
+                       .grep(/\Av\d+\.\d+\.\d+\z/)
+                       .map { |tag| [tag, Gem::Version.new(tag.delete_prefix('v'))] }
 
   target = Gem::Version.new(version.delete_prefix('v').split('-').first)
-  candidates.select { |tag| Gem::Version.new(tag.delete_prefix('v')) < target }
-            .max_by { |tag| Gem::Version.new(tag.delete_prefix('v')) }
+  candidates.select { |_tag, tag_version| tag_version < target }
+            .max_by { |_tag, tag_version| tag_version }
+            &.first
 end
 
 # `previous_release_tag`, erroring instead of returning nil.
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -122,8 +122,7 @@ lane :validate do |options|
   # Resolve the release-notes base now, while nothing has been published yet.
   # A wrong base produces notes that restate old releases — a silent failure
   # once the release is live, but a cheap re-run if it surfaces here.
-  previous_tag = previous_release_tag(version: version, token: token)
-  UI.user_error!("Could not resolve a previous release tag for #{version}.") if previous_tag.nil?
+  previous_tag = previous_release_tag!(version: version, token: token)
   UI.success("Release notes for #{version} will be generated against #{previous_tag}.")
 
   # Clear lane-context values populated by the probe calls above so a later
@@ -298,32 +297,34 @@ def previous_release_tag(version:, token:)
             .max_by { |tag| Gem::Version.new(tag.delete_prefix('v')) }
 end
 
-# Build the release body via GitHub's notes generator, pinned to an explicit
-# previous tag. `set_github_release`'s `is_generate_release_notes` cannot express
-# `previous_tag_name`, so call the endpoint directly and pass the result through
-# as `description`.
+# `previous_release_tag`, erroring instead of returning nil.
 #
-# Fails loudly rather than falling back to auto-generated notes: a wrong base
-# looks like a successful release, caught only by someone reading the page later.
-def generated_release_notes(version:, token:)
+# Falling back to GitHub's inference would restate every release back to
+# `v0.16.0` — a silent failure that looks like a successful publish and is only
+# caught by someone reading the release page later.
+def previous_release_tag!(version:, token:)
   previous_tag = previous_release_tag(version: version, token: token)
   UI.user_error!("Could not resolve a previous release tag for #{version}; refusing to publish notes that may restate old releases.") \
     if previous_tag.nil?
 
   UI.message("Generating release notes for #{version} against previous tag #{previous_tag}.")
+  previous_tag
+end
 
-  response = github_api(
-    api_token: token,
-    http_method: 'POST',
-    path: "/repos/#{GITHUB_REPO}/releases/generate-notes",
-    body: { tag_name: version, previous_tag_name: previous_tag }
+# Build the release body via GitHub's notes generator, pinned to an explicit
+# previous tag. `set_github_release`'s `is_generate_release_notes` cannot express
+# one.
+#
+# Uses `GithubHelper` rather than the `get_prs_between_tags` action that wraps it:
+# the action rescues every `StandardError` and returns the message as the
+# changelog, publishing a release whose notes read "❌ Error computing the list of
+# PRs…". The tag is already pushed by this point, so this must fail the step.
+def generated_release_notes(version:, token:)
+  Fastlane::Helper::GithubHelper.new(github_token: token).generate_release_notes(
+    repository: GITHUB_REPO,
+    tag_name: version,
+    previous_tag: previous_release_tag!(version: version, token: token)
   )
-
-  notes = response[:json]['body']
-  UI.user_error!("GitHub returned empty release notes for #{version} (status #{response[:status]}).") \
-    if notes.nil? || notes.strip.empty?
-
-  notes
 end
 
 def require_env_vars!(*keys)
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -119,8 +119,15 @@ lane :validate do |options|
 
   UI.user_error!("Release #{version} already exists on GitHub.") unless release.nil?
 
-  # Clear lane-context values populated by `get_github_release` so a later
-  # action doesn't see stale state from this probe call.
+  # Resolve the release-notes base now, while nothing has been published yet.
+  # A wrong base produces notes that restate old releases — a silent failure
+  # once the release is live, but a cheap re-run if it surfaces here.
+  previous_tag = previous_release_tag(version: version, token: token)
+  UI.user_error!("Could not resolve a previous release tag for #{version}.") if previous_tag.nil?
+  UI.success("Release notes for #{version} will be generated against #{previous_tag}.")
+
+  # Clear lane-context values populated by the probe calls above so a later
+  # action doesn't see stale state from them.
   [
     SharedValues::GITHUB_API_RESPONSE,
     SharedValues::GITHUB_API_STATUS_CODE,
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -252,6 +262,70 @@ def github_token!(options = {})
...
+# months of merged PRs. An explicit `previous_tag_name` sidesteps it.
+#
+# Prereleases are excluded as candidates, matching GitHub's default: a stable
+# release reports everything since the last stable one, including work already
+# listed in intervening alphas.
+#
+# Reads the Releases API rather than local tags: CI checkouts may not have
+# fetched every tag, the API reports `prerelease` authoritatively, and it ignores
+# stray tags never published as releases (e.g. `vtest-s3-xcframework-*`).
+def previous_release_tag(version:, token:)
```
AliSoftware, line 279: ⚠️  [We have an action in `release-toolkit` for that already](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/find_previous_tag.rb)

So no need to re-implement via API calls in your `Fastfile`.

fastlane/Fastfile
```diff
@@ -252,6 +262,70 @@ def github_token!(options = {})
...
+end
+
+# Build the release body via GitHub's notes generator, pinned to an explicit
+# previous tag. `set_github_release`'s `is_generate_release_notes` cannot express
+# `previous_tag_name`, so call the endpoint directly and pass the result through
+# as `description`.
+#
+# Fails loudly rather than falling back to auto-generated notes: a wrong base
+# looks like a successful release, caught only by someone reading the page later.
+def generated_release_notes(version:, token:)
```
AliSoftware: ⚠️  [We also already have a helper for that in `release-toolkit`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/get_prs_between_tags.rb#L43-L71) so no need to re-implement it in your `Fastfile` either.

AliSoftware (changes requested):
Most of this is already available via some actions we expose in `release-toolkit`, so no need to reinvent the wheel 😉

fastlane/Fastfile
dcalhoun: Thanks for the guidance! 🙇🏻‍♂️ 

Unfortunately, it seems `find_previous_tag` may not satisfy the need here. It uses `git describe --tags --abbrev=0`, which only finds tags **reachable from the current commit**—the same constraint that breaks GitHub's own inference. Our release tags are each committed on a local `release/vX.Y.Z` branch that's never pushed, so no release tag is an ancestor of any other and it always falls back to the last tag on `trunk`:

```
RELEASE            find_previous_tag   previous_release_tag
v0.17.2            v0.16.0             v0.17.1
v0.18.0            v0.16.0             v0.17.2
v0.18.1            v0.16.0             v0.18.0
v0.19.0            v0.16.0             v0.18.1
v0.20.0-alpha.0    v0.16.0             v0.19.0
```

For `v0.20.0-alpha.0` that's 25 PRs vs. the 3 that were actually new.

`previous_release_tag` asks a different question ("highest stable release below this version?"), which needs no ancestry. Reading the Releases API rather than local tags also means CI doesn't need every tag fetched, `prerelease` is authoritative rather than inferred from the name, and stray tags like `vtest-s3-xcframework-*` are ignored.

WYDT?

fastlane/Fastfile
dcalhoun: Good call. I dropped the custom API call in c608c5e4.

I went with `GithubHelper#generate_release_notes`—the helper that action wraps—rather than the action itself. `get_prs_between_tags` rescues every `StandardError` and returns the message as the changelog, so a transient API failure would publish a release whose notes read `❌ Error computing the list of PRs since v0.19.0: …`. Since we push the tag before creating the Release, that needs to fail the step rather than land in the release body.

I suppose we could consider adding a `fail_on_error:` opt-out to the upstream action.

AliSoftware: Good catch.

I think adding a fail_on_error upstream would actually benefit all apps and is probably better approach if you're up for it!

fastlane/Fastfile
```diff
@@ -252,6 +261,72 @@ def github_token!(options = {})
...
+  )[:json]
+
+  candidates = releases.reject { |release| release['draft'] || release['prerelease'] }
+                       .map { |release| release['tag_name'] }
+                       .reject { |tag| tag == version }
+                       .select { |tag| tag =~ /\Av\d+\.\d+\.\d+\z/ }
+
+  target = Gem::Version.new(version.delete_prefix('v').split('-').first)
+  candidates.select { |tag| Gem::Version.new(tag.delete_prefix('v')) < target }
+            .max_by { |tag| Gem::Version.new(tag.delete_prefix('v')) }
```
AliSoftware: I haven't thought about this deeply, but I wonder how it would behave (and how we _want_ it to behave) for cases of hotfixes 🤔

Say:
- On week 1, we publish 12.3
- On week 3, we publish 13.0
- On week 4, we publish 12.3.1 hotfix
- On week 5, we publish 13.1
- On week 6, we publish 13.1.1
- On week 7, we publish 13.2

Then:
- Should the 12.3.1 hotfix use 12.3 or 13.0 as previous tag?
- Should 13.1 use 12.3.1 or 13.0 as previous tag?
- Shoild 13.2 use 13.1.1 or 13.1 as previous tag?

I _think_ you cover it here by only considering versions before current as candidates (so 13.0 won't be a candidate for 12.3.1, and while 12.3.1 will be a candidate for 13.1 because the ts `< target` it'll be shadowed by 13.0 having a higher version.

But that will still make 13.2 use 13.1.1 as previous tag… which _might_ be what we want, but maybe  it (maybe it'd make more sense to pick 13.1 as latest non-hotfix instead?) 🤔🤷‍♂️

Anyway, since I'm trying to think of edge cases I wanted us to double-check such possible edge cases we might not have thought of at first (especially because it's too late for me to think clearly enough about those right now 😅😴)

AliSoftware (commented):
I'll try and take a closer look at it on my tomorrow. But ata glance I think it makes indeed sense to use the gh_api for finding last release's tag given those are not ancestors

fastlane/Fastfile
AliSoftware: Not tested, but I think this small refactor might work the same while being a bit more compact (and avoiding to redo the `Gem::Version.new(tag.delete_prefix('v')` twice)

```suggestion
  candidates = releases.reject { |release| release['draft'] || release['prerelease'] }
                       .map { |release| release['tag_name'] }
                       .reject { |tag| tag == version }
                       .select { |tag| tag =~ /\Av\d+\.\d+\.\d+\z/ }
                       .map { |tag| Gem::Version.new(tag.delete_prefix('v')) }

  target = Gem::Version.new(version.delete_prefix('v').split('-').first)
  candidates.select { |tag_version| tag_version < target }.max
```

AliSoftware (commented):
I still think it'd be nicer to add a `fail_on_error:` parameter to the `generate_release_notes` action in `release-toolkit` so we can use that action directly instead of the Helper.

Otherwise, the logic LGTM, and I'm ok with using the `github_api` directly to get the latest stable release (I simply suggested an implementation improvement)

fastlane/Fastfile
dcalhoun: Thanks for the nudge. As you noted, I implemented `fail_on_error:` in https://github.com/wordpress-mobile/release-toolkit/pull/772. 

I deferred relying on this new parameter until a new `release-toolkit` release is published. Tracked in https://github.com/wordpress-mobile/GutenbergKit/issues/591. I suggest we merge the current implementation to avoid incorrect GBK release notes now.

fastlane/Fastfile
dcalhoun: Implemented this spirit of this (deduplication) in 223f319983b0afce20e9e960339ae8cd2845f7ae. Claude adjusted the implementation to address an issue flagged with the suggestion that resulted in 400 responses from the GitHub API due to the logic returning an incorrect tag name (e.g., `0.19.0` instead of `v0.19.0`).

dcalhoun: Thank you for thinking through this so thoroughly. I agree with your `v12.3.1` and `v13.1.0` choices. I believe `v13.2.0` should choose `v13.1.1` as the latest stable release. 

I had Claude verify our current logic achieves this, and it claimed it will. If future issues arise in release notes, we can fix them and adjust this logic accordingly.

AliSoftware: Ah good catch, we max by the `Gem::Version` but we want to return the tag name not the parsed version 👍

PS: Technically I think we could have used `max_by(&:last)` as a shorter version instead of `.max_by { |_tag, tag_version| tag_version }`… but that's probably too nitpicky at this point 😉 😜

fastlane/Fastfile
```diff
@@ -252,6 +261,74 @@ def github_token!(options = {})
...
+end
+
+# Build the release body via GitHub's notes generator, pinned to an explicit
+# previous tag. `set_github_release`'s `is_generate_release_notes` cannot express
+# one.
+#
+# Uses `GithubHelper` rather than the `get_prs_between_tags` action that wraps it:
+# the action rescues every `StandardError` and returns the message as the
+# changelog, publishing a release whose notes read "❌ Error computing the list of
+# PRs…". The tag is already pushed by this point, so this must fail the step.
```
AliSoftware:
```suggestion
# Uses `GithubHelper` rather than the `get_prs_between_tags` action that wraps it
# until https://github.com/wordpress-mobile/release-toolkit/pull/772 gets ships in the next release-toolkit
# version (at which point we'll be able to use the action and its new `fail_on_error: true` parameter)
```

AliSoftware (approved):

fastlane/Fastfile
AliSoftware: cc @iangmaia if you're the one who will do the next `release-toolkit` version (possibly soon after [your PR 771](https://github.com/wordpress-mobile/release-toolkit/pull/771) gets approved and merged?), can I then ask you to handle the follow-up on GutenbergKit here to adopt https://github.com/wordpress-mobile/release-toolkit/pull/772 that would have shipped with it too?

iangmaia: Sure 👍

AliSoftware: Connecting the dots: now done via https://github.com/wordpress-mobile/GutenbergKit/pull/607

PR #592, opened 2026-08-20
build: fix `make` output formatting

dcalhoun:
## What?

Two unrelated `make` output annoyances found while working on the release script.

## Why?

`make release VERSION_TYPE=patch` printed 23 lines of usage text before doing anything, on every valid invocation.

`make help` hardcoded its column width at 25 characters, so `build-resources-xcframework` (27) overflowed and `test-android-library-e2e` (24) left a single space — both descriptions fell out of alignment.

## How?

Moved the usage text into the `VERSION_TYPE`-is-empty branch. Invalid (non-empty) values still fall through to `bin/release.sh`, which validates them properly.

`make help` now measures the longest target name and pads to it, so a future longer target won't reintroduce the misalignment.

## Testing Instructions

1. `make release` → usage text, exit 1.
2. `make release VERSION_TYPE=patch DRY_RUN=true` → no usage dump.
3. `make help` → descriptions align in one column.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,9 @@ help: ## Display this help menu
 	@echo ""
 	@echo "Available targets:"
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## ' $(MAKEFILE_LIST) | \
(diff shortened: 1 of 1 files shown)


adalpari (approved):
LGTM!

PR #593, opened 2026-08-20
ci(release): trigger the publish build via the Buildkite CLI

dcalhoun:
## What?

Step 2 of the release process required opening a Buildkite form and pasting in a commit SHA and version number by hand. The release script now triggers that build itself when the Buildkite CLI (`bk`) is available, printing the build URL.

## Why?

Pasting a SHA into a web form is easy to get wrong, and getting it wrong tags the wrong commit. The script already knows the SHA it just pushed.

## How?

`bk build create` takes the same three inputs the form does:

```bash
bk build create -p automattic/gutenbergkit -b trunk \
    -c <sha> -e NEW_VERSION=vX.Y.Z
```

The CLI is an optional dependency, detected at run time:

- Gated on `bk auth status`, not `command -v bk` — an installed-but-unconfigured CLI can't create builds, and finding that out *after* the version bump is pushed is the worst timing.
- Without it, the script prints the manual steps plus an install hint. No prompting, no state file.
- A failed trigger warns and falls back to the manual steps rather than exiting non-zero — the bump is already pushed, so failing the whole release would misreport it.

The install hint authenticates via `bk auth login`, which uses OAuth and stores the credential in the OS keyring (`bk configure` warns that config-file tokens are deprecated). It requests three scopes — `read_organizations` to complete the login itself, plus `read_pipelines` and `write_builds` to resolve the pipeline and create the build. Without `--scopes`, `bk` requests every scope the account can reach; on my account that was 35+, including `write_pipelines` and `delete_registries`.

Triggering immediately is safe: `:rocket: Publish Swift release` gates on `depends_on` (validate, lint, tests, XCFramework), so it waits for its dependencies regardless of when the build starts.

## Testing Instructions

> [!NOTE]
> Step one requires temporarily commenting out `check_branch` and `check_working_directory`.

1. `make release VERSION_TYPE=patch DRY_RUN=true` — with `bk` installed and authenticated, prints the `bk build create` command it *would* run; without, prints the manual steps and install hint. Nothing is triggered either way.
1. Output opens with `DRY RUN MODE` and closes with `This was a dry run`.

> [!IMPORTANT]
> **The live `bk build create` call has not been exercised through the release script.** Doing so means triggering a real publish build, so we opted to verify on the next real release. The `bk` invocation itself is known to work against this pipeline, and all four code paths are covered with stubs; a failed trigger degrades to the existing manual process. The one unknown is `bk`'s stdout format on success, which is cosmetic.

### DIFF
--- a/bin/release.sh
+++ b/bin/release.sh
@@ -258,21 +258,24 @@ push_changes() {
     print_success "Changes pushed successfully"
 }
 
-# Function to print the post-push instructions for kicking off the
-# Buildkite publish build. CI creates the tag and the GitHub release —
-# this script just bumps the version files on trunk.
+# Buildkite pipeline that publishes tagged releases.
+BUILDKITE_PIPELINE="automattic/gutenbergkit"
+
+# Function to print the manual steps for kicking off the Buildkite publish
+# build. CI creates the tag and the GitHub release — this script just bumps
+# the version files on trunk.
+#
+# $3 frames the steps for the situation that produced them: the fallback
+# after a failed trigger reads differently from the only-path-available case
+# when the `bk` CLI is absent.
 print_publish_instructions() {
     local version=$1
     local sha=$2
+    local heading=$3
     local tag="v$version"
-    local prefix=""
-
-    if [ "$DRY_RUN" = "true" ]; then
-        prefix="[DRY RUN] "
-    fi
 
     echo
-    print_status "${prefix}Next: trigger the Buildkite publish build."
+    print_status "$heading"
     echo
     echo "  1. Open https://buildkite.com/organizations/automattic/pipelines/gutenbergkit/builds/new"
     echo "  2. Branch: trunk"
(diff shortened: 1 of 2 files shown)


mokagio (commented):
@dcalhoun I haven't tested this manually, but it looks like it would work from skimming the code.

However, I wonder if we'd be better served by going through the already set up `release-toolkit` and using the [action to trigger Buildkite builds](https://github.com/wordpress-mobile/release-toolkit/blob/9322fff453fb4622d4d1add8040dc8d13cee2a1d/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/buildkite_trigger_build_action.rb) it exposes.

Two advantages:

- Less custom code to maintain
- No need to require `bk` to be installed on the machine

dcalhoun:
@mokagio thank you for reviewing. I agree, reusing existing tooling is the better approach. I was unaware of that tooling. 

That said, your suggestion prompted me to further consider what I originally perceived to be a relatively low-cost improvement to the current manual release steps. After rereading https://github.com/wordpress-mobile/GutenbergKit/pull/502 and reconsidering the overarching GutenbergKitResources Linear project, my attempt in this PR feels shortsighted and misguided. 

Closing this as it likely more appropriate to await the desired end state outlined in the GutenbergKitResources Linear project. We can revisit these types of release improvements when we have time. Thanks again for your time on this PR and the related project. 🙇🏻‍♂️

PR #594, opened 2026-08-21
fix: recover media uploads that fail server-side post-processing

dcalhoun:
## What?

Registers `@wordpress/api-fetch`'s `mediaUploadMiddleware`, which retries `post-process` when a media upload fails server-side, and makes both the native upload path and the orphan cleanup work with it.

## Why?

Ref CMM-2151. Ref CMM-2274. 

When `wp_generate_attachment_metadata()` fatals server-side — commonly a PHP `memory_limit` or `max_execution_time` fatal on a large image — the upload surfaces as a permanent failure. The user is left with an orphaned gray attachment, and retrying duplicates it (the `-2`/`-3` filename suffixes seen in the field).

WordPress already supports recovering from this: it creates the attachment row, sends `X-WP-Upload-Attachment-ID`, and expects the client to retry `POST /wp/v2/media/<id>/post-process`. Gutenberg's [`mediaUploadMiddleware`](https://github.com/WordPress/gutenberg/blob/da0bf8d33d8817d52c5a61d4c2dae543f70ea93b/packages/api-fetch/src/middlewares/media-upload.ts#L21-L26) implements that. WordPress core [registers it](https://github.com/WordPress/wordpress-develop/blob/8b91cc16cc78b817386b406f50ced8df86fb466d/src/wp-includes/script-loader.php#L372), but GutenbergKit never did.

## How?

Renames the local middleware to `stripDraftPostIdMiddleware` and registers `apiFetch.mediaUploadMiddleware`. `apiFetch.use` unshifts, so registration order is the reverse of execution order: core's is registered after the native one so it runs before it, wrapping it — the native middleware handles an upload without calling `next`, so core's would never run at all from below. Both stay above auth, namespacing, and the root URL so the `post-process` requests core issues stay authenticated.

Two things then had to change for that retry to work in practice:

- `relayResponse` rebuilt the response with a hardcoded `Content-Type`, dropping `x-wp-upload-attachment-id`. It is now relayed via an allowlist and exposed through CORS, and `nativeMediaUploadMiddleware` honors `parse: false` by yielding the `Response` core inspects.
- The orphan cleanup was blocked at CORS preflight: api-fetch tunnels `DELETE` as a `POST` carrying `X-HTTP-Method-Override`, which core's `rest_allowed_cors_headers` omits. Media deletions now route through the loopback server, which permits `DELETE`. This was a pre-existing bug, independent of the retry work.

Also fixes the namespace guard: `new RegExp('(' + [].join('|') + ')')` is `/()/`, which matches every string, so `alreadyHasSiteNamespace` was unconditionally true for self-hosted sites. That was load-bearing — it suppressed a rewrite that would otherwise interpolate `undefined` into every path.

Adds a wp-env media failure simulator (`make wp-env-media-failure MODE=off|recover|always`) so the retry can be exercised locally. See `docs/code/local-wordpress.md`.

### Known limitation

Recovery requires reading `x-wp-upload-attachment-id`, which is only possible same-origin or when the server exposes it via CORS. The native upload server exposes it, so **uploads routed through it recover on both platforms**. A _direct_ upload depends on the site: core's `rest_send_cors_headers()` does not expose the header, so it recovers only where the editor is genuinely same-origin with the site. There is no client-side fallback — core sends the header before `wp_generate_attachment_metadata()`, so the fatal leaves no body to parse the ID from.

## Testing Instructions

Requires the local WordPress environment and a build with native media upload enabled.

1. `make wp-env-start` (add `RESET=1` if credentials are stale)
2. `make wp-env-media-failure MODE=recover`
3. Build and run a demo app, open the editor against **Local WordPress (wp-env)**, and insert an Image block with any image.
4. **Expect:** the upload 500s, a `post-process` request follows and succeeds, and the image renders with no error notice. The attachment has all sub-sizes and there is exactly one of it.
5. `make wp-env-media-failure MODE=always`, then upload again.
6. **Expect:** the upload 500s, five `post-process` attempts fail, then a `DELETE` to `localhost:<port>/media/<id>` returns 200 and the media library is left empty.
7. `make wp-env-media-failure MODE=off` when finished.

Verified end to end on both platforms in all four combinations (iOS and Android × `recover` and `always`).

Not applicable — no user interface changes.

### DIFF
--- a/src/utils/api-fetch.js
+++ b/src/utils/api-fetch.js
@@ -43,10 +43,13 @@ export function configureApiFetch() {
 	// when the server exposes it via CORS:
 	//
 	// - Through the native upload server, which exposes it: works on both
-	//   platforms.
+	//   platforms. This is the only path that recovers everywhere.
 	// - Direct to WordPress, which does not expose it (see
-	//   `rest_send_cors_headers()`): works on Android, whose editor is
-	//   same-origin with the site, and fails on iOS, which loads from `file://`.
+	//   `rest_send_cors_headers()`): never recovers on iOS, which loads from
+	//   `file://`. On Android it recovers only when the editor is genuinely
+	//   same-origin — `GutenbergView` derives the asset domain from the site's
+	//   host, which drops the port, so a site on a non-default port (any local
+	//   dev setup) is cross-origin and does not recover.
 	//
 	// There is no fallback: core sends the header before
 	// `wp_generate_attachment_metadata()`, so the fatal leaves no body to parse
--- a/src/utils/api-fetch.js
+++ b/src/utils/api-fetch.js
@@ -31,6 +31,12 @@ export function configureApiFetch() {
 	apiFetch.use( apiPathModifierMiddleware );
 	apiFetch.use( tokenAuthMiddleware );
 	apiFetch.use( filterEndpointsMiddleware );
+	// `apiFetch.use` unshifts, so the last registration runs first. Core's media
+	// upload middleware is registered before the native one so that it runs
+	// after it, and below auth, namespacing, and the root URL so the
+	// `post-process` requests it issues through `next` stay authenticated and
+	// correctly addressed.
+	apiFetch.use( apiFetch.mediaUploadMiddleware );
 	apiFetch.use( nativeMediaUploadMiddleware );
 	apiFetch.use( stripDraftPostIdMiddleware );
 	apiFetch.use( transformOEmbedApiResponse );
--- a/src/utils/api-fetch.js
+++ b/src/utils/api-fetch.js
@@ -38,6 +38,19 @@ export function configureApiFetch() {
 	// above auth, namespacing, and the root URL, so the `post-process` requests
 	// core issues through `next` are still authenticated and correctly
 	// addressed.
+	//
+	// Recovery needs `x-wp-upload-attachment-id`, readable only same-origin or
+	// when the server exposes it via CORS:
+	//
+	// - Through the native upload server, which exposes it: works on both
+	//   platforms.
+	// - Direct to WordPress, which does not expose it (see
+	//   `rest_send_cors_headers()`): works on Android, whose editor is
+	//   same-origin with the site, and fails on iOS, which loads from `file://`.
+	//
+	// There is no fallback: core sends the header before
+	// `wp_generate_attachment_metadata()`, so the fatal leaves no body to parse
+	// the ID from.
 	apiFetch.use( nativeMediaUploadMiddleware );
 	apiFetch.use( apiFetch.mediaUploadMiddleware );
 	apiFetch.use( stripDraftPostIdMiddleware );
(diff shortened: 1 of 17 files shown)


wp-env/mu-plugins/gutenbergkit-cors.php
```diff
@@ -9,12 +9,13 @@
...
 	return array(
-		'http://localhost:5173',                    // Vite dev server
-		'http://localhost:4173',                    // Vite preview server
-		'http://10.0.2.2:5173',                    // Vite dev server (Android emulator)
-		'http://10.0.2.2:4173',                    // Vite preview server (Android emulator)
-		'https://appassets.androidplatform.net',   // Android production build (HTTPS site)
-		'http://appassets.androidplatform.net',    // Android production build (HTTP site)
+		'http://localhost:5173',                 // Vite dev server
+		'http://localhost:4173',                 // Vite preview server
+		'http://10.0.2.2',                       // wp-env site origin (Android emulator)
```
dcalhoun, line 14: In prior work, `wordpress-rs` began configuring the site URL based upon discovery. This adds the local `wp-env` site's Android emulator domain to avoid CORS errors. Strictly a dev environment fix, no production implications.

src/utils/api-fetch.js
```diff
@@ -67,16 +94,19 @@ function corsMiddleware( options, next ) {
...
 		);
 
+	// Escape the namespaces so each is matched literally rather than as a
+	// pattern.
 	const alreadyHasSiteNamespace =
-		namespaceRegex.test( options.path ) ||
-		/\/sites\/[^/]+\//.test( options.path );
+		new RegExp(
+			`(${ siteApiNamespace.map( escapeRegExp ).join( '|' ) })`
+		).test( options.path ) || /\/sites\/[^/]+\//.test( options.path );
```
dcalhoun, line 109: Address latent bugs by preventing matches from unescaped namespaces.

src/utils/api-fetch.js
```diff
@@ -183,16 +218,44 @@ function filterEndpointsMiddleware( options, next ) {
...
 	const { nativeUploadPort, nativeUploadToken } = getGBKit();
 
+	if ( ! nativeUploadPort || ! nativeUploadToken ) {
+		return next( options );
+	}
+
+	// Each helper returns `null` when the request is not its concern, so an
+	// unhandled request falls through to the default path.
+	return (
+		nativeMediaDelete( options, nativeUploadPort, nativeUploadToken ) ??
```
dcalhoun, line 228: We must relay the media delete request through the proxy to avoid CORS errors due to the critical `X-WP-Upload-Attachment-ID` header.

wp-env/mu-plugins/gutenbergkit-cors.php
dcalhoun: Noting this is an incorrect diagnosis of the problem and should be reverted before merge. The correct solution is implemented in https://github.com/wordpress-mobile/GutenbergKit/pull/599.

PR #600, opened 2026-08-22
refactor: vendor the default editor styles stylesheet

### LINKED ISSUE #590: Embed a local copy of the default editor styles stylesheet

`src/utils/editor-settings.js` imports the default editor styles directly from the `@wordpress/block-editor` package:

```js
import defaultEditorStyles from '@wordpress/block-editor/build-style/default-editor-styles.css?inline';
```

Gutenberg [#81793](https://github.com/WordPress/gutenberg/pull/81793) removed `packages/block-editor/src/default-editor-styles.scss` upstream, since the plugin hasn't enqueued it since [#41306](https://github.com/WordPress/gutenberg/pull/41306) and core owns it now. Our usage was flagged in [this review comment](https://github.com/WordPress/gutenberg/pull/81793/changes#r3807545617), which also noted the removal is a breaking change for the package.

The stylesheet still ships in our pinned `@wordpress/block-editor@16.0.0`, so nothing is broken today. But the next package bump that includes the removal will break the build, and GutenbergKit shouldn't hold up an upstream cleanup either way.

These styles matter for us specifically: `getDefaultEditorSettings()` supplies `defaultEditorStyles` when a site can't provide its own editor settings (no experimental REST API endpoint), which is a supported path in the mobile apps.

Vendor the stylesheet into GutenbergKit and import it locally instead of from the package.

The repo already does this for two similar cases — `src/components/visual-editor/default-theme-styles.scss` and `wp-common-styles.scss` — both imported with `?inline` in `visual-editor/index.jsx`. Same approach here:

1. Add the [removed SCSS](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/default-editor-styles.scss) as a local file (~25 lines: `$default-font` body type, paragraph line height, post title sizing), with a header comment pointing back at the upstream source.
2. Update the import in `src/utils/editor-settings.js` to the local `.scss?inline` path.
3. `@wordpress/base-styles` is already a dependency, so the `@use` of `variables` resolves as-is.

Worth deciding as part of this: whether to keep tracking core's version of these styles (`wp-includes/block-editor.php`) or treat the vendored copy as GutenbergKit's own defaults, which we'd then be free to tune for mobile.

dcalhoun:
## What?

Copies `default-editor-styles.scss` out of `@wordpress/block-editor` and into `src/utils/`, and imports it locally in `editor-settings.js`.

Fixes #590. Fixes CMM-2330.

## Why?

Gutenberg [#81793](https://github.com/WordPress/gutenberg/pull/81793) removed the stylesheet from the package. It still ships in our pinned `@wordpress/block-editor@15.14.0`, so nothing is broken today — but the next bump that includes the removal breaks the build, and our usage shouldn't hold up the upstream cleanup.

The styles matter for us: `getDefaultEditorSettings()` supplies `defaultEditorStyles` when a site can't provide its own editor settings, a supported path in the mobile apps.

## How?

The vendored copy is byte-identical to the upstream file at [`28c348a`](https://github.com/WordPress/gutenberg/blob/28c348acdf3e88bbd90f766e66acc6716bcf9103/packages/block-editor/src/default-editor-styles.scss) (the commit before removal), with a header comment pointing back at the source, the removal PR, and core's equivalent in `wp-includes/block-editor.php`. Same approach as `default-theme-styles.scss` and `wp-common-styles.scss`.

## Testing Instructions

Regression check; no user-facing change expected.

1. Load the editor against a site without the Gutenberg plugin (no editor-settings endpoint supplying styles).
2. Confirm body text renders at 18px, paragraphs at 1.8 line height, and the post title larger and bold — unchanged from `trunk`.

https://claude.ai/code/session_01YZ5JzgG1CKQwmSN2P3B4R6

### DIFF
--- a/src/components/visual-editor/default-theme-styles.scss
+++ b/src/components/visual-editor/default-theme-styles.scss
@@ -3,10 +3,6 @@
   * unavailable or when the editor's theme styles feature is disabled.
   */
 
-.gutenberg-kit-visual-editor__post-title-wrapper .wp-block-post-title {
-	font-size: 20px; // Set reasonable default size for small screens
-}
-
 .block-editor-default-block-appender__content {
 	opacity: 0.62; // Match empty post title/rich text placholder style
 }
--- a/src/utils/default-editor-styles.scss
+++ b/src/utils/default-editor-styles.scss
@@ -1,0 +1,34 @@
+/**
+ * Default editor styles.
+ *
+ * These styles are shown if a theme does not register its own editor style,
+ * a theme.json file, or has toggled off "Use theme styles" in preferences.
+ *
+ * Vendored from the `@wordpress/block-editor` package, which removed the
+ * stylesheet because the Gutenberg plugin no longer enqueues it and WordPress
+ * core owns the equivalent styles.
+ *
+ * @see https://github.com/WordPress/gutenberg/blob/28c348acdf3e88bbd90f766e66acc6716bcf9103/packages/block-editor/src/default-editor-styles.scss
+ * @see https://github.com/WordPress/gutenberg/pull/81793
+ * @see https://github.com/WordPress/wordpress-develop/blob/db9654e6d386c2c69401b9c77c6ce1ae7460a6e4/src/wp-includes/block-editor.php#L218C4-L218C23
+ */
+
+@use "@wordpress/base-styles/variables" as *;
+
+body {
+	font-family: $default-font;
+	font-size: 18px;
+	line-height: 1.5;
+	--wp--style--block-gap: 2em;
+}
+
+p {
+	line-height: 1.8;
+}
+
+.editor-post-title__block {
+	margin-top: 2em;
+	margin-bottom: 1em;
+	font-size: 2.5em;
+	font-weight: 800;
+}
--- a/src/utils/default-editor-styles.scss
+++ b/src/utils/default-editor-styles.scss
@@ -26,9 +26,7 @@ p {
 	line-height: 1.8;
 }
 
-.editor-post-title__block {
-	margin-top: 2em;
-	margin-bottom: 1em;
-	font-size: 2.5em;
+.editor-post-title {
+	font-size: 20px; // Set reasonable default size for small screens
 	font-weight: 800;
 }
--- a/src/utils/editor-settings.js
+++ b/src/utils/editor-settings.js
@@ -1,9 +1,13 @@
 /**
  * WordPress dependencies
  */
-import defaultEditorStyles from '@wordpress/block-editor/build-style/default-editor-styles.css?inline';
 import { store as editorStore } from '@wordpress/editor';
 import { select } from '@wordpress/data';
+
+/**
+ * Internal dependencies
+ */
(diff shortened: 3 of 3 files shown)


nbradbury:
@dcalhoun I deactivated the Gutenberg plugin on my test site, and when I tried the iOS demo app with that site it never loaded - I was stuck with a progress spinner in the middle of the screen. However, the next time I tried it loaded just fine. I asked Claude to investigate and here's what it reported:

> Step 1’s wording, “a site without the Gutenberg plugin”: on iOS a 404 from the settings endpoint throws out of fetchDependencies() rather than falling back (only URLErrors are caught, EditorService.swift:150), so the editor fails to load instead of using defaults. The path you actually want a tester on is theme styles off, or the demo app’s Standalone Editor.

dcalhoun:
@nbradbury thanks for reviewing. I don't believe the hang you experienced relates to the changes in the PR. Also, I'm unsure Claude's reported cause holds. 

When a site lacks the Gutenberg plugin, then it lacks the `/wp-block-editor/v1/settings` REST API endpoint. In that context, GBK never requests the endpoint because the theme styles capability is negotiated up front from route discovery. So, a 404 should not occur. 

There is one context where a 404 can occur for the GBK Demo app due to its rudimentary implementation. 

1. Activate Gutenberg
2. Open the site's Editor Configuration in the GBK Demo app 
3. Note **Support Theme Styles** should be ON
4. Deactivate Gutnberg
5. Open the editor 

A 404 error occurs. This is because a stale capability is captured by the Demo app. However, the UX I observe is an error message: 

<details><summary>Capability mismatch after plugin deactivation</summary>
<p>

image

</p>
</details> 

That said, the hang you experienced is real. It's worthy of a separate issue if we can reproduce it. A few questions...

- When the hang occurred, did you observe an error message in Xcode's console? 
- If not, are you able to reproduce the issue and observe one? 
- Was your local build loading the dev server (via the `GUTENBERG_EDITOR_URL` env var) or a local build (`make build`)? An absent dev server or stale local build could cause issues (although, it'd be odd that it only occurred once). 
- Was the test site newly added or one that Demo app already utilized with Gutenberg/theme styles? 
- Are you able to reproduce the hang on Android? 

It may be worth merging these changes as unrelated to the hang. WDYT?

nbradbury:
> When the hang occurred, did you observe an error message in Xcode's console?
If not, are you able to reproduce the issue and observe one?

Unfortunately, I couldn't reproduce the problem, and I was running the iOS simulator by itself rather than in Xcode. I could not reproduce the problem on Android, either.

> Was the test site newly added or one that Demo app already utilized with Gutenberg/theme styles?

It was a site previously added to the demo app that I then disabled the Gutenberg plugin. In trying to reproduce this I re-enabled the plugin, ran the demo app with that site, then again disabled the plugin. No problems occurred 🤷 

> Was your local build loading the dev server 

Yes.

> It may be worth merging these changes as unrelated to the hang. WDYT?

Claude agrees with you so I'll approve this :)

nbradbury (approved):
:shipit:

PR #601, opened 2026-08-24
build: set project Ruby and Xcode versions

dcalhoun:
## What?

Configure and track the project Ruby version (3.4.9), and bump the Xcode version from 26.0 to 26.6 to match.

## Why?

Pinning the Ruby version avoids unexpected breakages and divergences between local and CI environments. 3.4.9 matches WordPress-iOS.

The Xcode bump is required for the Ruby pin to work. The available Ruby is a property of the CI Xcode image, and `xcode-26.0` does not ship 3.4.9, so pinning it alone broke `rbenv` on the build agents:

```
rbenv: version `3.4.9' is not installed (set by .../.ruby-version)
```

WordPress-iOS pins `26.6` and runs 3.4.9 successfully, so the two move together here as well — the same way WordPress-iOS did it in a single change.

## How?

- `.ruby-version`: new file pinning 3.4.9.
- `.xcode-version`: `26.0` → `26.6`, which feeds `IMAGE_ID` via `.buildkite/shared-pipeline-vars`.

## Testing Instructions

Verified the Demo app ran via Xcode. Verified CI checks succeeded. 

N/A, no user-facing changes.

## Screenshots or screencast 
N/A, no user-facing changes.


mokagio (approved):
Thanks @dcalhoun ! I pushed a few additional update/hygiene commits.

Ship it.

Issue #591, opened 2026-08-19
Use get_prs_between_tags for release notes once release-toolkit ships fail_on_error


dcalhoun:
`fastlane/Fastfile` builds the GitHub Release body by calling
`Fastlane::Helper::GithubHelper#generate_release_notes` directly, rather than the
`get_prs_between_tags` action that wraps it.

That was a deliberate workaround: the action rescued every `StandardError` and
returned the message as the changelog, so a transient API failure would publish a
release whose notes read `❌ Error computing the list of PRs…`. We push the tag
before creating the Release, so that needs to fail the step instead.

wordpress-mobile/release-toolkit#772 added a `fail_on_error:` parameter and is
merged, but not yet in a released gem — we're pinned to `~> 14.11`, locked at
`14.11.3`, and the entry is still under `## Trunk` in the toolkit CHANGELOG.

Once it ships:

- [ ] Bump `fastlane-plugin-wpmreleasetoolkit` in `Gemfile.lock` to a version
      containing #772
- [ ] Replace the `generated_release_notes` helper's `GithubHelper` call with
      `get_prs_between_tags(..., fail_on_error: true)`
- [ ] Drop the now-stale comment explaining why we bypass the action

`fail_on_error: true` is required, not optional. Without it the action swallows the
`400 Invalid previous_tag parameter` that a malformed `previous_tag` produces and
publishes the error text as the release body — the exact failure #772 exists to
prevent.

Context: https://github.com/wordpress-mobile/GutenbergKit/pull/588#discussion_r3815118392

PR #607, opened 2026-08-27
ci(release): use get_prs_between_tags with fail_on_error

mokagio:
AI-generated below:

## What?

Bumps `fastlane-plugin-wpmreleasetoolkit` to 15.0.0 and switches `generated_release_notes` back to the `get_prs_between_tags` action, now passing `fail_on_error: true`.

Closes [CMM-2332](https://linear.app/a8c/issue/CMM-2332) / #591.

## Why?

#588 deliberately bypassed the action and called `GithubHelper#generate_release_notes` directly, because the action rescued every `StandardError` and returned the message as the changelog — a transient API failure would publish a release whose notes read `❌ Error computing the list of PRs…`. We push the tag before creating the Release, so that has to fail the step instead.

[release-toolkit#772](https://github.com/wordpress-mobile/release-toolkit/pull/772) added the opt-out, and it has now shipped in 15.0.0, so the workaround can go.

## How?

`fail_on_error: true` is load-bearing rather than cosmetic, which is the one thing the diff doesn't say on its own — hence the comment that replaces the old workaround note.

The major bump is manual: Dependabot only proposes this repo's toolkit updates through the `ruby-minor-and-patch` group. Neither of 15.0.0's breaking changes reaches this repo — there are no `*_build_preflight` call sites, and `EnvManager` now populating the process `ENV` by default is a no-op here, where no `.env` file ships.

## Testing Instructions

The release lane only runs on CI, so I exercised the action directly against the live API from this branch:

```sh
bundle exec fastlane run get_prs_between_tags \
  repository:wordpress-mobile/GutenbergKit tag_name:v0.19.0 \
  previous_tag:v0.18.1 fail_on_error:true
```

Returns the v0.18.1 → v0.19.0 notes. Swapping in a nonexistent `previous_tag` exits 1 with `Octokit::BadRequest: 400 - Invalid previous_tag parameter`, where the same call with `fail_on_error:false` exits 0 and hands back the `❌ Error computing the list of PRs…` string — i.e. exactly the failure this change exists to prevent.

Heads-up for whoever reads the next release page: the action rewrites GitHub's `## What's Changed` heading into `## New PRs since [<previous_tag>](<link>)`.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -317,14 +317,16 @@ end
 # previous tag. `set_github_release`'s `is_generate_release_notes` cannot express
 # one.
 #
-# Uses `GithubHelper` rather than the `get_prs_between_tags` action that wraps it
-# until https://github.com/wordpress-mobile/release-toolkit/pull/772 gets ships in the next release-toolkit
-# version (at which point we'll be able to use the action and its new `fail_on_error: true` parameter)
+# `fail_on_error` is required, not cosmetic: the tag is already pushed by this
+# point, and the action otherwise returns the API error as the changelog itself,
+# publishing a release whose notes read "❌ Error computing the list of PRs…".
 def generated_release_notes(version:, token:)
-  Fastlane::Helper::GithubHelper.new(github_token: token).generate_release_notes(
+  get_prs_between_tags(
     repository: GITHUB_REPO,
     tag_name: version,
-    previous_tag: previous_release_tag!(version: version, token: token)
+    previous_tag: previous_release_tag!(version: version, token: token),
+    fail_on_error: true,
+    github_token: token
   )
 end
 
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -316,7 +316,6 @@ end
 # Build the release body via GitHub's notes generator, pinned to an explicit
 # previous tag. `set_github_release`'s `is_generate_release_notes` cannot express
 # one.
-#
 def generated_release_notes(version:, token:)
   get_prs_between_tags(
     repository: GITHUB_REPO,
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -317,14 +317,13 @@ end
 # previous tag. `set_github_release`'s `is_generate_release_notes` cannot express
 # one.
 #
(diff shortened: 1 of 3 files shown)


iangmaia (approved):
👍 Thanks for addressing this.

fastlane/Fastfile
```diff
@@ -317,14 +317,16 @@ end
...
+# point, and the action otherwise returns the API error as the changelog itself,
+# publishing a release whose notes read "❌ Error computing the list of PRs…".
 def generated_release_notes(version:, token:)
-  Fastlane::Helper::GithubHelper.new(github_token: token).generate_release_notes(
+  get_prs_between_tags(
     repository: GITHUB_REPO,
     tag_name: version,
-    previous_tag: previous_release_tag!(version: version, token: token)
+    previous_tag: previous_release_tag!(version: version, token: token),
+    fail_on_error: true,
```
mokagio, line 326:
```suggestion
    # Hard fail on errors. Otherwise, we might get release notes reading something
    # like "Error computing the list of PRs."
    fail_on_error: true,
```

fastlane/Fastfile
```diff
@@ -317,14 +317,16 @@ end
 # previous tag. `set_github_release`'s `is_generate_release_notes` cannot express
 # one.
 #
-# Uses `GithubHelper` rather than the `get_prs_between_tags` action that wraps it
-# until https://github.com/wordpress-mobile/release-toolkit/pull/772 gets ships in the next release-toolkit
-# version (at which point we'll be able to use the action and its new `fail_on_error: true` parameter)
+# `fail_on_error` is required, not cosmetic: the tag is already pushed by this
+# point, and the action otherwise returns the API error as the changelog itself,
+# publishing a release whose notes read "❌ Error computing the list of PRs…".
```
mokagio:
```suggestion
```

fastlane/Fastfile
```diff
@@ -317,14 +317,15 @@ end
 # previous tag. `set_github_release`'s `is_generate_release_notes` cannot express
 # one.
 #
```
mokagio:
```suggestion
```

AliSoftware (approved):

PR #608, opened 2026-08-27
build: add stylelint for RTL-unsafe logical CSS properties

### LINKED ISSUE #564: Add stylelint with logical-property rule to catch RTL styling bugs

akashfer:
Fixes #564

## What?

Adds a stylelint setup to GutenbergKit's SCSS, using `stylelint-plugin-logical-css` to flag physical (`left`/`right`/`padding-left`/etc.) properties that should be written as logical properties (`inset-inline-start`, `padding-inline`, etc.) for RTL correctness. Converts the ~29 existing violations the new rule flags to their logical equivalents.

## Why?

GutenbergKit's own SCSS is never processed by `rtlcss`, and nothing currently guards against physical-property RTL bugs — one (`border-right-color` misplacing the toolbar divider) had to be found by eye during RTL testing. There was no stylelint setup at all: no config, no dependency, no `lint:css` script, no Makefile target.

## How?

- Added `stylelint`, `stylelint-plugin-logical-css` (pinned to `^1.2.3`, matching the version upstream Gutenberg pins in `tools/stylelint/config.js` — v2 renamed the rule this issue asks for), `@wordpress/stylelint-config`, and `postcss-scss` as dev dependencies.
- Added `.stylelintrc.mjs` extending `@wordpress/stylelint-config/scss` and enabling `plugin/use-logical-properties-and-values`, reusing upstream's `ignore` list for properties that don't affect RTL (`margin-top`, `width`, `overflow-y`, `border-top`, etc.).
- The base `@wordpress/stylelint-config/scss` ruleset is otherwise stricter than this codebase's existing conventions (BEM-style double-underscore class names, blank-line formatting, specificity ordering). Relaxed the same rules upstream's own config relaxes for the same reason, to keep this PR scoped to the logical-properties rule rather than a repo-wide style rewrite.
- Added `lint:css` / `lint:css:fix` npm scripts and matching `lint-css` / `lint-css-fix` Makefile targets, mirroring the existing `lint-js` pattern.
- Wired `lint-css` into the Buildkite pipeline alongside `lint-js`, including the release-gate `depends_on` list.
- Converted the ~29 flagged declarations (all symmetric `left`/`right` pairs, or single vertical properties like `scroll-padding-bottom`) to their logical equivalents.
- Two declarations in the toolbar's scroll-indicator gradients (`::before`/`::after`) are *not* a symmetric pair — they're the left- and right-scroll affordances, and whether they should flip in RTL needs its own investigation. Left those physical with an inline `stylelint-disable-next-line` and a comment, rather than guessing at a fix.
- Along the way, fixed a handful of small pre-existing issues the new linter surfaced independent of RTL (a dead duplicate `position: absolute` overridden by a later `position: fixed !important`, two duplicate selector blocks that can merge into one, three `0px` → `0` unit cleanups, and a few non-shorthand/named colors).

## Testing Instructions

1. `npm run lint:css` (or `make lint-css`) — should pass with no errors.
2. `make lint-js` / `make test-js` / `make build` — included here to confirm the change doesn't affect JS lint, unit tests (227 passing), or the Vite build.
3. Spot-check the converted rules still render correctly, e.g. the fixed toolbar (`src/components/editor-toolbar/style.scss`) and the visual editor toolbar/error boundary (`src/components/visual-editor/style.scss`) in both LTR and RTL (`?lang=ar` or similar).

No UI/behavior changes — this is a build-tooling and CSS-property-name change only; visual output is unchanged (logical properties resolve to the same physical box-model values in LTR, which is this project's current default and only tested direction).


AliSoftware:
In terms of infra/tooling (`Makefile` change & `package.json` changes) the changes looks ok to me.
But in terms of the CSS changes, I'm adding @dcalhoun as reviewer because he'll probably be more knowledgable in the changes this PR makes to the CSS properties in the GutenbergKit's code itself than our team would.

dcalhoun:
@akashfer thank you for exploring this. It may take me a little while to begin review of this, but hopefully I can next week. I'll follow up.

akashfer:
Okay thanks

src/components/editor-toolbar/style.scss
```diff
@@ -19,9 +18,11 @@ $scroll-indicator-elevation: 32;
 }
 
 // Disable overflow scrolling for the block toolbar, and rely upon scrolling the
-// parent editor toolbar to avoid nested scrolling views.
+// parent editor toolbar to avoid nested scrolling views. Also disable scrolling
+// of the block toolbar itself, relying on the parent container scrolling instead.
```
dcalhoun, line 22: It seems we can forgo the copy of this comment from line 119. It duplicates the existing comment. 
```suggestion
// parent editor toolbar to avoid nested scrolling views. 
```

.stylelintrc.mjs
```diff
@@ -0,0 +1,51 @@
+/** @type {import('stylelint').Config} */
+export default {
+	extends: '@wordpress/stylelint-config/scss',
+	plugins: ['stylelint-plugin-logical-css'],
```
dcalhoun, line 4: *Finding from Claude:*

This file fails `prettier --check` (wp-prettier's bracket spacing). CI won't catch it: formatting is enforced via the `prettier/prettier` ESLint rule, and `lint:js` runs `eslint . --ext js,jsx`, so `.mjs` is never linted.

```suggestion
	plugins: [ 'stylelint-plugin-logical-css' ],
```

Worth adding `mjs,cjs` to the `--ext` list in a follow-up — `.eslintrc.cjs`, `.prettierrc.cjs`, and `e2e/.eslintrc.cjs` all already pass, so it'd be a no-op today.

.stylelintrc.mjs
```diff
@@ -0,0 +1,51 @@
...
+	rules: {
```
dcalhoun, line 5: *Finding from Claude:*

Gutenberg's `tools/stylelint/config.js` sets both of these; without them the `stylelint-disable-next-line` comments added below can rot silently, and a stale disable will suppress whatever lands on the line after it. `lint:js` already enforces the equivalent via `--report-unused-disable-directives`.

```suggestion
	reportNeedlessDisables: true,
	reportDescriptionlessDisables: true,
	rules: {
```

`reportDescriptionlessDisables` requires a `-- reason` on each disable directive, which pairs with the toolbar comments.

.stylelintrc.mjs
```diff
@@ -0,0 +1,51 @@
...
+		'plugin/use-logical-properties-and-values': [
```
dcalhoun, line 6: *Finding from Claude:*

Gutenberg is on `stylelint-plugin-logical-css@^2.1.0`, where this rule was renamed and split in two. Worth moving now rather than landing disable comments under a name we'd have to rename later — `^1.2.3` can't cross the major, so the pin freezes here otherwise.

I ran v2 against `src/**/*.scss` with this exact ignore list: the only violations are the two toolbar lines that already carry disables, flagged because v2 doesn't recognise the v1 name in them. Nothing else in the codebase changes.

```suggestion
		'logical-css/require-logical-keywords': true,
		'logical-css/require-logical-properties': [
```

Goes together with the version bump on `package.json` and the two disable directives in `editor-toolbar/style.scss`.

.stylelintrc.mjs
```diff
@@ -0,0 +1,51 @@
...
+					'margin-top',
+					'margin-bottom',
+					'overflow-x',
+					'overflow-y',
+					'padding-top',
+					'padding-bottom',
+					'scroll-margin-top',
+					'scroll-margin-bottom',
+					'top',
+					'bottom',
```
dcalhoun, line 28: *Finding from Claude:*

The list is copied from Gutenberg and inherits its gaps: it ignores the block-axis _shorthands_ but not the longhands. Per the plugin's `physical.js`, these are also flagged and equally unable to flip in RTL:

```suggestion
					// Doesn't affect RTL styles
					'border-bottom',
					'border-bottom-color',
					'border-bottom-style',
					'border-bottom-width',
					'border-top',
					'border-top-color',
					'border-top-style',
					'border-top-width',
					'box-orient',
					'contain-intrinsic-height',
					'contain-intrinsic-width',
					'width',
					'min-width',
					'max-width',
					'height',
					'min-height',
					'max-height',
					'margin-top',
					'margin-bottom',
					'overflow-x',
					'overflow-y',
					'overscroll-behavior-x',
					'overscroll-behavior-y',
					'padding-top',
					'padding-bottom',
					'scroll-margin-top',
					'scroll-margin-bottom',
					'scroll-padding-top',
					'scroll-padding-bottom',
					'top',
					'bottom',
```

(`border-top-left-radius` and friends stay flagged — they name a side on the inline axis.) This is what forced the `scroll-padding-bottom` conversion in `src/index.scss`.

package.json
```diff
@@ -97,17 +99,21 @@
...
 		"jsdom": "^29.0.2",
 		"magic-string": "^0.30.21",
 		"patch-package": "^8.0.1",
+		"postcss-scss": "^4.0.9",
 		"prettier": "npm:wp-prettier@^3.0.3",
 		"react-devtools": "^7.0.1",
 		"react-devtools-core": "^7.0.1",
 		"sass-embedded": "^1.98.0",
+		"stylelint": "^16.26.1",
+		"stylelint-plugin-logical-css": "^1.2.3",
```
dcalhoun: *Finding from Claude:*

These two are inverted. `stylelint-scss` is a declared peer of `@wordpress/stylelint-config` and its `scss.js` loads it by name (`plugins: ['stylelint-scss']`) — it only resolves today because npm auto-installs and hoists the peer, so a non-hoisting installer or `--legacy-peer-deps` breaks config load. `postcss-scss`, meanwhile, is a direct dependency of `stylelint-config-recommended-scss`, which requires it internally; nothing here imports it.

```suggestion
		"prettier": "npm:wp-prettier@^3.0.3",
		"react-devtools": "^7.0.1",
		"react-devtools-core": "^7.0.1",
		"sass-embedded": "^1.98.0",
		"stylelint": "^16.26.1",
		"stylelint-plugin-logical-css": "^1.2.3",
		"stylelint-scss": "^6.4.0",
```

src/index.scss
```diff
@@ -11,7 +11,7 @@ $toolbar-height: 56px;
 
 html {
 	// Ensure the toolbar does not overlap the editor content
-	scroll-padding-bottom: $toolbar-height;
+	scroll-padding-block-end: $toolbar-height;
```
dcalhoun, line 14: The block axis is always vertical and RTL flips only the inline axis. I suggest we revert this change to keep "bottom" both here to match `.gutenberg-kit-visual-editor__toolbar`. 

The ignore list also already treats `bottom`, `padding-bottom`, `margin-bottom` and `scroll-margin-bottom` as non-flipping—`scroll-padding-bottom` is outside it only by omission.

```suggestion
	scroll-padding-bottom: $toolbar-height;
```

Note this depends on the ignore-list suggestion on `.stylelintrc.mjs`; applied on its own it will fail the new rule.

src/index.scss
```diff
@@ -82,7 +83,7 @@ body.gutenberg-kit:not(.is-ios) .block-editor-block-list__layout::selection {
 	}
 
 	.block-settings-menu__close {
-		margin: 8px 8px 0px 0px;
+		margin: 8px 8px 0 0;
```
dcalhoun, line 86: *Finding from Claude:*

This shorthand is still RTL-broken and the new rule can't see it — it only inspects longhand property names and single keyword values, never multi-value shorthands. I verified: a file containing `margin: 8px 8px 0 0;` reports 0 problems while `padding-left` on the next line is flagged.

`.block-settings-menu__header` is a column flex container with `align-items: end`, so the close button renders at the inline end — the left edge in RTL — while `margin-right: 8px; margin-left: 0` puts the gap on the inside.

```suggestion
		margin-block: 8px 0;
		margin-inline: 0 8px;
```

Upstream pairs the logical rule with `declaration-property-max-values` (`margin: 3, padding: 3, border-width: 3, border-color: 3, border-style: 3, border-radius: 3, inset: 3`) under the comment "Prevents left/right values with shorthand property names (unclear for RTL)". Without it this whole class passes CI.

src/index.scss
```diff
@@ -130,16 +131,8 @@ body.gutenberg-kit:not(.is-ios) .block-editor-block-list__layout::selection {
...
-		padding-right: 8px !important;
-	}
-
-	// Inserter title
-	.components-popover__header-title {
-		font-size: 17px;
-		font-weight: 600;
-		text-align: center;
+		padding: 0 !important;
+		padding-inline: 8px !important;
```
dcalhoun, line 135: *Finding from Claude:*

This relies on declaration order: `padding: 0` zeroes all four sides, then `padding-inline` re-applies the inline ones. Swap the two lines — via a `--fix` run, a merge, or a reordering formatter — and the 8px gutters silently collapse. `declaration-block-no-duplicate-properties` is configured with `ignore: ['consecutive-duplicates']`, so stylelint won't warn either.

```suggestion
		padding-block: 0 !important;
		padding-inline: 8px !important;
```

src/components/visual-editor/default-theme-styles.scss
```diff
@@ -16,9 +16,9 @@
 	--wp--preset--aspect-ratio--2-3: 2/3;
 	--wp--preset--aspect-ratio--16-9: 16/9;
 	--wp--preset--aspect-ratio--9-16: 9/16;
-	--wp--preset--color--black: #000000;
+	--wp--preset--color--black: #000;
 	--wp--preset--color--cyan-bluish-gray: #abb8c3;
-	--wp--preset--color--white: #ffffff;
+	--wp--preset--color--white: #fff;
```
dcalhoun, line 21: *Finding from Claude:*

This file transcribes core's `theme.json` preset output and is injected verbatim into the canvas. Core emits the six-digit form, so shortening these purely to satisfy `color-hex-length` guarantees two spurious lines whenever someone re-syncs.

```suggestion
	--wp--preset--color--black: #000000;
	--wp--preset--color--cyan-bluish-gray: #abb8c3;
	--wp--preset--color--white: #ffffff;
```

Let's add a narrow `overrides` entry disabling `color-hex-length` for this file rather than `ignoreFiles`, which would also switch off the logical-CSS checking we want here.

src/components/editor-toolbar/style.scss
```diff
@@ -46,6 +47,10 @@ $scroll-indicator-elevation: 32;
 		content: "";
 		position: absolute;
 		top: 0;
+		// Not a symmetric pair with the `right` below (see the `::after` gradient) —
+		// whether this should flip in RTL needs its own investigation, so it is
+		// intentionally left physical for now rather than converted here.
+		// stylelint-disable-next-line plugin/use-logical-properties-and-values
```
dcalhoun, line 53: *Finding from Claude:*

The stated reason doesn't hold — this isn't pending investigation, it's a permanent constraint. `left: 0` pairs with the hard-coded `linear-gradient(to right)` three lines below; `linear-gradient` has no logical direction, so converting the position alone would put the fade on the wrong edge in RTL. Both have to flip together or neither does.

```suggestion
		// stylelint-disable-next-line plugin/use-logical-properties-and-values -- paired with the gradient direction below
```

The directive will need renaming to `logical-css/require-logical-properties` if the plugin is bumped to v2.

src/components/editor-toolbar/style.scss
```diff
@@ -66,6 +71,9 @@ $scroll-indicator-elevation: 32;
 		content: "";
 		position: absolute;
 		top: 0;
+		// Not a symmetric pair with the `left` above — see the note on the
+		// `::before` gradient.
+		// stylelint-disable-next-line plugin/use-logical-properties-and-values
```
dcalhoun, line 76: *Finding from Claude:*

Same as above — `right: 0` pairs with `linear-gradient(to left)` below, which is the real reason this can't be converted in isolation.

```suggestion
		// stylelint-disable-next-line plugin/use-logical-properties-and-values -- paired with the gradient direction below
```

The directive will need renaming to `logical-css/require-logical-properties` if the plugin is bumped to v2.

Makefile
```diff
@@ -231,6 +231,14 @@ wp-env-android-urls: ## Report whether WordPress emits emulator-reachable URLs,
 format: npm-dependencies ## Format code
 	npm run format
 
+.PHONY: lint-css
+lint-css: npm-dependencies ## Lint SCSS code
+	npm run lint:css
+
+.PHONY: lint-css-fix
+lint-css-fix: npm-dependencies ## Lint and auto-fix SCSS code
+	npm run lint:css:fix
```
dcalhoun: *Finding from Claude:*

`lint-css` lands as a required CI step, but none of the three places documenting the pre-submit lint workflow mention it:

- `AGENTS.md:68` — "**CRITICAL**: Always run these commands..." lists only `lint-js-fix` and `lint-swift-fix`
- `CONTRIBUTING.md:22` — "Run `make format` and `make lint-js` before submitting changes"
- `docs/code/testing.md:110-123`

Someone following those verbatim finds out about `lint-css` from a red build. Worth adding `make lint-css-fix` alongside in all three.

Also: `.claude/settings.json` wires a `PostToolUse` hook for prettier and SwiftLint but not stylelint, so the in-editor loop the repo already relies on for Swift is missing for SCSS.

package.json
```diff
@@ -97,17 +99,21 @@
...
 		"@wordpress/prettier-config": "^4.43.0",
+		"@wordpress/stylelint-config": "^24.3.0",
 		"ajv": "^8.18.0",
 		"eslint": "^8.57",
 		"eslint-import-resolver-typescript": "^4.4.4",
 		"eslint-plugin-react-refresh": "^0.4.26",
 		"jsdom": "^29.0.2",
 		"magic-string": "^0.30.21",
 		"patch-package": "^8.0.1",
+		"postcss-scss": "^4.0.9",
```
dcalhoun, line 110: *Finding from Claude:*

`postcss-scss` is a direct dependency of `stylelint-config-recommended-scss`, which requires it internally to set `customSyntax`. Nothing in this repo imports it, so declaring it here is dead weight we'd have to version-bump by hand.

```suggestion
		"patch-package": "^8.0.1",
```

package.json
dcalhoun, line 116: *Finding from Claude:*

Bump to match Gutenberg — see the rule-name comment on `.stylelintrc.mjs`, which has the measured cost. Apply both together or neither; the renamed rules won't resolve on v1.

```suggestion
		"stylelint-plugin-logical-css": "^2.1.0",
```

package.json
```diff
@@ -97,17 +99,21 @@
...
 		"prettier": "npm:wp-prettier@^3.0.3",
 		"react-devtools": "^7.0.1",
 		"react-devtools-core": "^7.0.1",
 		"sass-embedded": "^1.98.0",
+		"stylelint": "^16.26.1",
+		"stylelint-plugin-logical-css": "^1.2.3",
 		"vite": "^8.0.16",
```
dcalhoun, line 117: *Finding from Claude:*

`stylelint-scss` is missing. `@wordpress/stylelint-config` declares it as a peer and its `scss.js` loads it by name (`plugins: ['stylelint-scss']`), so it only resolves today because npm auto-installs and hoists the peer. Under a non-hoisting installer or `--legacy-peer-deps`, config load fails outright.

```suggestion
		"stylelint-scss": "^6.4.0",
		"vite": "^8.0.16",
```

dcalhoun (changes requested):
@akashfer thank you for contributing this! 🙇🏻‍♂️ 

I tested the editor with the style changes. I did not encounter any regressions. 

In my review with Claude, I did uncover several findings that I believe are legitimate and worth addressing. **Would you please take a look at addressing the inline comments?** 

Lastly, if you merge the latest trunk branch into this branch, we'll receive https://github.com/wordpress-mobile/GutenbergKit/pull/615, which will help avoid cryptic lint run hangs.

PR #609, opened 2026-08-28
Add initial iOS Find in Post support

yunuschoudhurywork-lang:
## What?

Adds the initial iOS native Find in Post spike to GutenbergKit by enabling `WKWebView`'s native Find interaction and exposing a `presentFindNavigator()` method from `EditorViewController`.

## Why?

This is part of the work for WordPress-Android #23222 ("Find in post").

The editor is shared through GutenbergKit, so the Find behavior should be implemented at the GutenbergKit level rather than separately in the host applications.

This spike is intended to verify whether the native `WKWebView` Find interaction is suitable for Gutenberg's `contenteditable` editor before proceeding with the Android implementation.

## How?

- Enables `WKWebView.isFindInteractionEnabled`.
- Adds `presentFindNavigator()` to `EditorViewController`, which presents the native iOS Find navigator.
- Keeps the implementation limited to the native iOS Find API for this initial feasibility spike.

## Testing Instructions

Runtime testing is still pending.

I am developing on Windows and currently do not have access to a macOS/iOS environment to run the GutenbergKit iOS application.

Once an iOS environment is available, verify:

1. Open a post containing repeated instances of the same text.
2. Open the native Find navigator.
3. Search for the repeated text.
4. Verify that matching text is highlighted.
5. Navigate between matches using next/previous.
6. Verify that the editor scrolls to the selected match.
7. Dismiss the Find navigator.
8. Verify that searching does not modify the post or create an undo/redo entry.
9. Verify that Gutenberg's caret/block selection is not unexpectedly changed.

## Accessibility Testing Instructions

Runtime accessibility testing is pending because an iOS/macOS environment is not currently available.

Once available, verify that the native Find navigator can be opened and operated using VoiceOver, including entering a search query, navigating between matches, and dismissing the navigator.

## Screenshots or screencast

Not available yet because runtime iOS testing is pending.

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
+++ b/ios/Sources/GutenbergKit/Sources/EditorViewController.swift
@@ -245,6 +245,7 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         config.applicationNameForUserAgent = "GutenbergKit/\(GutenbergKitVersion.version)"
 
         self.webView = GBWebView(frame: .zero, configuration: config)
+        self.webView.isFindInteractionEnabled = true
         self.webView.scrollView.keyboardDismissMode = .interactive
 
         self.isWarmupMode = isWarmupMode
@@ -556,6 +557,11 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
         guard isReady else { return }
         evaluate("editor.redo();")
     }
+
+    public func presentFindNavigator() {
+        webView.findInteraction?.presentFindNavigator(showingReplace: false)
+    }
+
 
     /// Dismisses the topmost modal dialog or menu in the editor
     public func dismissTopModal() {


dcalhoun:
@yunuschoudhurywork-lang thank you for exploring this. It may take me a little while to begin review of this, but hopefully I can next week. I'll follow up.

yunuschoudhurywork-lang:
@dcalhoun sure sir!! I'll wait for your confirmation then only I proceed further.

yunuschoudhurywork-lang:
@dcalhoun hi sir! Any update.

dcalhoun:
@yunuschoudhurywork-lang no, I was unable to review last week. It remains on my list, so I intend to follow up. Thank you for your patience.

dcalhoun:
@yunuschoudhurywork-lang testing this went well. I pushed few commits to this branch during my testing. 

<details><summary>Screen recording of usage</summary>
<p>


https://github.com/user-attachments/assets/11730236-ecdd-4300-85c4-55d7c6a6ff1f



</p>
</details> 

The usage of the native `UIFindInteraction` API seems like a valid path for iOS. That finding doesn't necessarily decide what we should do for Android, but it rules out that iOS **requires** a JS-based implementation. 

I believe we can proceed forward exploring the Android implementation as discussed in https://github.com/wordpress-mobile/WordPress-Android/issues/23222#issuecomment-5380715518. 

Below are a details summarized by Claude, where I discovered a few issues that we'll need to explore on iOS. 

---

**Spike result: the native `UIFindInteraction` path works.**

The blocking question from the exploration notes — does native find decoration behave inside `contenteditable` — comes back clean. Highlighting, the match counter, and prev/next all work, with no observed disruption to caret position, block selection, dirty state, or undo history.

Three things to carry forward:

**1. Match placement is wrong while the software keyboard is up.** WebKit's find scroll path subtracts the keyboard from a viewport GBK has already subtracted it from, so the match parks against the top edge and clips by ~5pt. Measured: −5pt with the keyboard up, vs. 41% down the viewport with it dismissed. Caret scroll-into-view is unaffected, so this is specific to the find path rather than an editor layout bug. CSS `scroll-padding`/`scroll-margin` and `scrollView.contentInset` have no effect (verified with controls). The natural flow works around it — type the query, dismiss the keyboard with the search key, then step through matches. A real fix means moving the bottom toolbar's keyboard avoidance into the web layer, since the web view's shrink is what keeps the toolbar above the keyboard.

**2. "Find Selection" comes free in the text-selection menu, and it's buggy.** `isFindInteractionEnabled` adds it automatically. Sessions started that way never scroll to the first match until the query is modified — both in the demo and WP-iOS. The "Find in Post" entry point is unaffected. We can filter the item out of the edit menu if we'd rather not ship it.

**3. There are no find-session callbacks.** `UIFindInteraction.delegate` is readonly (set only via the initializer) and `findNavigatorVisible` isn't KVO-compliant, so notifying the host when find opens or closes would require polling. Worth dropping the idea.

Android is unblocked — the iOS verification that gated it passed. Item 1 is WebKit-specific and shouldn't recur there, but a custom find bar is still required regardless.

PR #612, opened 2026-09-01
fix(ios): refuse a passthrough upload whose body measures zero

dcalhoun:
## What?

The native upload passthrough could send WordPress an empty body and report the result to the editor as a successful upload.

Found while reviewing #611, but pre-existing and independent of it, so this lands on its own. The two touch `MediaUploadServer.swift` in different functions and merge cleanly in either order.

## Why?

`RequestBody.count` is a file-size lookup for a disk-backed body, and reports zero when the lookup fails. `passthroughUpload` sent that as the `Content-Length`, so nothing was uploaded — and WordPress accepts an empty media POST as a no-op and answers 2xx, which the editor reports as success for a file that never arrived.

The likeliest cause of a zero — a temp file that has gone missing — already throws in `makeInputStream()`, which both callers turn into a 500. So this closes the window that remains rather than a live failure: a body that opens cleanly but whose length is unusable.

## How?

Read the length once, refuse a zero before the request is built, and log it. The caller already converts the throw into a 500, so the editor surfaces a real failure instead of a false success.

## Testing Instructions

```sh
make test-swift-package
```

The new test drives a body that opens cleanly and measures zero, then asserts nothing reached WordPress. Reverting the guard and re-running shows what it catches:

```
Expectation failed: an error was expected but none was thrown and
"MediaUploadResponse(statusCode: 201, body: 2 bytes)" was returned
(client.lastURL → https://example.com/wp-json/wp/v2/media) == nil
```

That 201 is the failure mode: an upload of nothing, reported as success.

### DIFF
--- a/ios/Sources/GutenbergKit/Sources/Media/MediaUploadServer.swift
+++ b/ios/Sources/GutenbergKit/Sources/Media/MediaUploadServer.swift
@@ -480,10 +480,22 @@ class DefaultMediaUploader: @unchecked Sendable {
     /// Used when the delegate's `processFile` returned the file unchanged —
     /// the incoming multipart body is already valid for WordPress.
     func passthroughUpload(body: RequestBody, contentType: String, query: String) async throws -> MediaUploadResponse {
+        // `count` is a file-size lookup for a disk-backed body, and reports zero
+        // when it fails. Sending that as the `Content-Length` would upload an
+        // empty body, which WordPress accepts as a no-op and answers with a 2xx
+        // — reported to the editor as a successful upload of nothing. Fail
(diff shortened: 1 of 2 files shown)


dcalhoun:
After further review, this was deemed invalid. 

> The guard defended against `RequestBody.count` reporting zero from a failed file-size lookup, which is the behavior of `Storage.file(url)`. The parser never produces that case: `extractBody` yields `.data` below the in-memory threshold and `.fileSlice` above it, and a slice's `count` returns the length the parser recorded, with no file-system call and no failure mode. The streaming branch only sees `.fileSlice`, so the guard could not fire and its 500 response was unreachable.

PR #625, opened 2026-09-05
fix(ios)!: own the media upload delegate, and give hosts a way out

jkmassel:
Stacked on #624. Second of nine PRs splitting #621.

## What?

A host that releases its `mediaUploadDelegate` while an upload is in flight gets the file forwarded to WordPress **unprocessed**. Hold the delegate strongly, as Android already does — and pay the cost that creates rather than leaving it in the docs: `stopMediaHandling()` is the way out for a host that closes the cycle owning it makes possible.

## Why?

The server reads the delegate three times per request — once at the admission gate (`handlesFile`), then again for `processFile` and `uploadFile`. Those reads are separated by a synchronous disk copy and an unbounded `processFile`. Held weakly, a host releasing its delegate in that window changes the answer between them: a file admitted for processing is delivered untouched.

The `weak` bought no leak protection to trade away. The cycle it named runs through `EditorViewController.mediaUploadDelegate` — a host object retaining the view controller forms `EditorViewController → delegate → EditorViewController` regardless of how the server's container holds it. What it did buy was the reference vanishing mid-request.

### 1. The server and the editor both own the delegate

**MediaUploadServer.swift**: `UploadContext` holds the delegate strongly and becomes a `struct`, dropping its `@unchecked Sendable` opt-out — `MediaUploadDelegate` is `Sendable` and `DefaultMediaUploader` is `@unchecked Sendable`, so it is implicitly Sendable. Immutable strong references make the three reads agree by construction.

**EditorViewController.swift**: `mediaUploadDelegate` becomes strong, moves into `init`, and becomes `public private(set)`. It only ever took effect if it was in place before the editor loaded — a contract enforced at runtime by a `precondition` on the setter. Taking it at construction makes that failure unrepresentable rather than caught, so `hasStartedLoading`, `mediaUploadDelegateWasAssigned` and the released-before-load trap go with it.

SwiftLint's `weak_delegate` is suppressed with the reasoning inline. The rule is arguably right that the name no longer fits; #685 renames the property and the suppression goes away with it.

### 2. `stopMediaHandling()`, for the cycle this creates

A host object that owns the editor and is also its delegate closes a cycle ARC cannot break, so `deinit` never runs and every editor opened strands a bound loopback `NWListener` with a live token. `stopMediaHandling()` stops the server, drops the delegate, and withdraws the endpoint from the page.

It has to be the host's call — **not** because UIKit can't report a teardown. It can: an ancestor walk (WordPress-iOS already ships `isBeingDismissedDirectlyOrByAncestor()`) plus an orphan check at `viewDidDisappear` fires on every dismissal and pop across fourteen hosting shapes, including this editor's shape in WordPress-iOS, with no false positive on being covered, tab-switched, re-parented, or left by a cancelled interactive pop. What is *not* observable is whether a detachment is **permanent**. A host may re-present the same editor, and since this call is terminal, guessing wrong disables media in an editor that survived.

Nothing here hands the delegate the editor — every value crossing that boundary is a value type — so the cycle is entirely host-authored, and the rule is narrower than "don't retain the editor": don't conform the object that owns it. That costs nothing, because `processFile` runs off the main actor and could not have reached that object's state anyway.

### 3. Stopping withdraws the endpoint from the page

Found while reviewing this, and the reason `stopMediaHandling()` is safe to expose at all. The port and token are injected once at document start, and `nativeMediaUploadMiddleware` refuses to retry a failed native upload directly, on the stated assumption that an advertised port is a reachable one (*"cleared on stop"*). Nothing cleared it. A stopped server left every image insert failing with a connection error on a working connection.

`revokeNativeUploadEndpoint()` clears all three copies: the live page, the `localStorage` copy `getGBKit()` falls back to, and the injected user script that would otherwise restore the dead port at the next document start — including the reload that recovers a terminated WebContent process.

### 4. Teardown that releases what it should

- **`HTTPServer.stop()` clears the listener's `newConnectionHandler`.** That block holds the request handler and, through it, the delegate — and `cancel()` alone does not drop it, so Network.framework performed the final release on its own queue after cancellation completed. A delegate reached through `EditorViewController.deinit` deallocated off the main thread in 46 of 50 measured runs. This makes teardown synchronous for an **idle** server; a request in flight holds its own copy until it unwinds, which the docs now say on all three surfaces a host reads.
- **Cancellation is checked before delivery**, in `processAndUpload` and before the passthrough forward, so a torn-down editor can't put bytes on the wire. The guarantee comes from the server rather than from URLSession noticing — `URLSessionProtocol` is public, and the obvious conformance for a host wrapping a callback-based stack has no cancellation awareness at all.
- **`startUploadServer()` stores the server only if the delegate is still there.** The assignment landed after the `await`, so a `stopMediaHandling()` inside the bind window was undone. That window measures at or under a millisecond and has no callers yet, so this is an invariant rather than a bug report: `startUploadServer()` has one call site and runs at most once per editor, so with the guard, "terminal" is a property of the code rather than of how fast a listener binds.

### 5. Making the mistake findable

A DEBUG-only census counts live upload servers and logs a fault past four, naming the delegate type. Each live server is a bound loopback listener, one per editor, so monotone growth is this cycle and nothing else produces it — `warmup()` passes no delegate and starts no server. It is the only detectable symptom: a `deinit` assertion cannot fire, because a cycle is what stops `deinit` from running. Logged, never fatal.

The rule also moved to where hosts read it: a `- Parameters:` block on `init`, which had no doc comment at all, and a Media Handling section in `docs/integration.md` with the leaking shape and the leaf-object fix side by side.

## Compatibility

`mediaUploadDelegate` is no longer settable — pass it to `init` instead. WordPress-iOS has no `mediaUploadDelegate` call sites, and both of its `GutenbergKit.EditorViewController(...)` call sites use argument labels, so the added parameter is source-compatible there.

Android needs no counterpart: its `uploadDelegate` is already a plain `val`, and a tracing GC collects the cycle this one cannot.

## Testing Instructions

`doesNotStronglyRetainDelegate` pinned the invariant being removed, so it is replaced by `retainsDelegateForServerLifetime`, asserting both halves — the server owns the delegate while it runs, and releases it afterward. `stopReleasesDelegateThatRetainsTheServer` covers the server-side cycle, `processesForHostReleasedDelegate` covers the original bug, and `EditorViewControllerMediaTeardownTests` covers both editor shapes.

- [x] Every ownership assertion mutation-checked. `UploadContext(uploadDelegate: nil, …)` fails `retainsDelegateForServerLifetime` and `stopReleasesDelegateThatRetainsTheServer`; dropping `mediaUploadDelegate = nil` from `stopMediaHandling()` fails `stopMediaHandlingBreaksTheOwnershipCycle`; no-op'ing `releaseConnectionHandler()` still fails the release half
- [x] `swift test` — 395 tests, host suite green
- [x] iOS Simulator `xcodebuild test` — 583 + 395 green, including the UIKit-gated teardown tests the host suite compiles out
- [x] Release build for `generic/platform=iOS`, so the DEBUG census compiles out cleanly
- [x] `xcodebuild build` of the iOS demo app, which no CI step covers
- [x] SwiftLint clean

Two paths are deliberately untested, both for the same reason: the cancellation checks and the bind-window guard each need teardown driven between two points of a live in-flight request, and a timing-based approximation would pass whether or not it reached the window — the failure mode `retainsDelegateForServerLifetime` was already bitten by.

## Related issues

- #685 renames the property and drops the `AnyObject` constraint, so a host can conform with a value type that cannot participate in a cycle at all


ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -109,28 +109,27 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
...
+    /// a retain cycle and neither is freed.
+    // Ownership here is the point: the editor holds this for its lifetime so an
+    // in-flight upload can't lose the delegate mid-request. `weak_delegate` is not
+    // wrong about the risk it names: strong here is precisely what lets a delegate
+    // that retains the editor back close a cycle ARC cannot break, and `weak` would
+    // rule that out. It is a deliberate trade — losing the delegate mid-request was
+    // the failure actually being hit — not an oversight. #630 drops the class
+    // requirement from the protocol so a host can conform with a value type.
+    // swiftlint:disable:next weak_delegate
+    public var mediaUploadDelegate: (any MediaUploadDelegate)? {
```
dcalhoun, line 131: *Finding from Claude:*

Is editor ownership a goal here, or only the fix for the mid-request read?

If it's only the fix, reading the delegate once at the top of `handleUpload` and passing it down keeps the whole request on one delegate while the property stays weak. If it's a goal (assign-and-forget, value-type conformers in #630, Android parity), it brings the cycle from Known issue, and #630 and #631 already build on it.

**If the editor keeps ownership**
- A host that hits the cycle has no way out. The setter traps after load, and the server holds its own reference, so clearing the property wouldn't break the cycle anyway. Is the doc warning enough, or is an explicit teardown worth adding?
- `releaseConnectionHandler` doesn't cover an in-flight upload. The task holds the delegate until `processFile` returns, so when the editor is its only owner (as the docs invite), the last release can still land off main. Worth noting on the property, and scoping `releaseConnectionHandler`'s doc to an idle server.

**If we read once per request**
- Revert the property and `UploadContext` to weak. #630 keeps `: AnyObject`, and #631's `Handler` holds the processor weakly.
- `processesForHostReleasedDelegate` would need to drop the host's reference inside `handlesFile`, since dropping it before the request should pass through under a weak contract.
- Restore `doesNotStronglyRetainDelegate`. `releaseConnectionHandler` becomes cleanup rather than a fix.

ios/Sources/GutenbergKitHTTP/HTTPServer.swift
```diff
@@ -305,15 +305,35 @@ public final class HTTPServer: Sendable {
...
+    /// Drops the connection handler so teardown releases what it captured *here*,
+    /// on the caller's thread.
+    ///
+    /// `newConnectionHandler` retains the request handler, and through it whatever
+    /// the caller's closure captured. `cancel()` alone does not drop the block:
+    /// Network.framework holds the listener until cancellation completes on its own
+    /// queue, so the final release — and therefore the captured object's `deinit` —
+    /// lands there rather than wherever `stop()` was called. For GutenbergKit's
+    /// upload server that means a host's media handler could be deallocated off the
+    /// main thread on a path that started in `EditorViewController.deinit`.
```
dcalhoun, line 328: *Finding from Claude:*

Nit: `GutenbergKitHTTP` doesn't otherwise know about its consumers, and the preceding sentence already covers the effect.

```suggestion
    /// lands there rather than wherever `stop()` was called.
```

dcalhoun (commented):
I feel this type of design decision is worth @crazytonyli's review. 

The alignment with Android and proposed changes make sense.  I tested these changes in the Demo app using #629. 

I captured findings by Claude, which I believe are legitimate and worth considering.

PR #627, opened 2026-09-05
refactor: rename DefaultMediaUploader to InternalMediaClient

jkmassel:
Stacked on #626. Fourth of ten PRs splitting #621. No behavior change beyond error-message text.

## What?

`DefaultMediaUploader` → `InternalMediaClient` on both platforms, and the `defaultUploader` parameters and properties that carry it.

## Why?

The name reads as an implementation of a host-facing protocol — the "default" one, as against a host's. It is not. It is GutenbergKit's own HTTP client for the configured site: it performs the uploads no host took over, and it relays every media delete, because the editor only ever asks to delete `/wp/v2/media/<id>` on the configured site.

The name gets actively confusing in #628, which adds a `MediaUploader` protocol. Renaming first keeps that PR's diff about the new API rather than about disambiguating two similar names.

## How?

Mechanical, ~21 sites on trunk plus tests, plus a sweep of the prose and error strings that used the retired vocabulary — including the `UploadContext` doc header and Android's three media-client messages. The host-facing docs still say "the default uploader" as a role: `InternalMediaClient` is internal on both platforms, so naming it in prose a host reads would be worse.

On iOS it also narrows two signatures: `passthroughResponse` and `handleDelete` took the whole `UploadContext` and touched only the client, so they now take it directly.

On the delete path that is more than tidiness — a deletion always relays to the configured site, never to a delegate. That was a convention the signature let you break; now the type won't. The three functions that keep the context genuinely need every field.

Android's server holds the client as a constructor property rather than threading a context, so it needs the rename only — and because its `handleDelete` is an instance method with the delegate in scope, the delete-path convention stays a convention there. The type-level guarantee is iOS-only.

## Testing Instructions

- [x] `swift test` — host suite green (969 tests, 57 suites)
- [x] Android `:Gutenberg:testDebugUnitTest` green
- [x] iOS Simulator `xcodebuild`
- [x] SwiftLint + Detekt clean


dcalhoun (approved):
Changes look good. I tested these changes in the Demo app using #629.

PR #628, opened 2026-09-05
feat: add MediaUploader, for a host that owns the whole upload

jkmassel:
Stacked on #627. Fifth of ten PRs splitting #621, and the heart of it.

**A host that doesn't adopt `MediaUploader` sees no behavior change.** `MediaUploadDelegate.uploadFile` still works and still returns what it returned; it gains a deprecation warning pointing at the replacement, and #629 removes it. What changes shape is internal: the upload server's startup gate now admits an uploader as well as a delegate, and the pipeline threads the delegate's metadata answer through to processing.

## What?

A `MediaUploader` protocol that takes over performing a media upload on the host's own stack, and owns its whole lifecycle: retries, recovery, and cleanup.

## Why?

Performing a media upload — and retrying it — should be a single, all-or-nothing responsibility: either GutenbergKit performs the upload and owns its retries, or the host does. Both go to the same configured site; the only difference is who executes the requests.

`MediaUploadDelegate.uploadFile` doesn't offer that. A host performs the `POST /wp/v2/media` and returns the raw response it received — then the editor, reading that response, drives the `post-process` retries and the orphan cleanup behind it, **through the WebView rather than the host's stack**. A host that took over uploads to run them through its own networking still didn't own the retries. It also receives no form fields, so an attachment it uploads lands unattached to its post.

Neither is fixable while the hook returns a raw response, which is what the replacement changes.

### `MediaUploader`

`upload(_:)` returns the finished attachment or throws. There is no raw response left for the editor to retry behind it, so the host drives its own `post-process` recovery and force-deletes its own orphan on terminal failure.

### `MediaUpload`

Carries the file, its metadata, the editor's non-file form fields (`post`, additionalData) and the request query (`?_embed`) — everything needed to reproduce a native request.

### `MediaUploadField`

Fields are an ordered list rather than a dictionary, so repeated names (a `field[]` array) survive verbatim and in order. A named type rather than a tuple: tuples are not nominal, so a tuple-typed property would permanently block `Equatable`/`Hashable`/`Codable` synthesis on `MediaUpload`, and that is not fixable later without a source break for every host.

### Precedence

`uploadFile` still works and is marked deprecated, pointing hosts at the replacement; an uploader takes precedence when both are set. #629 removes the old hook, so hosts get a migration window rather than a flag day.

This deliberately leaves **one deprecation warning** in GutenbergKit's own build, at the call site that supports the old hook (`MediaUploadServer.swift:397`). The marker exists to tell *hosts* to migrate, and supporting the hook until it is removed means calling it. It goes away with #629.

### What the metadata gate decides

`handlesFile` is the delegate's cheap, type-only veto, consulted before the upload is copied to a temp file. With an uploader set it no longer decides whether the upload happens — an uploader takes over delivery for **every** file, so there is no passthrough left to decline to.

It still decides whether `processFile` runs. A file the delegate declined by type is delivered to the uploader **unprocessed**, rather than handed to a delegate that just said it won't touch a file like that — an image-only delegate never sees a `.mov`. `handlesFile` is consulted exactly once per upload, on both platforms.

### Cancellation before delivery

Android now checks for cancellation immediately before handing the file to the uploader, matching the iOS check #626 added. A host uploader that isn't cancellation-cooperative — a background service, a work queue — would otherwise finish a `POST` for an editor that is already gone, leaving an attachment that, by this protocol's own contract, nobody cleans up.

### Recovery is the host's, and it needs one parameter that isn't obvious

When `POST /wp/v2/media` fatals in server-side post-processing it returns a 5xx carrying the attachment ID in `x-wp-upload-attachment-id` — the attachment exists but is unfinished. GutenbergKit is out of the network for an uploader host, so the editor-side middleware never sees that 5xx and cannot recover it.

The protocol doc spells out the recipe, including the part that bites: `POST /wp/v2/media/<id>/post-process` needs a body of `{"action": "create-image-subsizes"}`. Core registers `action` as **required**, so a recovery loop that omits it returns 400 on every attempt and then force-deletes an attachment `wp_update_image_subsizes()` would have finished.

## Testing Instructions

Five tests on iOS and six on Android. Both platforms cover delivery, form fields and query, precedence over `uploadFile`, the declined-file case, and a terminal throw surfacing without GutenbergKit re-delivering. Android adds two for the startup gate — that `mediaUploader` alone brings the server up, and that assigning it after load traps — which iOS cannot cover from a test target, because `EditorViewController` is behind `#if canImport(UIKit)`.

- [x] Build [#2715](https://buildkite.com/automattic/gutenbergkit/builds/2715) green — 18/18 checks, including `swift-ios-simulator-tests`, `android-test-android-library`, and both E2E suites
- [x] `swift test` — 974 tests, 0 failures
- [x] `:Gutenberg:testDebugUnitTest` — 672 tests, 0 failures
- [x] iOS Simulator `xcodebuild` — compiles the UIKit-gated `EditorViewController`, which the host test target does not
- [x] SwiftLint + Detekt clean; `:Gutenberg:compileDebugUnitTestKotlin` emits no warnings

To exercise the recovery path against a real fatal, `make wp-env-media-failure MODE=always` forces `wp_generate_attachment_metadata()` to fail — see `docs/code/local-wordpress.md`. Note that a `MediaUploader` host is responsible for the retry loop itself, so this exercises the host's implementation, not GutenbergKit's.

`MediaUploadServerTest` crosses Detekt's `LargeClass` threshold; baselined rather than split, which is its own change.


adalpari (approved):
It looks good to me.

Claude found this nuances on Android. Not a blocker, though.

> Uploader-only hosts silently get no upload server — android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt:707 (same on iOS, EditorViewController.swift:488)
The start guard still bails on empty authHeader/siteApiRoot. With an uploader set, mediaUploader is then never called, uploads fall back to the WebView path, and nothing is logged. The guard’s rationale (“nothing to upload through”) no longer applies once an uploader exists.

> Android cancellation gap remains in passthroughResponse — android/Gutenberg/src/main/java/org/wordpress/gutenberg/MediaUploadServer.kt:403
The PR closes the gap in processAndUpload but not here, where iOS has a second Task.checkCancellation(). A cancelled request can still enqueue the passthrough POST and orphan an attachment.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -715,6 +733,7 @@ class GutenbergView : FrameLayout {
             uploadServer = MediaUploadServer(
                 uploadDelegate = mediaUploadDelegate,
                 internalClient = internalClient,
+                uploader = mediaUploader,
```
dcalhoun, line 736: *Finding from Claude:*

With `mediaUploader` set, a server that doesn't come up still falls back to the WebView uploading directly, so the uploader is bypassed without any signal to the host. That happens here when cleartext to localhost isn't permitted (`:716`, the default at `targetSdk` 28+ without a network-security entry) or this constructor throws (`:742`), and on iOS when `MediaUploadServer.start` throws or times out (`EditorViewController.swift:508`). WordPress-Android isn't affected, since its `base-config` permits cleartext, but other hosts would be.

#632 traps the missing-credentials case because falling back "would silently drop" the uploader. Should these paths follow suit for an uploader host? A `check` fits the cleartext case, which is configuration like #632's, and a runtime start failure could fail uploads rather than bypass the uploader.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/MediaUploadServer.kt
```diff
@@ -462,7 +573,30 @@ internal class MediaUploadServer(
...
+            // never runs its own recovery behind it.
+            uploader?.let { hostUploader ->
+                val upload = MediaUpload(
+                    file = targetFile,
+                    mimeType = targetMimeType,
+                    filename = targetFilename,
+                    fields = formFields(extraParts),
+                    query = query
+                )
+                return UploadResult.Uploaded(MediaUploadResponse(201, hostUploader.upload(upload)))
```
dcalhoun, line 594: *Finding from Claude:*

If a host's own `withTimeout` fires inside `upload`, its `TimeoutCancellationException` is a `CancellationException`, so `processAndRespond` rethrows it as teardown (`:507`). The socket closes without a response, and after the server's 5s idle timeout the editor shows "Could not get a valid response from the server." iOS returns a 500 with the message in the same case.

Rethrowing only when this coroutine is actually cancelled would fix it, replacing the catch body at `:507`:

```kotlin
} catch (e: kotlin.coroutines.cancellation.CancellationException) {
    currentCoroutineContext().ensureActive()
    Log.e(TAG, "Upload failed", e)
    return errorResponse(500, e.message ?: "Upload failed")
}
```

<details>
<summary>Repro: local, uncommitted test on #629's head (<code>d8b85e24</code>)</summary>

```kotlin
private class TimingOutUploader : MediaUploader {
    override suspend fun upload(upload: MediaUpload): ByteArray = withTimeout(50) { awaitCancellation() }
}
```

A `POST /upload` with this uploader gets no status line, and the connection closes after 5s. The same request with an uploader that throws `IOException` gets a 500 carrying its message in 6ms.

</details>

ios/Sources/GutenbergKit/Sources/Media/MediaUploadDelegate.swift
```diff
@@ -94,7 +103,98 @@ extension MediaUploadDelegate {
...
+    /// core recovers its own uploads (up to 5 attempts), then return the finished
+    /// attachment. That request needs a body of `{"action": "create-image-subsizes"}`
+    /// — core registers `action` as **required**, so a post-process request without
+    /// it fails with a 400 every time rather than recovering.
+    ///
+    /// Owning the upload means owning cleanup on the server too: if post-process
+    /// can't be recovered, force-delete the orphan
+    /// (`DELETE /wp/v2/media/<id>?force=true`) before you `throw`, or it stays on the
+    /// site — neither GutenbergKit nor the editor cleans up behind you.
+    func upload(_ upload: MediaUpload) async throws -> Data
```
dcalhoun, line 199: *Finding from Claude:*

A throw is the only way to report failure, and every throw reaches the editor as a 500 `upload_error` carrying `error.localizedDescription` (`MediaUploadServer.swift:280`; Android uses `e.message`). A host that throws a plain Swift error for WordPress's `403 rest_cannot_create` gets "The operation couldn't be completed. (…)" in the editor's failure notice. A `LocalizedError` gets WordPress's message through, but its status and error code are lost either way, and #629 removes `uploadFile`, the one hook that relayed them.

Document the `LocalizedError` behavior here, or give hosts a way to hand back WordPress's status and body?

<details>
<summary>Repro: local, uncommitted tests on #629's head (<code>d8b85e24</code>)</summary>

An uploader throwing `WordPressRejection(code: "rest_cannot_create", message: "Sorry, you are not allowed to upload this file type.")`:

- As a plain `Error`: status 500, `code` is `upload_error`, and `message` starts with "The operation couldn" and omits WordPress's message.
- As a `LocalizedError` returning that message: status 500, `code` is `upload_error`, and `message` is WordPress's message.

</details>

ios/Sources/GutenbergKit/Sources/Media/MediaUploadDelegate.swift
```diff
@@ -94,7 +103,98 @@ extension MediaUploadDelegate {
...
+    public init(name: String, value: String) {
+        self.name = name
+        self.value = value
+    }
+}
+
+/// Everything a ``MediaUploader`` needs to reproduce a native upload: the file to
+/// send, its metadata, the editor's non-file form fields, and the request's query.
+public struct MediaUpload: Sendable {
+    /// The file to upload — already processed, if a ``MediaUploadDelegate`` ran.
```
dcalhoun, line 134: *Finding from Claude:*

The protocol docs suggest a background session or an offline queue, but `fileURL` is a temp file deleted as soon as `upload` returns, throws, or is cancelled (`MediaUploadServer.swift:172` and `:331-335`; Android's `finally` at `MediaUploadServer.kt:519`). A host that hands the URL to a persistent queue can find it gone by the time the job runs. `upload(_:)` could also use a line on cancellation: honor it, and delete an attachment created after it, since nothing else will.

```suggestion
    /// The file to upload — already processed, if a ``MediaUploadDelegate`` ran.
    ///
    /// A temporary file, deleted as soon as ``MediaUploader/upload(_:)`` returns,
    /// throws, or is cancelled. Copy it first if your upload needs it after that.
```

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -130,26 +134,41 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
...
-                "mediaUploadDelegate must be set before the editor loads (e.g. right after init). "
-                    + "It is captured into the editor configuration at load; setting it afterward has no effect."
-            )
+            precondition(!hasStartedLoading, Self.lateMediaAssignmentMessage("mediaUploadDelegate"))
         }
     }
 
+    /// Takes over media upload on the host's own stack (background session, offline
+    /// queue, resumable transport). Setting it makes the host own every upload and its
+    /// whole lifecycle; GutenbergKit stays out of the network entirely for media.
```
dcalhoun: *Finding from Claude:*

Is "out of the network entirely" accurate? Media deletes still go through `InternalMediaClient` with an uploader set (`MediaUploadServer.swift:104`), which is why #632 requires credentials for an uploader host. A host adopting `MediaUploader` to log every request, the protocol's own example, won't see them. The same wording is on `GutenbergView.mediaUploader` and on `MediaUploader` on both platforms.

```suggestion
    /// whole lifecycle. The editor's media deletes still go through GutenbergKit.
```

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -130,26 +134,41 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
...
+    ///
+    /// Same lifecycle rules as ``mediaUploadDelegate``: set it before the editor loads.
+    /// The editor owns it for its lifetime (releasing it on `deinit`), so you needn't
+    /// retain it yourself — just don't strongly retain this `EditorViewController`
+    /// from your uploader.
+    ///
+    /// Takes precedence over the deprecated ``MediaUploadDelegate/uploadFile(at:mimeType:filename:)``:
+    /// with an uploader set, that hook is never called.
+    public var mediaUploader: (any MediaUploader)? {
```
dcalhoun, line 152: *Finding from Claude:*

This inherits the ownership question from the #625 thread, so the decision there should cover both properties. An uploader is the likelier of the two to reference the editor's owner and close the cycle.

The off-main release point applies here too. A local, uncommitted test on #629's head (`d8b85e24`) stops the server on the main thread while `upload` is still running: the uploader outlives `stop()`, then deallocates off the main thread once `upload` returns. The same holds for a delegate still in `processFile`. An idle server does release it on main, so `releaseConnectionHandler` covers only the case with no upload in flight.

android/Gutenberg/src/main/java/org/wordpress/gutenberg/GutenbergView.kt
```diff
@@ -123,14 +123,32 @@ class GutenbergView : FrameLayout {
...
+     * Same lifecycle rules as [mediaUploadDelegate]: set it before the editor loads,
+     * and this view owns it for its lifetime — so you needn't retain it yourself, just
+     * don't strongly retain this [GutenbergView] from your uploader.
+     *
+     * Takes precedence over the deprecated [MediaUploadDelegate.uploadFile]: with an
+     * uploader set, that hook is never called.
+     */
+    var mediaUploader: MediaUploader? = null
+        set(value) {
+            check(!hasStartedLoading) { lateMediaAssignmentMessage("mediaUploader") }
```
dcalhoun, line 144: *Finding from Claude:*

Nit: the `hasStartedLoading` docs (`:155-157` here, `EditorViewController.swift:107-109`) still name only the delegate, though a late `mediaUploader` write now throws (traps on iOS) too.

dcalhoun (approved):
I tested these changes in the Demo app using #629. 

I captured a few findings by Claude that seem worthwhile.

PR #629, opened 2026-09-05
feat!: remove MediaUploadDelegate.uploadFile

jkmassel:
Stacked on #628. Sixth of ten PRs splitting #621. Almost entirely deletions.

## What?

Removes `uploadFile` from `MediaUploadDelegate` on both platforms. `MediaUploader` (#628) replaces it.

## Why?

Returning a raw response split one upload's HTTP across two owners — the host performed the `POST`, the editor drove the `post-process` retries and orphan cleanup behind it — and the hook received no form fields, so an attachment it uploaded landed unattached to its post. Neither is fixable while the hook returns a raw response.

What is left is a clean division: a delegate transforms bytes and GutenbergKit owns delivery and its retries; a `MediaUploader` owns delivery and its retries entirely. There is no longer an in-between where the host performs the upload but the editor retries it.

## How?

- The protocol requirement, its default implementation, and the delivery branch that called it, on both platforms.
- `handlesFile` no longer gates the temp copy for two callers, only for `processFile`. With a `MediaUploader` set it can't decline the upload itself — an uploader delivers every file, so there is no passthrough to fall to — but it still gates `processFile`, so a declined file reaches the uploader unprocessed. Pinned by `an uploader sees a file the delegate's metadata gate would have declined` on both platforms.
- `MediaUploadResponse` drops to internal on both platforms: `uploadFile` was the only public API that named it. This is load-bearing for migration safety — Swift requires an explicit return type, so a host's stale `uploadFile` cannot compile without naming the type. Leave it `public` and the same removal breaks hosts silently instead.
- The deprecation warning #628 left in GutenbergKit's own build goes away with the hook.

A second commit corrects the docs the removal left stale: the `mediaUploadDelegate` property summaries — what Xcode Quick Help and IDE hover show — still advertised "custom upload" on both platforms and pointed nowhere at `MediaUploader`, and `UploadError.noUploader`'s message still named the delegate as an alternative to the internal client.

## Testing Instructions

Tests for the removed path are rewritten rather than deleted — the delegate now processes and the internal client delivers, which is what the remaining path does.

- [x] `swift test` — 974 tests in 57 suites, green
- [x] `xcodebuild test -scheme GutenbergKit-Package -sdk iphonesimulator` — 585 tests in 36 suites, green. Covers the UIKit-gated code the host `swift test` compiles as empty, including `EditorViewController`.
- [x] `:Gutenberg:testDebugUnitTest` — 672 tests, 0 failures
- [x] `:app:assembleDebug` — Android demo builds
- [x] `swift build --build-tests` — zero warnings
- [x] SwiftLint and Detekt clean

## Breaking change

Hosts implementing `uploadFile` must conform to `MediaUploader` instead. Hosts that only implement `processFile` / `handlesFile` are unaffected. Migration needed in WordPress-iOS, WordPress-Android and Jetpack.

The break is a hard compile error on both platforms, not a silent redirect — Android on `override` of nothing, iOS on `MediaUploadResponse` no longer being visible. `uploadFile` shipped in `v0.20.0-alpha.0` and no stable tag.


ios/Sources/GutenbergKit/Sources/Media/MediaUploadServer.swift
```diff
@@ -356,12 +356,7 @@ final class MediaUploadServer: Sendable {
             return .uploaded(MediaUploadResponse(statusCode: 201, body: attachment))
         }
 
-        // The deprecated delegate path: the host performs the POST but returns the raw
-        // response, leaving the editor to drive post-process recovery behind it.
-        if let delegate = context.uploadDelegate,
-           let result = try await deprecatedUploadFile(delegate, uploadURL, uploadMimeType, uploadFilename) {
-            return .uploaded(result)
-        } else if let internalClient = context.internalClient {
+        if let internalClient = context.internalClient {
```
dcalhoun, line 359: *Finding from Claude:*

Nit: with the `uploadFile` read gone, `UploadContext`'s doc (`:510`) still cites "the three reads within a request". Two remain: `handlesFile` and `processFile`.

dcalhoun (approved):
Looks good. I tested these changes in the Demo app.

PR #630, opened 2026-09-05
refactor!: rename MediaUploadDelegate to MediaProcessor, and drop its class bound

jkmassel:
Stacked on #629. Seventh of ten PRs splitting #621.

Two breaking changes — the rename and the class-bound relaxation that follows from it — plus two `fix(ios):` commits an adversarial review turned up. Both fixes are real behavior changes, so they are called out below rather than left to be found in the diff.

> **Note for release notes:** a PR lands in one category and `[Type] Breaking Change` wins, so **"start the upload server for a host that supplies only an uploader" will not appear under Bug Fixes.** A host that hit that bug — the host's `upload(_:)` silently never called — should be pointed here.

## What?

`MediaUploadDelegate` → `MediaProcessor`, `mediaUploadDelegate` → `mediaProcessor`, the server parameter → `processor`, the file → `MediaHandlers.swift`, and Android's demo → `DemoMediaProcessor`.

## Why?

The protocol no longer uploads anything — #629 removed `uploadFile`, leaving `handlesFile` and `processFile`. "UploadDelegate" now describes the one thing it can't do, and next to `MediaUploader` the two names read as variations on the same job rather than the two halves of a deliberate split.

`MediaProcessor` says what is left: it transforms bytes, GutenbergKit delivers them.

## How?

A rename sweep, ~190 sites across both platforms including the demo apps, `docs/integration.md`, and the media test suites' own vocabulary — test function names, `@Test` display strings and Kotlin backtick names are what CI prints, and they named a type this stack deletes. Prose in doc comments follows the types, including the property abstracts Xcode Quick Help and Android Studio hover actually render. `MediaHandlers.swift` because the file holds both protocols now.

The `weak_delegate` suppression added in #625 goes away with the name: the rule was arguably right that a strongly-held "delegate" is a smell, and the answer was that this was never a delegate. The rule keys on the identifier suffix, so it cannot fire on `mediaProcessor` — the suppression and the paragraph arguing with it are both dead, and nothing in CI would ever have said so.

Following that through: `MediaProcessor` and `MediaUploader` also **drop `: AnyObject`**. Nothing needed class-boundness — there is no `weak`, `===`, or `ObjectIdentifier` use against either protocol anywhere in the tree — and `EditorViewController` holds both strongly, so a class-bound protocol was quietly steering hosts toward a conformer that holds the view controller back and closes a retain cycle ARC cannot break. Dropping it lets a host conform with a value type capturing only what the work needs. Every existing conformer is a class and is unaffected.

Two fixes ride along, each in its own `fix(ios):` commit rather than buried in the rename.

**An uploader-only host never got an upload server.** `startUploadServer()` asks "did the host supply a handler" twice — once before starting, once after the bind returns, because `stopMediaHandling()` can land across that `await`. The two reads had drifted: #628 widened the first to cover `mediaUploader` and left the second checking the processor alone. A host passing only an uploader bound a listener, failed the second check, and stopped the server it had just started — so `nativeUploadPort` was nil, `api-fetch.js` fell through to the WebView path, and `upload(_:)` was **never called for any file**, silently. Both reads now go through one `hasMediaHandling`. Android already pinned this gate; iOS had no equivalent, which is why it survived three commits green.

**The leak census named the wrong object.** It resolved as `processor ?? uploader`, so with both set it always accused the processor — which, after the class bound comes off, may be a value type holding nothing at all. It now names every handler supplied.

## Testing Instructions

- [x] iOS Simulator `xcodebuild test` — 585 + 395 tests green
- [x] Android `:Gutenberg:test` green (re-run with `--rerun-tasks`); Android and iOS demo apps compile
- [x] SwiftLint 0 violations across 152 files; Detekt clean
- [x] `docs/integration.md` swept — `git grep MediaUploadDelegate` over the tree now returns nothing
- [x] New `uploadServerStartsForAnyHandler` test mutation-checked — restoring the old post-bind guard fails **only** the uploader-only case
- [x] New `valueTypeProcessorRuns` test mutation-checked — re-imposing `: AnyObject` fails to compile (`non-class type 'ValueTypeProcessor' cannot conform to class protocol`)

## Breaking change

`mediaUploadDelegate` is now `mediaProcessor`, and `MediaUploadDelegate` is `MediaProcessor`. Conformances need no changes beyond the name.

`MediaProcessor` and `MediaUploader` are also no longer `AnyObject`-bound. Class conformers are unaffected; a host that declared its own `weak` reference to one of these existentials would need to hold it strongly instead.

Note this relaxes a constraint rather than fixing the cycle outright — a `struct` that stores the `EditorViewController` cycles just the same. The doc comments say so rather than implying the type system settles it.

A value-type conformer also carries a caveat a class did not: it is **copied** when you hand it to the initializer, and the editor holds that copy for its lifetime. Mutating your own instance afterwards changes nothing the editor will run, and there is no way to swap in a new value — the property is `private(set)`, so a different processor means a different editor. If you need settings the host can change mid-session, read them inside `processFile` through a reference the conformer captures. Documented on both protocols.


ios/Sources/GutenbergKit/Sources/Media/MediaHandlers.swift
```diff
@@ -55,7 +55,41 @@ public enum ProcessedProxyFile: Sendable {
...
+/// processor that is never called. That
+/// reference must itself be `Sendable` — an actor, or a class made safe with a lock —
+/// because this protocol is `Sendable` and a `struct` conformer's stored properties
+/// inherit that requirement.
+///
+/// Two requirements a `struct` makes easy to miss, both of which compile silently:
+/// `processFile` cannot be `mutating` (a `mutating` witness does not satisfy a
+/// non-mutating requirement), and its argument labels must match exactly. Either
+/// mistake resolves to the no-op default below instead of failing to build, leaving a
+/// processor that is never called.
```
dcalhoun: There appears to be erroneous duplication between these comments. 

Atop that, _all_ of the comments in this PR feel verbose. Can we make them more concise while retaining the important pieces? In general, I feel comments should only explain "why" for code that is unclear, we should avoid letting agents explain the reasoning for how they arrived at the code. 

Maybe have an agent reassess and update comments with this guidance? WDYT?

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -579,12 +594,12 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
...
     /// because `nativeUploadPort` will be nil in GBKit).
-    private func startUploadServer() async {
+    func startUploadServer() async {
         // Nothing to route through the native server unless the host provided a
-        // delegate or an uploader. The editor owns whichever it was given — both
+        // processor or an uploader. The editor owns whichever it was given — both
         // properties are strong — so there's no released-before-load case to guard
         // against; they live as long as it does.
-        guard mediaUploadDelegate != nil || mediaUploader != nil else {
+        guard hasMediaHandling else {
```
dcalhoun, line 602: This appears technically correct, but it may not be worth addressing given current use cases. 

*Finding from Claude:*

The handler gate widened here, but the credentials guard just below (606-619) did not — it still returns when `siteApiRoot`/`authHeader` are unusable. An uploader-only host has no reason to supply either (`MediaUploader`'s own doc promises GutenbergKit stays out of the network entirely for media), so it clears this guard and dies on the next one: server down, `nativeUploadPort: nil`, silent fallback to the WebView path. That is the failure this PR fixes, one guard lower — and both returns are silent, so nothing is logged.

Only `InternalMediaClient` needs those credentials, and `MediaUploadServer.start` already takes it as optional:

```swift
let internalClient = MediaServerCredentials.areUsable(
    siteApiRoot: configuration.siteApiRoot,
    authHeader: configuration.authHeader
) ? InternalMediaClient(...) : nil

guard internalClient != nil || mediaUploader != nil else { return }
```

`handleDelete` already guards on a nil client, so the delete path degrades rather than breaking.

ios/Tests/GutenbergKitTests/Media/EditorViewControllerMediaTeardownTests.swift
```diff
@@ -99,3 +145,29 @@ private final class StandaloneDelegate: MediaUploadDelegate {
...
 #endif
+
+/// Supplied only to bring the upload server up; never invoked by these tests.
+private struct InertUploader: MediaUploader {
+    func upload(_ upload: MediaUpload) async throws -> Data { Data() }
+}
+
+/// Whether `HTTPServer` can bind here — it cannot in some sandboxes, and these tests
+/// assert on a real listener.
+private let canBindUploadServer: Bool = {
```
dcalhoun: *Finding from Claude:*

This duplicates `_canStartUploadServer` in `MediaUploadServerTests.swift:7`, `UnsafeSendableBox`/`UnsafeMutableSendablePointer` and all — and both block a thread in `semaphore.wait()` while awaiting a `Task`.

`.enabled(if:)` is evaluated on the cooperative pool, so on a 2-core runner two suites probing at once can occupy both threads and leave neither inner `Task` anywhere to run: no signal, no timeout, hung suite. One probe was safe for exactly that reason; two may not be.

Worth lifting the existing one into `ios/Tests/GutenbergKitTests/Helpers/` and sharing it?

ios/Sources/GutenbergKit/Sources/Media/MediaHandlers.swift
```diff
@@ -83,7 +117,7 @@ public protocol MediaUploadDelegate: AnyObject, Sendable {
 }
 
 /// Default implementations.
-extension MediaUploadDelegate {
+extension MediaProcessor {
```
dcalhoun, line 120: May not be worthwhile if we are unconcerned with this edge case. 

*Finding from Claude:*

The `processFile` default predates this PR, but the new protocol doc now leans on it: "Either mistake resolves to the no-op default below instead of failing to build."

Could we delete that default (125-127) so it _does_ fail to build? A conformer that doesn't process files has nothing to do, and the silent miss isn't free at runtime: `handlesFile` still defaults to `true`, so the server materializes the whole part to a temp file (`MediaUploadServer.swift:219-225`), calls the no-op, gets `.original`, and throws the copy away — a full-size write and delete on every upload for a processor that never runs.

Dropping it turns both the `mutating` witness and the mistyped label into compile errors, which is the outcome the doc is asking prose to achieve.

ios/Sources/GutenbergKit/Sources/Media/MediaHandlers.swift
```diff
@@ -47,21 +47,55 @@ public enum ProcessedProxyFile: Sendable {
...
+/// reference must itself be `Sendable` — an actor, or a class made safe with a lock —
+/// because this protocol is `Sendable` and a `struct` conformer's stored properties
+/// inherit that requirement.
+///
+/// Two requirements a `struct` makes easy to miss, both of which compile silently:
+/// `processFile` cannot be `mutating` (a `mutating` witness does not satisfy a
+/// non-mutating requirement), and its argument labels must match exactly. Either
+/// mistake resolves to the no-op default below instead of failing to build, leaving a
+/// processor that is never called.
+public protocol MediaProcessor: Sendable {
```
dcalhoun: *Finding from Claude:*

Is a shim worth it here, so hosts get a fix-it instead of "cannot find type 'MediaUploadDelegate' in scope"?

```swift
@available(*, unavailable, renamed: "MediaProcessor")
public typealias MediaUploadDelegate = MediaProcessor
```

Same shape for `EditorViewController.mediaUploadDelegate`, and `@Deprecated(..., ReplaceWith("mediaProcessor"))` on `GutenbergView.kt`. Zero runtime cost, and it makes the migration one click per site in both host apps — or is the plan to land the break bare and fix up WordPress-iOS/-Android in the same cycle?

ios/Sources/GutenbergKit/Sources/EditorViewController.swift
```diff
@@ -168,7 +172,18 @@ public final class EditorViewController: UIViewController, GutenbergEditorContro
...
+    /// Read twice by `startUploadServer()` — once before starting, once after the bind
+    /// returns — and the two reads have to agree. They did not: the first gained
+    /// `mediaUploader` and the second was left checking the processor alone, so an
+    /// uploader-only host bound a listener, immediately stopped it, and fell back to the
+    /// WebView path with nothing logged. One property, so they cannot disagree again.
+    private var hasMediaHandling: Bool {
+        mediaProcessor != nil || mediaUploader != nil
+    }
+
+    private(set) var uploadServer: MediaUploadServer?
```
dcalhoun, line 186: May not be worth refactoring if we are less concerned about this edge case or the now public attributes. 

*Finding from Claude:*

This and `startUploadServer()` (597) both dropped `private` for the new tests. A second in-module call to `startUploadServer()` now overwrites `uploadServer` at 644 without stopping the previous one — stranding a bound loopback listener, which is the leak the census at `MediaUploadServer.swift:105` exists to catch.

`buildEditorConfiguration`'s `nativeUploadPort` (578) already reflects whether the server came up, and it is what the page actually consumes. Could the new tests assert on that instead and leave both members private?

ios/Tests/GutenbergKitTests/Media/EditorViewControllerMediaTeardownTests.swift
```diff
@@ -99,3 +145,29 @@ private final class StandaloneDelegate: MediaUploadDelegate {
 }
 
 #endif
```
dcalhoun, line 147: *Finding from Claude:*

`InertUploader`, `canBindUploadServer`, and `UnsafeSendableBox` land after this `#endif`, while their only call sites are the tests inside it.

`Package.swift` declares macOS too, so on that build the tests vanish and these three compile as dead code. And if `MediaUploader`/`MediaUploadServer` ever move under a UIKit guard — as `EditorViewController` and `MediaServerCredentials` already are — the macOS test target stops compiling on code that was meant to be excluded. Move the `#endif` to the end of the file?

ios/Tests/GutenbergKitTests/Media/EditorViewControllerMediaTeardownTests.swift
```diff
@@ -57,6 +57,52 @@ struct EditorViewControllerMediaTeardownTests: MakesTestFixtures {
...
+        )
+        defer { editor.stopMediaHandling() }
+
+        await editor.startUploadServer()
+
+        #expect(editor.uploadServer != nil, "\(label): no upload server, so the host's media handling never runs")
+    }
+
+    @MainActor
+    @Test("no handler leaves the upload server down", .enabled(if: canBindUploadServer))
```
dcalhoun: *Finding from Claude:*

This one asserts the server is _not_ started, so it never needs a bindable listener. Gating it means the sandbox the probe exists for also loses the one assertion that would still hold there — the check that an editor with no handlers doesn't bring a server up.

```suggestion
    @Test("no handler leaves the upload server down")
```

docs/integration.md
```diff
@@ -246,33 +246,34 @@ val configuration = EditorConfiguration.builder()
 
 ## Media Handling
 
-The host can customize how media is processed and uploaded by supplying a
-`MediaUploadDelegate` at init:
+The host can transform media before upload by supplying a `MediaProcessor` at init.
+To take over the upload itself, supply a `MediaUploader` instead — a processor only
+changes bytes; GutenbergKit still delivers them.
```
dcalhoun: *Finding from Claude:*

The doc doesn't mention the PR's other breaking change — that neither protocol is class-bound any more — nor the copy-at-`init` caveat the protocol docs now carry.

```suggestion
The host can transform media before upload by supplying a `MediaProcessor` at init.
To take over the upload itself, supply a `MediaUploader` instead — a processor only
changes bytes; GutenbergKit still delivers them. Neither protocol is class-bound, so a
`struct` conformer works; the editor copies it at `init`, so mutating your own instance
afterwards changes nothing the editor runs.
```

"Reusing a processor across editor sessions" (302-310) is reference-type guidance throughout — "keep your own reference, the editor drops only its own" buys a `struct` conformer nothing. Worth a qualifying line there too.

ios/Tests/GutenbergKitTests/Media/MediaUploadServerTests.swift
```diff
@@ -1085,12 +1125,23 @@ private final class DeclineByMetadataDelegate: MediaUploadDelegate, @unchecked S
   }
 }
 
-/// A delegate that produces a new file with changed metadata (e.g. a transcode).
-private final class ResizingDelegate: MediaUploadDelegate, @unchecked Sendable {
+/// A processor that produces a new file with changed metadata (e.g. a transcode).
+/// A value-type processor. `struct`, and `Sendable` without `@unchecked` — both are the
+/// point: this is the shape ``MediaProcessor``'s documentation now recommends.
+private struct ValueTypeProcessor: MediaProcessor {
```
dcalhoun, line 1131: *Finding from Claude:*

`ValueTypeProcessor` inherited `ResizingProcessor`'s doc line, so it now opens with two unrelated sentences and `ResizingProcessor` (1140) has none.

```suggestion
/// A value-type processor. `struct`, and `Sendable` without `@unchecked` — both are the
/// point: this is the shape ``MediaProcessor``'s documentation now recommends.
private struct ValueTypeProcessor: MediaProcessor {
```

The transcode line belongs back above `ResizingProcessor`.

ios/Sources/GutenbergKit/Sources/Media/MediaHandlers.swift
```diff
@@ -47,21 +47,55 @@ public enum ProcessedProxyFile: Sendable {
...
+/// processor that is never called. That
+/// reference must itself be `Sendable` — an actor, or a class made safe with a lock —
+/// because this protocol is `Sendable` and a `struct` conformer's stored properties
+/// inherit that requirement.
+///
+/// Two requirements a `struct` makes easy to miss, both of which compile silently:
+/// `processFile` cannot be `mutating` (a `mutating` witness does not satisfy a
+/// non-mutating requirement), and its argument labels must match exactly. Either
+/// mistake resolves to the no-op default below instead of failing to build, leaving a
+/// processor that is never called.
```
dcalhoun, line 91: There appears to be erroneous duplication between these comments. 

Atop that, _all_ of the comments in this PR feel verbose. Can we make them more concise while retaining the important pieces? In general, I feel comments should only explain "why" for code that is unclear, we should avoid letting agents explain the reasoning for how they arrived at the code. 

Maybe have an agent reassess and update comments with this guidance? WDYT?

dcalhoun (approved):
Changes look good. I tested uploads using the Demo app. I captured a few findings that are worth considering.

PR #639, opened 2026-09-09
fix(android)!: preserve the post title when the editor cannot be read

dcalhoun:
## What?

`GutenbergView.getTitleAndContent` reported a fabricated result when its JSON could not be parsed, letting hosts persist an empty post title over the user's own.

## Why?

**Live, silent data loss — confirmed on device, and confirmed reaching the server.**

The parse failure was logged and discarded, `changed` stayed `false`, and the callback fired normally:

```kotlin
} catch (e: JSONException) {
    Log.e("GutenbergView", "Received invalid JSON from editor.getTitleAndContent")
}
val title = lastUpdatedTitle ?: ""       // ← empty
callback.onResult(title, content)        // ← fires as if it succeeded
```

That path is reachable in normal use. When the `ErrorBoundary` catches, React unmounts the editor and deletes every `window.editor.*` method while `window.editor` survives as an empty object, so `evaluateJavascript` returns the string `"null"`. `isEditorLoaded` is never reset, so the call still goes out.

Because the callback fired normally, hosts could not tell a failed read from a title the user cleared. In WordPress-Android the empty title flows into `PostModel.setTitle("")`, persists to SQLite, and reaches the server on the next save. The post list keeps rendering the last synced title, so nothing looks wrong until the post is reopened.

[CMM-2008](https://linear.app/a8c/issue/CMM-2008) covers the iOS half of the same teardown, where the equivalent call throws instead — which happened to be protective.

## How?

Parse into a `Result` and report a new, required `TitleAndContentCallback.onError(error: Throwable)` rather than inventing a result. The error is an `EditorNotReadyException` when the editor has not loaded or the view has been detached, or the `JSONException` when its result could not be parsed — the same distinction iOS makes by throwing `EditorNotReadyError` or an invalid data format error.

**Breaking:** hosts must implement `onError`. A no-op default would compile without complaint and leave a host that waits only on `onResult` waiting forever. WordPress-Android needs a one-line `onError` that releases its latch when it upgrades.

The distinction that matters: separating _the read failed_ from _the title is empty_. A blanket empty-title guard in the host would have broken clearing a title legitimately, which is covered by a test.

`getTitleAndContent` also returned silently when `isEditorLoaded` was false, so a host awaiting a result had to wait out its own timeout, or never resumed at all. It now reports `EditorNotReadyException`, posted to the main thread like every other result. `isEditorLoaded` is `@Volatile`, since that check runs on the host's thread.

Detaching the view clears its queued callbacks and destroys the web view, which could drop a read's error or leave a read unanswered. Reads are now tracked until they report, and any still pending when the view detaches report `EditorNotReadyException`.

The demo app only resumed its `suspendCancellableCoroutine` from `onResult`, so it hung forever on a failed read; it now fails the save instead, and lets cancellation propagate rather than reporting it as a failure.

`isEditorLoaded` is still not reset when the editor crashes; that is [#640](https://github.com/wordpress-mobile/GutenbergKit/pull/640), which this PR is deliberately independent of.

## Testing Instructions

End to end, against WordPress-Android:

1. Open a post with a non-empty title and at least one block.
2. Trip the boundary from `chrome://inspect` (needs `setWebContentsDebuggingEnabled(true)` on a debug build):
   ```js
   const be = wp.data.select("core/block-editor");
   be.getBlockOrder = () => {
     throw new Error("repro");
   };
   wp.data.dispatch("core/block-editor").updateSettings({});
   ```
3. Close or save, then reopen the post.
4. Before: the title is gone and reaches the server on the next save. After: intact, and the save is skipped.
5. `adb logcat | grep GutenbergView` still shows `Received invalid JSON from editor.getTitleAndContent`, now with the cause attached.

N/A — no UI changes.

## Screenshots or screencast

N/A

https://claude.ai/code/session_01VuxMbKtUsaUF8nUVgKdxwK

### DIFF
--- a/android/Gutenberg/src/androidTest/java/org/wordpress/gutenberg/EditorBridgeTeardownInstrumentedTest.kt
+++ b/android/Gutenberg/src/androidTest/java/org/wordpress/gutenberg/EditorBridgeTeardownInstrumentedTest.kt
@@ -4,6 +4,7 @@ import android.webkit.WebView
 import android.webkit.WebViewClient
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
+import org.json.JSONException
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertTrue
 import org.junit.Test
@@ -12,15 +13,10 @@ import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 
 /**
- * Pins what a real [WebView] hands back once the editor's bridge is gone.
+ * Pins what a real [WebView] returns for [getTitleAndContentScript] once the
+ * editor's bridge is gone, and that [parseTitleAndContent] reports it as a failed
+ * read.
  *
- * When the editor's `ErrorBoundary` catches, React unmounts the editor and
- * `useHostBridge`'s cleanup deletes every `window.editor.*` method, while
- * `window.editor` survives as an empty object because it is assigned at module
- * scope. `GutenbergView.isEditorLoaded` is never reset, so `getTitleAndContent`
- * still reaches the web view in that state.
- *
- * The exact string returned here is what [parseTitleAndContent] has to reject.
  * A unit test asserting `"null"` in isolation would keep passing even if the
  * platform started returning something else, which is why this runs on a device.
  *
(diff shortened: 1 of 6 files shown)


adalpari:
Taking a look, it seems like there are a couple of scenarios that could lead into a confusing state. Could you take a look?

> 1. GutenbergView.kt:774 — the default no-op onError silently converts a wrong answer into a hang.
Adding a defaulted onError changes the contract from “always calls onResult” to “may call nothing”. A host that resumes its continuation only from onResult — exactly the pattern this repo’s own demo app used until this PR — now hangs forever on a failed read instead of resuming with a stale title. The compiler gives an unadopted host no signal at all. Only WordPress-Android’s timeout was verified as a backstop.
> 
> 2. GutenbergView.kt:857 — the error can be dropped entirely on teardown.
EditorNotReadyException is delivered via handler.post, but onDetachedFromWindow calls handler.removeCallbacksAndMessages(null). A last-chance save from onPause/onStop can have its queued onError discarded, leaving the host with no callback whatsoever — precisely the outcome this new error channel exists to prevent.

dcalhoun:
@adalpari thank you for the review. I addressed the two findings: 329e6eac463c98cc92a84ef02c260494d0aa74ff and 49f62f51d5769b24c370adb75e37c545c4bdcf7e. The former is a breaking change that will need to be addressed when we bump the GBK version in WordPress-Android. 

Ready for your review.

adalpari (approved):
LGTM!

PR #644, opened 2026-09-09
build!: sync the Android toolchain with WordPress-Android

dcalhoun:
## What?

Syncs the Android toolchain with WordPress-Android: AGP 8.10.1 → 9.3.1, Gradle 8.12.1 → 9.5.0, Kotlin 2.1.21 → 2.4.10.

> [!WARNING]
> **Host apps need Kotlin 2.3+.** The AAR's Kotlin metadata moves from 2.1.0 to 2.4.0, and it now depends on `kotlin-stdlib` 2.4.10. No API changes. WordPress-Android (Kotlin 2.4.10) is unaffected.

## Why?

Gradle refuses to run a composite build spanning two AGP versions, so every local WordPress-Android build that sets `localGutenbergKitPath` currently fails during configuration:

```
Using multiple versions of the Android Gradle Plugin [9.3.1, 8.10.1] across Gradle builds is not allowed.
Affected builds: [:, :android].
```

Anyone developing GutenbergKit against a local WordPress-Android checkout is blocked outright — the build fails before any task runs. Published builds are unaffected, which is why this went unnoticed for so long.

The drift began with [WordPress-Android#22713](https://github.com/wordpress-mobile/WordPress-Android/pull/22713) (merged 2026-03-24), which moved that repo from Gradle 8.12.1 / AGP 8.10.1 to 9.1.0 / 9.0.1. Dependabot has since carried it to 9.3.1 (#22735, #22791, #22837, #23105, #23146). GutenbergKit has no Gradle Dependabot coverage, so it never followed.

Follow-up: #645 adds automation so this drift is reported rather than discovered.

## How?

AGP 9 required source changes beyond the version bump:

- **Kotlin is built into AGP 9**, so `org.jetbrains.kotlin.android` now collides with the `kotlin` extension AGP registers. Plugin removed from both modules.
- **`applicationVariants` / `mergeAssetsProvider` are removed**, and passing a `Provider` to the source-set DSL is now an error rather than a warning. Both generated-source wirings — the demo app's OAuth assets and the library's locales generator — move to `androidComponents.onVariants` with `addGeneratedSourceDirectory`, which carries the task dependency itself. That replaces the task-name matching the locales generator previously relied on.
- **`kotlinOptions` is removed.** The Java 8 target still comes from `compileOptions`, so the bytecode target is unchanged (verified below).
- **Parcelize.** The sub-plugin bundled with AGP's Kotlin (2.2.10) only recognises the legacy `BaseExtension` DSL, so `ParcelizeSubplugin.isApplicable` returns false on AGP 9 projects. It silently adds neither the compiler plugin nor the runtime dependency, and `@Parcelize` classes fail to compile. Pinning the Kotlin plugins at the root puts a working Parcelize on the build classpath, matching how WordPress-Android does it.
- **`publish-to-s3`** looks up the legacy `com.android.build.gradle.LibraryExtension`, which AGP 9 no longer registers, so its `singleVariant("release")` call silently no-ops and the `release` software component never exists. Declared in the library instead, with a comment to remove it once the plugin supports AGP 9. Filed upstream as Automattic/publish-to-s3-gradle-plugin#42 — note that `0.11.0-rc1` does **not** fix this.
- **`android.nonTransitiveRClass` removed**; it is the default in AGP 9.

The **Gradle wrapper** is regenerated, so the jar and start scripts are 9.5.0's rather than 8.12.1's.

## Testing Instructions

1. `make build` (populates `supported-locales.json`, required by the library build)
2. `./android/gradlew -p ./android :Gutenberg:assembleRelease` — succeeds
3. `./android/gradlew -p ./android :app:assembleDebug` — succeeds
4. `make lint-android` — passes
5. Confirm the generated `SupportedLocales.kt` still reaches the published sources jar — this PR rewires how generated sources get there, and a plain build passes either way:

   ```
   ./android/gradlew -p ./android :Gutenberg:sourceReleaseJar
   unzip -l android/Gutenberg/build/intermediates/source_jar/release/release-sources.jar | grep SupportedLocales
   ```

To verify the composite build this PR actually fixes, point `local-builds.gradle` in a WordPress-Android checkout at this branch and run:

```
./gradlew :libs:editor:compileDebugKotlin
```

It fails on `trunk` with the AGP error above and succeeds on this branch.

### Notes for the reviewer

- **Bytecode target is unchanged.** `classes.jar` is still major version 52 (Java 8), so removing `kotlinOptions` only dropped redundancy.
- **The AAR now declares `minCompileSdk=34`** (previously 1). AGP 9 sets it to the library's `compileSdk`. It adds no new requirement, because `androidx.core` 1.13.1 and `androidx.activity` 1.9.3 already declare 34.
- **AGP 9 defaults change two behaviours.** WordPress-Android behaves the same way on both.
  - Unit tests run for the debug build type only, and `testReleaseUnitTest` no longer exists. The library only reads `BuildConfig.GUTENBERG_EDITOR_URL`, which is the same in both build types.
  - The library's instrumented-test APK now targets `compileSdk` (34) instead of `minSdk` (24).
- **detekt's type-resolution tasks (`detektMain` etc.) are no longer registered.** detekt 1.23.8 only registers them when `org.jetbrains.kotlin.android` is applied, and this PR removes that plugin in favour of AGP 9's built-in Kotlin. CI and `make lint-android` run plain `detekt`, which is unaffected. WordPress-Android is in the same position.
- The Java 8 `source value 8 is obsolete` warning predates this PR; it appears identically on `trunk`. Addressed separately in #646.


nbradbury:
Via Claude:

The review of PR #644 (AGP 9 / Gradle 9.5 upgrade) turned up exactly one issue, and it is
in the PR description rather than the code.

The description lists the loss of detekt’s type-resolution tasks under
”AGP 9 defaults change three behaviours”. That attribution is wrong. detekt 1.23.8’s
DetektPlugin.registerDetektAndroidTasks gates the whole Android task set
(detektMain, detektTest, detektDebug, …) on plugins.withId("kotlin-android") —
confirmed by the constant pool of DetektPlugin.class in the resolved 1.23.8 jar. The PR
drops alias(libs.plugins.jetbrains.kotlin.android) from
android/Gutenberg/build.gradle.kts:2 in favour of AGP 9’s built-in Kotlin support, so the
gate never fires. No AGP flag restores those tasks.

Practical impact is nil — make lint-android and CI run the plain detekt task, whose
source set (src/main/kotlin, src/test/kotlin, …) and baseline are unchanged by this PR.
The cost is that the stated cause would send the next person hunting for an AGP setting
that does not exist.

nbradbury (approved):
@dcalhoun This looks good and testing steps passed so I'm fine to approve. I'll leave it up to you whether to address Claude's comment. :shipit:

dcalhoun:
@nbradbury thanks for the review and sharing the finding. I updated the PR description accordingly.

PR #678, opened 2026-09-17
build(android)!: compile the library against SDK 36

dcalhoun:
## What?

Raises `:Gutenberg`'s `compileSdk` from 34 to 36. This also raises the published AAR's `minCompileSdk` to 36, so consumers must compile against 36 or later, WordPress-Android already does.

## Why?

Both the Demo app and WordPress-Android use 36. 

<p>

The library compiled against API 34 while `:app` and WordPress-Android were both on 36, and AndroidX has moved past it. `androidx.core:core` 1.16.0 declares a minimum compileSdk of 35 and `com.google.android.material` 1.14.0 depends on it, so the entire Android dependency group fails before it can be reviewed:

```
Dependency 'androidx.core:core:1.16.0' requires libraries and applications that
depend on it to compile against version 35 or later of the Android APIs.
:Gutenberg is currently compiled against android-34.
```

36 rather than 37, which newer releases already want, because AGP writes this value into the published AAR as `minCompileSdk` — consumers must compile against at least what this library does. WordPress-Android is on 36, so going further would break their build the way theirs would break ours.

The only visible effect is nine new deprecation warnings for the `java.util.Locale` constructors in `EditorConfigurationTest` and `LocaleResolverTest`. They're left alone deliberately — those call sites pass legacy codes (`iw`, `in`, `no`) precisely to exercise the resolver's alias handling, and `Locale.forLanguageTag` canonicalises them away. `Locale.of` would be the other option, but it needs Java 19 and this project targets 11.

Verified locally:

2. Warning diff between compileSdk 34 and 36 on an otherwise identical tree: 15 → 24, the nine `Locale` deprecations above, none in main source. Nothing sets `allWarningsAsErrors` or lint `abortOnError`.
3. `unzip -p android/Gutenberg/build/outputs/aar/Gutenberg-release.aar META-INF/com/android/build/gradle/aar-metadata.properties` reports `minCompileSdk=36`.

</p>
</details> 

## How?

Bump the version. 

## Testing Instructions

Smoke test the editor. 


adalpari (approved):
Tested among

And it looks good to me!

PR #679, opened 2026-09-17
build(deps): bump the Android dependency group

dcalhoun:
## What?

Superseded #674 and align Android dependency versions with WordPress-Android. 

## Why?

We recently enabled Dependabot for Android dependencies. This helps land the initial updates. 

## How?

Bump relevant versions, while avoiding surpassing WordPress-Android's own declared versions where it has a negative impact. 

<p>

Takes over #674 — Dependabot's `gradle-minor-and-patch` group — and pins `androidx.core` to WordPress-Android's version so it can pass CI. Closes #674 in favour of this.

The group was the first Android run Dependabot ever made here, so it carries months of catch-up. Ten of its bumps land on WordPress-Android's exact versions, which is where this library should be: `:Gutenberg` declares these as `implementation` dependencies, so they reach WordPress-Android's runtime classpath through the published POM and the higher version wins for both.

`androidx.core:core-ktx` is the exception. Dependabot proposed 1.19.0, which requires compileSdk 37; we compile against 36 to match WordPress-Android. 1.16.0 is both what they use and what our SDK level supports.

Cherry-picked commits from #674 onto this stack rather than fixed in place. The wrapper bump that rode inside #674 is gone — #670 excluded `gradle-wrapper` from this group, so Dependabot regenerated it as a catalog-only diff with no `gradlew`/`gradlew.bat` line-ending churn.

Verified locally on this branch:

2. Every bumped version matches WordPress-Android's `gradle/libs.versions.toml` except `robolectric` (4.17 vs their 4.16.1) and `androidx-recyclerview` (1.4.0 vs their 1.3.2), neither of which leaves this repo — `robolectric` is `testImplementation` and `recyclerview` is declared only in `:app`.

</p>
</details> 

## Testing Instructions

Smoke test the editor. 


adalpari (approved):
Tested among

And it looks good to me!

PR #680, opened 2026-09-17
build(android): align the remaining dependencies with WordPress-Android

dcalhoun:
## What?

Align three shared dependencies with WordPress-Android's versions. Revert https://github.com/wordpress-mobile/GutenbergKit/pull/670. 

## Why?

Update stale dependencies. 

Re: Compose BOM ignore removal, its addition to the `.github/scripts/android-toolchain-sync.js` script seems misguided. That script's (and the Dependabot ignore config) focus is ensuring that:

1. WordPress-Android + GutenbergKit composite builds do not break without notice;
2. And we do not surpass WordPress-Android's own version for the relevant packages. 

Compose BOM does not relate to those two points. Additionally, CI failures from updating Compose BOM provide actionable CI failures that can be managed separately. 

## How?

Restore `android-toolchain-sync.js` and Android Dependabot configuration to its previous, narrow state. Bump dependency versions. 

<p>

Four version catalog bumps, plus reverting #670's Compose BOM hunks:

```
androidx.webkit:webkit              1.11.0     -> 1.16.0
androidx.activity:activity-compose  1.9.3      -> 1.10.1
androidx.activity:activity          1.9.0      -> 1.10.1
androidx.compose:compose-bom        2024.12.01 -> 2026.06.01
```

These were stale, which is the direction that has actually cost us: the library compiled against one version while WordPress-Android's app resolved a newer one at runtime. `androidx.webkit` most of all — five minors behind on the WebView library this editor is built on.

Dependabot proposed none of them. `androidx.webkit` publishes `1.18.0-alpha01` as its release metadata, which appears to suppress the update; `activity-compose` was simply never offered; and the BOM was ignored by config. `activity` is here because `activity-compose` 1.10.1 depends on it, so Gradle already resolved 1.10.1 and the old `1.9.0` entry described nothing.

Comparing every `implementation` dependency of `:Gutenberg` against WordPress-Android's catalog by Maven coordinate, all were behind theirs and none ahead. #679 covers most of the rest; `okhttp` is handled separately in #677, and Kotlin (2.4.10 vs their 2.4.20) is left for its own change since matching their toolchain is a different decision from a dependency bump.

**On removing the Compose BOM ignore.** #670 added it, and its `Android toolchain sync` entry, on the premise that a BOM from another release cycle pairs Compose runtime artifacts with a compiler the pinned Kotlin doesn't match. That premise doesn't hold — 2026.06.01 builds cleanly against Kotlin 2.4.10. What actually stopped Dependabot's 2026.09.00 was `checkReleaseAarMetadata` rejecting artifacts that need compileSdk 37:

```
Dependency 'androidx.compose.ui:ui-android:1.12.1' requires libraries and
applications that depend on it to compile against version 37 or later
```

A hard, self-explaining CI failure that needs no help from configuration. AGP and Kotlin stay ignored and watched: their mismatches break a composite build that no CI on either side exercises, which is the case the workflow was built for.

Verified locally on this branch:

2. `make lint-js-fix` and Prettier pass on the three `.github` files.
3. Ran the sync script against WordPress-Android's live `trunk`: AGP and Gradle in sync, Kotlin drifted, one issue filed for it. Replayed the 21-scenario harness from #645 with no regressions.

</p>
</details> 

## Testing Instructions

Smoke test the editor. 


adalpari (approved):
Tested among

And it looks good to me!

PR #688, opened 2026-09-17
docs: state the rule that makes the media field decode safe

jkmassel:
Stacked on #687. Last of nine PRs splitting #621.

## What?

Writes down why decoding the editor's form fields as UTF-8 can't corrupt them, and adds tests for the one condition this code could break on its own.

## Why?

`formFields` reads every non-file form value as UTF-8. UTF-8 decoding doesn't fail on bad bytes — it quietly swaps in a replacement character. So if bad bytes ever did arrive, a caption would come back mangled and nothing would report an error.

That can't happen today, but only because of who is on the other end, and nothing in the code says so. Three things have to stay true:

1. Only the editor's own page can reach this server. It only accepts connections from the device itself, and every request has to carry a per-session token.
2. Text the editor puts in a form field is already valid Unicode. The browser guarantees that when the value is set.
3. The only way a browser can put raw bytes in a form is a file or a Blob, and those always arrive with a filename. Anything with a filename is handled as the file, never as a field.

**If one of those stops being true, iOS and Android mangle the value differently**, so there's no single behaviour that could be documented in place of the rule. Turning off each platform's filename check shows it: the same three bad bytes (`ED A0 80`) come back as three replacement characters on iOS and one on Android. Both measured, not guessed.

## How?

- The three conditions, as a comment on `formFields`, on both platforms.
- Rewrote the comment where the request gets rebuilt — the path used when a processor has changed the file. It said the field bytes were appended "so a non-UTF-8 value is forwarded verbatim rather than coerced", which reads as though bad values were expected there. They can't get there. On iOS the real reason is that `String(data:encoding:)` returns nil on bad UTF-8, and the `?? ""` you'd reach for behind it would throw away the whole field. On both platforms, using the raw bytes keeps this path sending exactly the same request as the plain pass-through it stands in for, so an upload doesn't change shape just because the image got resized.

## Testing Instructions

The tests cover condition 3, the only one this code controls. A request carrying a second part that looks like a Blob — it has a filename, and its bytes aren't valid UTF-8 — must not show up in `fields`, and must not get uploaded in place of the real file.

Both test bodies put `file` first and the other fields after, which is the order the editor actually sends them. Keep it that way. The server treats the first part that has a filename as the file, so moving the Blob ahead of `file` makes the Blob the upload and throws the real file away — while the check on `fields` still passes, so nothing catches it.

The Blob part reaches neither `fields` nor the request sent on to WordPress; it is dropped. This PR doesn't change that.

- [x] The Blob test fails if the server stops excluding parts that have a filename — both platforms
- [x] It also fails if the server takes the *last* part with a filename instead of the first. The check on the uploaded filename is what catches that; without it, the test still passed under that change
- [x] `swift test` — 396 tests green
- [x] `:Gutenberg:testDebugUnitTest` — 30 tests in `MediaUploadServerTest` green
- [x] SwiftLint clean


ios/Sources/GutenbergKit/Sources/Media/MediaUploadServer.swift
```diff
@@ -462,6 +462,22 @@ final class MediaUploadServer: Sendable {
...
+        /// This decode can't mangle anything, but only because of who is on the other end —
+        /// nothing in the code enforces it. Three things have to stay true:
+        ///
+        /// 1. Only the editor's own web page can reach this server. It listens on loopback,
+        ///    and every request has to carry a per-session token.
+        /// 2. Text the editor puts in a form field is already valid Unicode. The browser
+        ///    guarantees that when the value is set, so it cannot hand us bad bytes.
+        /// 3. The only way a browser can put *raw* bytes in a form is a file or a Blob, and
+        ///    those always arrive with a filename. Anything with a filename is handled as
+        ///    the file, never as a field — so raw bytes never reach this decode.
```
dcalhoun, line 475: *Finding from Claude:*

Two predicates doing two different jobs on the same array:

- `:210` — `parts.first(where: { $0.filename != nil })` picks **the file**: the first filename-bearing part.
- `:216` — `parts.filter { $0.filename == nil }` picks **the fields**: only the filename-*less* ones.

So a *second* filename-bearing part is neither. `first` has already returned, and the filter excludes it for having a filename — it falls out of the request entirely. (On this path, at least; the passthrough at `:301` relays the original body verbatim, so it survives there.)

The rule's conclusion is still safe — `:216` excludes *every* filename-bearing part, so raw bytes can't reach the decode. It's the stated reason that's off: "never as a field" holds for all of them, but "handled as the file" holds only for the first, and implies a second Blob gets processed as an upload when it's actually dropped.

```suggestion
        /// 3. The only way a browser can put *raw* bytes in a form is a file or a Blob, and
        ///    those always arrive with a filename. Anything with a filename is excluded from
        ///    the fields, so raw bytes never reach this decode. (The first such part is the
        ///    file; on this path any others are dropped.)
```

Same wording on `MediaUploadServer.kt:614-616`.

ios/Sources/GutenbergKit/Sources/Media/MediaUploadServer.swift
```diff
@@ -750,11 +766,12 @@ class InternalMediaClient: @unchecked Sendable {
...
-        // ahead of the streamed file. They are small, so keeping them in memory is
-        // fine; `contentLength` counts them via `preamble.count`. Field values are
-        // appended as raw bytes (not through String) so a non-UTF-8 value is
-        // forwarded verbatim rather than coerced to empty.
+        // The non-file parts (post, additionalData) go into the preamble ahead of the streamed
+        // file; they're small, and `contentLength` counts them via `preamble.count`. Their
+        // values are appended as raw bytes rather than through `String(data:encoding:)`, which
+        // returns nil on bad UTF-8 — and the `?? ""` you'd reach for behind it would quietly
+        // drop a whole field. Raw bytes also keep this re-encode byte-for-byte identical to
+        // the plain passthrough it replaces. (Bad bytes can't get here; see `formFields`.)
```
dcalhoun, line 774: *Finding from Claude:*

This says bad bytes can't reach `multipartBodyStream`, but `multipartBodyPreservesNonUTF8FieldValue` — "forwards a non-UTF-8 field value verbatim", `MediaUploadServerTests.swift:836` — calls this function with exactly those bytes. The rationale this replaced was the text that test pointed at.

The risk is someone reads the new sentence and deletes the test as vacuous. Scoping it would keep both true, e.g. "Bad bytes can't reach here *via the server*; this function's own handling of them is pinned by `multipartBodyPreservesNonUTF8FieldValue`."

Same on `MediaUploadServer.kt:719`.

ios/Tests/GutenbergKitTests/Media/MediaUploadServerTests.swift
```diff
@@ -491,6 +491,76 @@ struct MediaUploadServerTests {
...
+  @Test("keeps a binary Blob part out of an uploader's fields")
+  func binaryPartExcludedFromFields() async throws {
+    // Pin rule 3: a Blob always has a filename, so it's dropped before the decode.
+    let uploader = RecordingUploader()
+    let server = try await MediaUploadServer.start(uploader: uploader, internalClient: MockInternalMediaClient())
+    defer { server.stop() }
+
+    let boundary = UUID().uuidString
+    var body = Data()
+    // Ordered as `uploadToServer` emits it: the file first, then additionalData.
```
dcalhoun: *Finding from Claude:*

This body pins the safe ordering, and `#expect(received.filename == "photo.jpg")` at `:529` does catch a *last*-wins regression. What stays unpinned is the hazard the PR description itself calls out: a filename-bearing part placed *before* `file`.

`parts.first(where: { $0.filename != nil })` (`MediaUploadServer.swift:210`) selects positionally, while everything else selects by name — `body.get('file')` in `api-fetch.js:267`, `addFormDataPart("file", …)` at `.kt:723`. What keeps that safe today is `uploadToServer` appending `file` before iterating `additionalData`, which is upstream and unpatched here.

A sibling test with the Blob part first, asserting `filename == "photo.jpg"`, would fail today — which is the point of adding it. Narrowing `:210` to `$0.name == "file"` would remove the dependency outright.

Reasonable as a follow-up since `:210` is pre-existing and this is a docs PR — but right now the warning exists only in the PR body, and that disappears on merge.

ios/Tests/GutenbergKitTests/Media/MediaUploadServerTests.swift
```diff
@@ -491,6 +491,76 @@ struct MediaUploadServerTests {
...
+    let uploader = RecordingUploader()
+    let server = try await MediaUploadServer.start(uploader: uploader, internalClient: MockInternalMediaClient())
+    defer { server.stop() }
+
+    let caption = "Grüße 🎉 日本語"
+    let boundary = UUID().uuidString
+    var body = Data()
+    body.append("--\(boundary)\r\n")
+    body.append("Content-Disposition: form-data; name=\"caption\"\r\n\r\n")
+    body.append("\(caption)\r\n")
```
dcalhoun: *Finding from Claude:*

The PR description says "Both test bodies put `file` first and the other fields after… Keep it that way." This one puts `caption` first — same on `MediaUploadServerTest.kt:333`.

No hazard in the test itself: `caption` carries no filename, so there's only one candidate for the file part and ordering can't change the selection. But that sentence is what a maintainer consults before reordering a body, and it's wrong about its own diff — so the likely outcome is someone "fixing" the Blob test instead, which is the one where ordering is load-bearing.

Narrowing the instruction to the Blob test would settle it.

dcalhoun (approved):
Looks good. Captured a few inline suggestions from Claude, but none of them are blocking.

PR #700, opened 2026-09-18
test: extend vitest's default excludes and skip build/

dcalhoun:
## What?

Configure Vitest `exclude` with defaults and the project's `build` directory.

## Why?

Avoid unnecessary directory traversals.

## How?

Spread the Vitest defaults along with `build`.

## Testing Instructions

N/A, no user-facing changes.

N/A, no user-facing changes.

## Screenshots or screencast 

N/A, no user-facing changes.

<details>
<summary>Agent summary</summary>

Spreads vitest's `defaultExclude` into `vitest.config.js`'s `exclude` and adds `build/**`.

`build/` is gitignored scratch and output space, so a `*.test.js` file that lands there — an extracted archive, a downloaded artifact — is collected and run as one of ours.

The config also replaced vitest's `exclude` defaults wholesale rather than extending them, which left two smaller gaps: `**/.git/**` was never excluded (vitest globs with `dot: true`, so `.git` was walked on every run), and `node_modules/**` was root-anchored, so a nested `node_modules` anywhere else in the tree had its tests collected. Spreading `defaultExclude` covers both and keeps the list in step with future vitest releases.

CI is unaffected either way, since `build/` is empty there; this fixes local runs.

Verified against a copy of the previous config: `*.test.js` files placed in `build/`, in `vendor-scratch/node_modules/` and in `.scratch-nested/node_modules/` were all collected before the change and none are after. The real suite is unchanged — `make test-js` passes with 17 files and 231 tests, and `make lint-js-fix` is clean.

</details>


nbradbury (approved):
Looks good! :shipit:

PR #702, opened 2026-09-18
build(deps-dev): migrate to ESLint 10 and flat config

dcalhoun:
## What?

Update to ESLint 10 now that Gutenberg core's packages support it. 

## Why?

Keep up-to-date with Gutenberg and other dependencies. 

## How?

- Address lint errors 
- Bump relevant dependencies 
- Remove Dependabot configuration excluding ESLint version bumps 
- Add nuance to agent instructions regarding lint tooling 

## Testing Instructions

N/A, no user-facing changes. 

N/A, no user-facing changes. 

## Screenshots or screencast 

N/A, no user-facing changes. 

---

<details>
<summary>Agent summary</summary>

[WordPress/gutenberg#64782](https://github.com/WordPress/gutenberg/issues/64782) is closed — `@wordpress/eslint-plugin` supports ESLint 9 and 10 as of 25.0.0, which also dropped eslintrc. The ESLint 8 pin had frozen the WordPress ruleset two majors back at 24.3.0 (no plugin bump since March 2026).

**Toolchain**: `eslint` 8.57 → 10.10, `@wordpress/eslint-plugin` 24.3 → 26.0, `eslint-plugin-react-refresh` 0.4 → 0.5, `globals` added. Matches Gutenberg, which runs ESLint 10.

**Config**: `.eslintrc.cjs`, `e2e/.eslintrc.cjs`, and `.eslintignore` become `eslint.config.js`. Three details are deliberate:

- The node `import/resolver` is carried over. Without it, `import/no-unresolved` fails on the patched `@wordpress/block-editor` subpaths (extensionless `exports` targets) and on `@wordpress/preferences-persistence` (its `types` condition names unpublished build output).
- `test-playwright` skips `e2e/**/*.test.js`. `eslint-plugin-playwright` goes 0.15 → 2.x, whose rules read vitest `it()` blocks as stray assertions — 17 errors in `e2e/fetch-json.test.js` otherwise.
- `.build/` is ignored explicitly. Flat config does not skip dot directories, so ESLint otherwise walks the 1.5 GB SwiftPM output.

**Peer range**: `eslint-plugin-import@2.32.0` still caps at ESLint 9, so its `eslint` peer is overridden to the root version — narrower than the repo-wide `legacy-peer-deps` Gutenberg uses.

**Code changes**: 9 unused `catch` bindings (ESLint 9 flipped `no-unused-vars` `caughtErrors` to `"all"`) and 2 `forwardRef` components renamed to Gutenberg's `Unforwarded*` convention (ESLint 10 `no-shadow` flags a named function expression shadowing its own binding).

**Why stacked on #691**: plugin 26 depends on `@wordpress/theme` 2.x, whose modern subtree makes npm dedupe 10 in-range production `@wordpress/*` packages upward. On trunk that is unwanted drift in a lint PR; on #691 those packages already sit at exactly those versions, so the lock diff stays confined to the lint subtree.

**Verified**: `make lint-js` clean (112 files linted vs. 110 before — gains `.prettierrc.cjs` and the config itself, none dropped), 231 unit tests pass, `check:wp-packages` passes. E2E not run.

**Follow-up**: the `@typescript-eslint/typescript-estree` → `minimatch` `9.0.9` override is now a forced major downgrade (typescript-estree 8 declares `^10.2.2`) with no corresponding advisory. npm will not drop the lock entry without hand-editing it, so it is left for a separate PR.

</details>


package.json
```diff
@@ -141,6 +142,9 @@
 		},
 		"@wordpress/babel-preset-default": {
 			"@babel/core": "7.29.7"
+		},
+		"eslint-plugin-import": {
+			"eslint": "$eslint"
```
dcalhoun, line 147: The latest `eslint-plugin-import` release declares a peer of `eslint@^9`. This overrides the peer to match the root `eslint` version.

adalpari (approved):
LGTM!

PR #712, opened 2026-09-20
docs: correct the SwiftLint CI paragraph in AGENTS.md

mokagio:
## What?

Rewrites the four-line paragraph in `AGENTS.md` § Swift Linting that describes how SwiftLint runs on CI.

## Why?

The paragraph said CI is "the exception" to the single `swiftlint_version` pin — that the `linter` agent queue "invokes that image's own SwiftLint binary rather than the pinned one", and that CI could therefore fail on rules that pass locally.

That is not what happens. `command: swiftlint` on `queue: linter` is not a shell command: it dispatches to the CI toolkit's `run_swiftlint`, which reads `swiftlint_version` out of this repo's own `.swiftlint.yml` and runs `ghcr.io/realm/swiftlint:"$SWIFTLINT_VERSION"` under Docker, erroring if the key is missing. Laptop and CI resolve to the same declared version from the same single declaration, by two different mechanisms.

This matters more than four lines usually would: it lives in an `AGENTS.md`, so it was loaded into agent context on every session in this repo — teaching both people and agents that a working system is broken. Three independent passes during the Apps Infra support rotation survey reached the wrong conclusion, two of them citing this paragraph.

Raised as [AINFRA-3002](https://linear.app/a8c/issue/AINFRA-3002), which holds the full dispatch trace.

## How?

Docs only. No behaviour change.

## Testing Instructions

The claim is verified against a real build rather than by reading code. The [`:swift: SwiftLint` job of build 2936](https://buildkite.com/automattic/gutenbergkit/builds/2936#01a0b998-c850-4160-9a8e-b1de33119880) (commit `7fda90ef` on `trunk`) logs, under `set -x`:

```
++ docker run --rm -v …/gutenbergkit:/workspace -w /workspace \
     --entrypoint swiftlint ghcr.io/realm/swiftlint:0.63.3 \
     lint --quiet --reporter relative-path --strict
```

`0.63.3` is exactly the `swiftlint_version` in `.swiftlint.yml`.

Grepping the `AGENTS.md` of every Apple-platform repo in the fleet turned up no other copy of the claim, so this is the only file that needed correcting.

---


AGENTS.md
```diff
@@ -94,10 +94,12 @@ time, so for local runs and the editor hook the binary version and the rule set
...
-CI is the exception: the `:swift: SwiftLint` step runs on the shared `linter`
-agent queue, which invokes that image's own SwiftLint binary rather than the
-pinned one. If the image moves far enough ahead of `swiftlint_version`, CI can
-fail on a rule that passes locally — bump the pin to resolve it.
+CI reaches the same version by a different route. The `:swift: SwiftLint` step
+runs on the shared `linter` agent queue, where the CI toolkit's `run_swiftlint`
+reads `swiftlint_version` out of `.swiftlint.yml` and runs
+`ghcr.io/realm/swiftlint:<version>` under Docker. Local runs, the editor hook
+and CI therefore all resolve to the same single declaration, and bumping
+`swiftlint_version` moves all three together.
```
mokagio:
```suggestion
CI also uses the SwiftLint version defined in `.swiftlin.yml` `swiftlint_version`, but notice it does so by fetching the matching Docker image.
```

dcalhoun (approved):

